PR #2, opened 2018-08-30
Make hourly dungeons resets patch appropriate

ghost:


ratkosrb:
Wowwiki seems to say 1.9, not 1.10.
http://wowwiki.wikia.com/wiki/Instance_limitations?oldid=114615

We can verify this by checking the existence of the TRANSFER_ABORT_TOO_MANY_INSTANCES error message or the SMSG_TRANSFER_ABORTED packet in general when we have support for those clients.

ratkosrb:
I just tested on the 1.9.4 client and the error message exists there. So it cant have been added in 1.10. When we have 1.8 client support i well test it there as well to see if it is indeed a 1.9 addition.

PR #3, opened 2018-08-30
Increase Kurrinaxx's pull range

ghost:
It shouldn't be possible to skip Kurrinaxx.


ratkosrb:
How about we just add an aggro range multiplier column to creature_template?

ghost:
Dunno if it's worth it for a hand full of NPCs

Kittnz:
Would be a bit overkill to add this on creature_template.

ratkosrb:
It's on the screenshot of blizzard's creature editor, and is implemented in cmangos. This is like saying it's overkill to implement gossip menus, when we can hardcode it in the AI for every creature.

PR #5, opened 2018-08-30
Quests should be doable in Raid < Patch 1.3

ghost:
http://wowwiki.wikia.com/wiki/Patch_1.3.0

### DIFF
--- a/src/game/QuestDef.cpp
+++ b/src/game/QuestDef.cpp
@@ -223,8 +223,5 @@ bool Quest::IsAllowedInRaid() const
     if (m_QuestFlags & QUEST_FLAGS_RAID)
         return true;
 
-    if ((sWorld.GetWowPatch() < WOW_PATCH_103) || sWorld.getConfig(CONFIG_BOOL_QUEST_IGNORE_RAID))
-        return true;
-    
-    return false;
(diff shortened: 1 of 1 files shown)


ratkosrb:
This is actually wrong, the patch notes don't mean what you think.
```
Quests that are not completable while in a raid group are no longer allowed if the instance is owned by a raid group, regardless of whether you are currently in that raid group or not. 
```

By the first phrase it is already clear that the restriction of not being to do normal quests in a raid did exist prior to the patch. What they actually changed is the ability to clear part of the dungeon in a raid group, and then come back with only 5 people to finish a non-raid quest.

Lets say you make a 10 man raid to clear half of one of the mini-raid dungeons like Stratholme (10 man before 1.10), Scholomance (10 man before 1.10) or Blackrock Spire (15 man before 1.10). Then you go in and kill a few of the harder bosses, and disband. Then later you come back with only 5 of those people in a normal group and try to complete some of the non-raid quests that require you to do something deep inside the dungeon. After patch 1.3, you can no longer do that. You cannot clear everything prior to the quest in a raid group, and then come back in a normal party to do your non-raid quest. If you killed anything inside the dungeon in raid group, this save now belongs to raid, and you may not use it to finish your non-raid quests.

PR #8, opened 2018-08-30
NPC Lothos Riftwalker shouldn't port you to MC < Patch 1.3

ghost:


src/scripts/eastern_kingdoms/searing_gorge/searing_gorge.cpp
```diff
@@ -75,12 +75,7 @@ bool GossipHello_npc_lothos_riftwaker(Player* pPlayer, Creature* pCreature)
     if (pCreature->isQuestGiver())
         pPlayer->PrepareQuestMenu(pCreature->GetGUID());
 
-    // Note that the attunement quests were not actually added to the game until 1.3
-    // Prior to this, there was simply a 'discovery quest', and you must run through
-    // BRD to get into MC.
-    // Furthermore, such a system is not currently supported in our core, since if you
-    // were to die in MC there is no way to recover your corpse!
-    if (pPlayer->GetQuestRewardStatus(7487) || pPlayer->GetQuestRewardStatus(7848))
+    if ((pPlayer->GetQuestRewardStatus(7487) || pPlayer->GetQuestRewardStatus(7848)) && sWorld.GetWowPatch() >= WOW_PATCH_103)
```
ratkosrb: The patch check here is actually redundant, since the quests themselves don't exist in 1.2. The quest check will always return false in 1.2.

PR #20, opened 2018-09-11
Progression: Changes to item cooldowns

ratkosrb:
Corrected pre-1.11 cooldowns for a ton of items. Used the item history page on allakhazam, the official patch notes, a 1.10 db for mangos from 2006, and multiple wowemu scp databases from different dates to find out what the old values were and when they changed.


DarkmoonRabbit:
good

PR #30, opened 2018-10-23
Taming the Beast improvements

ghost:
Tamed beasts should be removed correctly on quest hand in.
Also improved pet follow angle in this quest.


Phatcat:
What? No it's shouldn't.. https://youtu.be/FaV6oAteJGI?t=6894 (also the pet is not supposed to despawn but supposed to do like any other charm and attack on charm end; https://youtu.be/FaV6oAteJGI?t=7207)

ratkosrb:
Good find @Phatcat. I had changed it to despawn on dismiss based on another video, but it seems it worked differently in vanilla.

PR #163, opened 2019-03-16
Implement Warden Anticheat

ratkosrb:
Credits to mangoszero for their public implementation.

You need to make a folder named "warden_modules" and put these there:
https://github.com/vmangos/warden_modules


bomrox:
checked, works!

bomrox:
there are no settings in the config, and the server automatically reboots every 15 minutes

Kittnz:
This is because of Honor calculation, look it up in the core. Has nothing to do with this.

PR #202, opened 2019-06-04
Prevent equipment fist weapons without having proper skill

Kittnz:


ratkosrb:
The before and after code is equivalent, except that GetSkill() gets needlessly called twice.

PR #218, opened 2019-06-22
Cherry-pick's GuybrushGit's melee hit table changes

mako-taco:
Straight port of @GuybrushGit 's commit to lh-core. Replaced calls to `std::min` with ternaries, but otherwise no changes.

See original commit here: https://github.com/lh-server/core/commit/705b0325b96410e47d17a278c87353ec9fbfd273


mako-taco:
vs lvl 63:
image

PR #219, opened 2019-06-22
Support player names in `.god` command

mako-taco:
Allows for GMs to use `.god <on/off> <name>` to toggle god mode on players without targeting them. Useful in macros to flag an entire raid group at once.

Current behavior remains unchanged.


ratkosrb:
Btw, you can let players toggle their own god mode on or off without being GMs with these spells.
https://classicdb.ch/?spell=20115
https://classicdb.ch/?spell=20114

Can be inserted into `playercreateinfo_spell` and any new characters will have the ability in their spell book.

PR #308, opened 2019-11-19
Fix a ton of warnings found using clang.

ratkosrb:
Installed clang and trying to fix all the warnings it produces.


src/game/HardcodedEvents.cpp
```diff
@@ -1156,8 +1154,8 @@ uint32 ScourgeInvasionEvent::GetNewRandomZone(uint32 curr1, uint32 curr2)
 void ScourgeInvasionEvent::UpdateWorldState()
 {
     // Updating map icon worlstate
-    int ATTACK_ZONE1 = sObjectMgr.GetSavedVariable(VARIABLE_NAXX_ATTACK_ZONE1);
-    int ATTACK_ZONE2 = sObjectMgr.GetSavedVariable(VARIABLE_NAXX_ATTACK_ZONE2);
+    //int ATTACK_ZONE1 = sObjectMgr.GetSavedVariable(VARIABLE_NAXX_ATTACK_ZONE1);
+    //int ATTACK_ZONE2 = sObjectMgr.GetSavedVariable(VARIABLE_NAXX_ATTACK_ZONE2);
```
Kittnz, line 1158: this isn't used?

ratkosrb: nope, does not appear to be since it compiles

Issue #342, opened 2020-01-10
Missing fishing loot (Tirisfal - ID: 6291)


schell244:
Raw Brilliant Smallfish can't be fished from waters in Tirisfal (or loot chance is set to low).
(This fish is required for a cooking recepie 1-65)
I did not check the other locations, where it should be part of the loot table as well.
All locations can be found at:
https://classic.wowhead.com/item=6291/raw-brilliant-smallfish

**Steps to reproduce**:
1. learn fishing 
2. got to "Brightwater Lake"
3. cast fishing 

I increased my fishing skill from to 0-75 and did not get any Raw Brilliant Smallfish.

**Client version**: 
1.21.1

**Commit hash**: 
Core revision: 2749d0ccd44b0e9814e0 / 2020-01-09 22:43:06 +0200 / Win64 (little-endian)

**Operating system**: 
Win10

schell244:
Closed with #354

PR #350, opened 2020-01-19
Fix fishing loot tables in Tirisfal

schell244:
fix for issue https://github.com/vmangos/core/issues/342

I think it would be nice to rename the table "entry" to "area_id" in fishing_loot_template, to not be confused with zone id.


ratkosrb:
All the loot tables have the same structure because they share the same loading code.

schell244:
Okay, in this case renaming makes no sense. Thank you.

schell244:
@ratkosrb are you on discord? If you are ok with that, I would like to ask you a few questions regarding this project and it's dev process.

schell244:
I'll do some more detailed research and rework this again.

PR #368, opened 2020-02-12
Added Okla event and fixed Gubber Blump vendor list.

GrenderG:
https://wowwiki.fandom.com/wiki/Okla

Some wait times and orientation might not be 100% correct, someone should check it on live servers. The rest should be okay.


GrenderG:
I have been sent a video showing how this event works in Classic and it's a bit different than this implementation. Leave this PR open in case anyone wants to make it completely Blizzlike.

Proof I received: https://streamable.com/vj9x1

NickTyrer:
@ratkosrb can this PR be closed now?

PR #377, opened 2020-02-19
Implemented Lordaeron's Blessing, fixes #338.

### LINKED ISSUE #338: Lordaeron's Blessing (30238) doesn't have any effect.
**Current behavior**: 
You get this buff (https://classic.wowhead.com/spell=30238/) from https://classic.wowhead.com/object=181682 but it doesn't have any effect.

**Expected behavior**: 
You should have 5% more HP while in Eastern Plaguelands, Scholomance or Stratholme.

GrenderG:
I haven't figured out a better way to only give the effect while inside Eastern Plaguelands, Scholomance or Stratholme, I wanted to handle it inside `Player::UpdateAreaDependentAuras` but I couldn't.

If any of you comes with a better way of handling it, just let me know. Or, @ratkosrb, if you know a better way, you can also merge this and commit the improvement after merging.

Apart from that, it works perfectly fine at this moment.

### DIFF
--- a/src/game/Objects/Player.cpp
+++ b/src/game/Objects/Player.cpp
@@ -6619,6 +6619,23 @@ void Player::UpdateArea(uint32 newArea)
     }
 
     UpdateAreaDependentAuras();
+
+    if (HasAura(30238)) // Lordaeron's Blessing
+    {
+        bool isEffectActive = HasAura(31906);
+        switch (newArea)
+        {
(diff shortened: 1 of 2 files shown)


ratkosrb:
Fixed in https://github.com/vmangos/core/commit/0e0211909e606d40f5c80630f8323feb4e832ebc.

PR #382, opened 2020-02-25
Generate correct random property on buying item from vendor.

ghost:
* Correct enchantment for Shimmering Gloves to match Classic data: https://classic.wowhead.com/item=6565/shimmering-gloves


sql/migrations/20200225214500_world.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200225214500');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200225214500');
+-- Add your query below.
+
+-- Correct Shimmering Gloves enchantment id to match Classic data: https://classic.wowhead.com/item=6565/shimmering-gloves
+UPDATE `item_template` SET `random_property`='710' WHERE `entry`='6565';
```
ratkosrb: The random property used by this item is already 710. Template data for all items was updated based on the blizzcon classic demo back in 2018.

sql/migrations/20200225214500_world.sql
```diff
@@ -0,0 +1,22 @@
...
+-- However, they shouldn't be sold at NPC Zizzek: https://classic.wowhead.com/npc=3572/zizzek
+-- Seems, it was a human error to add Shimmering Gloves (6565) instead of a Strong Fishing Pole (6365): http://classicdb.ch/?npc=3572
+DELETE FROM `npc_vendor` WHERE `entry` = '3572' AND `item` = '6565';
```
ratkosrb: This item was already removed from the vendor back in 2017.
https://github.com/vmangos/core/blob/58e83ed56bd863c3ad433931a3a08b5f4fac5e76/sql/old_migrations/20171026201902_world.sql#L78

ghost: Sorry, I first reinstated item at vendor in my DB, and only then noticed it shouldn't be here.

PR #389, opened 2020-02-28
Enable GM to modify player's rank points online.

lbcheng888:
setrp changes owner's rank point only ,while modify could change player's and owner's.


ratkosrb:
This is the same thing _.honor setrp_ does.
https://github.com/vmangos/core/blob/development/src/game/Commands/CharacterCommands.cpp#L2200

lbcheng888:
setrp only changes the owner's, modify can change player's and owner's.

ratkosrb:
I updated the other command to target other players.

lbcheng888:
Cool, I have checked, good fix!

PR #394, opened 2020-03-02
german script text locales - silverpine forest quests

schell244:
- Escorting Erland, source: https://www.youtube.com/watch?v=HOSYHXg3pdc
- Pyrewood Ambush, source: https://www.youtube.com/watch?v=LG9hExSKYLI

### DIFF
--- a/sql/migrations/20200302101500_world.sql
+++ b/sql/migrations/20200302101500_world.sql
@@ -1,0 +1,35 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200302101500');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200302101500');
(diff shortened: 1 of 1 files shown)


ratkosrb:
That is not what 'sniff' means.

schell244:
Would source be better?

ratkosrb:
We should just update the script to use broadcast text ids instead of this custom table.

schell244:
I didn't expect broadcast texts to contain these script texts. I'll try to find the matching ids and update this PR. Thank you.

ratkosrb:
Broadcast texts contain all script texts. The script is old and that's why it uses the `script_texts` table instead of `broadcast_text`.

schell244:
I found all matching broadcast_text strings:

script_texts = broadcast_text ID

Quest: Escorting Erland
-1000306 = 481
-1000307 = 482
-1000308 = 543
-1000309 = 544
-1000310 = 541
-1000311 = 483
-1000312 = 484
-1000314 = 535
-1000315 = 536
-1000316 = 537
-1000317 = 539
-1000318 = 538
-1000313 = 534

Quest: Pyrewood Ambush:
-1000553 = 542
-1000554 = 545

PR #400, opened 2020-03-04
Add command to mount any targeted creature.

ghost:
* For GMs. Don't need a spell or an item, just hop onto something and ride off into the sunset.


coolzoom:
@shenna 
having fun with it but im having issue with it as well
1, level a troll character
2, select a pig/boar
3, input mount, then you will become a pig:D



image

coolzoom:
same as the scorpion :D
image

coolzoom:
it may not issue at all becuase they may dont have mounting point. still have a lot fun.

ghost:
Yeah, if model doesn't have mounting points you'll just morph into it. Some humanoids will have you sitting on their right arm for some reason.

PR #430, opened 2020-03-26
More convenient way to fly for GMs.

ghost:
* Use two macros: **.fly on** and **.fly off** to travel, it works similar to Druid Flight Form.
* Character won't fall trying to jump while flying.
* I couldn't figure out why it goes straight down on .fly off instead of landing in a curve. 

Exclude visual changes (morph, size, spell) if you don't like them.
I personally prefer flying in the air over swimming :)


ratkosrb:

PR #447, opened 2020-04-05
Added support for item scripts.

ghost:
* This is not exactly Blizzlike feature but it's very useful when it comes to creation custom content and events.


ratkosrb:
This feature was purposefully removed because it has no blizzlike uses, and leads people to a non-blizzlike way of thinking how to implement things. All item effects in the game are Spells. If you want to have custom code that runs when an item is used, you should assign a Dummy Spell to the item like Blizzard does, and then handle that in SpellEffects.cpp. I will be closing this pull request now since the functionality it adds is redundant, you can accomplish the same goal with a spell. But thank you anyway for wanting to contribute.

Here is an example how to script an item using a spell:
https://github.com/vmangos/core/commit/9f619bd71c32b07671d0403645869e5203a79044
https://github.com/vmangos/core/commit/df6193548777073719131dcf4eb31f9981894bd7

Issue #434, opened 2020-03-27
Warlock quests a few bugs


NickTyrer:
**Current behavior**: 

1. ~~The quest 'An Imps Request' is incorrectly started by Richard Kerwin, Luther Pickman, Ursula Deline (5495), Sandahl (5496), Thistleheart (5171)  Alexander Calder (5173), Grol'dar and Mirket (see https://classic.wowhead.com/quest=8419/an-imps-request).~~

2. ~~The quest 'Mor'zul Bloodbringer' is incorrectly started by Greshka and should be started by Martha Strain (see https://classic.wowhead.com/quest=7562/morzul-bloodbringer).~~

3. ~~Briarthorn (5172) should start the quests An Imp's Request and In Search of Menara Voidrender (4736)~~

4. ~~Zevrost is offering the alliance version of the quest 'In Search of Menara Voidrender' (4736). He should offer quest id 4737~~

5. ~~Demon trainers Martha Strain and Kurgul are missing the grimoires 'Devour Magic Rank 3'~~

6. ~~Spackle Thornberry (5520) is missing 1 grimoire~~

7. ~~During the final encounter in DM West for the Warlocks epic mount the animation does not appear around the summoning circle to indicate that the process is complete.
(see https://youtu.be/Y3SqAUiSLGA). When the process is complete and all the glyphs have appeared around the summoning circle an animation appears showing that it is time to summon the dreadsteed. (see 22:55 https://www.youtube.com/watch?v=Tx8o9vRNlkg&t=390s)~~

8. ~~Cylina Darkheart (6374) should not sell  Grimoire of Lash of Pain (Rank 6)~~

9. ~~Wren Darkspring (6376) should not sell Grimoire of Soothing Kiss (Rank 2)~~

10. ~~Dane Winslow (6373) should not sell Grimoire of Lash of Pain (Rank 5)~~

11. ~~Warlock trainer Richard Kerwin is missing multiple trainable spells(he should teach them all).~~

12. ~~Warlock trainer Kartosh (988) is missing Hellfire Rank 1.~~

13. Gorzeeki Wildeyes should not sell the items Xorothian Glyphs, lodestone and J'eevee's Jar untill Imp Delivery is completed and you are on the dreadsteed of xoroth quest . Also he should only sell to warlocks. see https://classicdb.ch/?npc=14437#comments

[...]

PR #452, opened 2020-04-09
Fix warlock quests and trainers re #434

NickTyrer:
Fixes 1, 2, 3, 4, 5, 6, 8, 9, 10, 11, 12 and 14 in issue #434


ratkosrb:
Use the batch script to create a new migration. Currently you have a database id different from the file name, which will make the server unable to start. It will complain that migration 20200409171600 is missing, because you are inserting 20200405163750 in the table instead.

For Windows:
https://github.com/vmangos/core/blob/development/sql/make_migration.bat
For Linux:
https://github.com/vmangos/core/blob/development/sql/touch_migration.sh

sql/migrations/20200409171600_world.sql
```diff
@@ -0,0 +1,61 @@
...
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200409171600');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200409171600');
+-- Add your query below.
+
+
+--remove all incorrect trainers from An Imps Request (8419) Richard Kerwin (4565), Luther Pickman (4564), Ursula Deline (5495), Sandahl (5496), Thistleheart (5171), Alexander Calder (5173), Grol'dar (3324), Mirket (3325) and Kartosh (988)
+DELETE FROM mangos.creature_questrelation WHERE quest = 8419 AND id NOT IN(461,3326,4563);
```
Kittnz: don't need database ref here

NickTyrer: Cheers for the review, I have removed all database references.

NickTyrer:
@ratkosrb this should be good to go, all issies fixed in #434 except number 13 (couldent figure how to fix that).

ratkosrb:
That must be a condition on the gossip option.

PR #455, opened 2020-04-11
Add support for Drone CI and update Travis CI

tonymmm1:
Adds support for Drone CI automated builds, file has 2 pipelines configure for amd64 and arm64 respectively.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -6,7 +6,7 @@ platform:
   arch: amd64
 
 steps:
-- name: build
+- name: build-gcc
   image: ubuntu:xenial
   commands:
   - apt-get update -y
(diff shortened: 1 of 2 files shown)


ratkosrb:
Do I have to go enable this somewhere? For Travis I think I had to go sign up on their website?

tonymmm1:
This file allows either use of self-hosted or https://cloud.drone.io/ which is hosted by developers of Drone CI.

tonymmm1:
I just thought it would be convenient for people forking and or mirroring to have another CI option.

tonymmm1:
Also the arm64 pipeline is optional and can be removed or commented out.

ratkosrb:
To be honest the arm64 option is the more useful one, because amd64 linux builds already get tested on Travis. If you know how to get some MSVC and Clang based CI working that would be great too.

tonymmm1:
Do you want clang and gcc as separate pipelines, I will look into it and get back to you. I am not as familiar with compiling for Windows so that is not a guarantee.

ratkosrb:
Ideally we would have simultaneous builds running for GCC, Clang, and MSVC after every commit. At cmangos they already have automated Windows builds working, but I don't know how to set this up. Do you know if its possible to get those 2 working with Travis? They use AppVeyor at cmangos.

Untitled

https://ci.appveyor.com/project/cmangos/mangos-classic/build/0.18.1317

tonymmm1:
I just added Clang pipeline will finish with MSVC on Drone and then look at Travis.

tonymmm1:
I will look into this for now https://github.com/mstorsjo/msvc-wine.

tonymmm1:
I think this new Travis file should work for parallel builds of gcc and clang.

tonymmm1:
I don't quite know how to use MSVC on Linux without some weird hacks like Wine or on a native Windows appliance.

ratkosrb:
The MSVC build ought to be on Windows.

tonymmm1:
Yeah CI environment is pretty much Linux only unless there is a tool chain I do not know about and or Travis has a work around. Sorry.

tonymmm1:
@ratkosrb I will test out the travis file and see if it works on my branch.

tonymmm1:
@ratkosrb Ok build is running on two simultaneous compiles so it should be ready to go.
image

tonymmm1:
Ok now I am looking https://docs.travis-ci.com/user/reference/windows/ will try to get a Windows build going as well.

tonymmm1:
I am leaving it alone for now in working order as Travis direct Windows compile seems to be in beta and I am unable to get ACE libraries compiled for Windows. In the future I might create a custom Chocolatey library for ACE and then just have the Travis pull that.

Issue #436, opened 2020-03-29
Druid / Shaman quests a few bugs


NickTyrer:
**Current behavior**: 
**Druid**
1. ~~Gart Mistrunner (3060) and Fylerian Nightwing (4219) incorrectly starts the quest Heeding the Call (5924), Theridran (5505) should start the quest~~

2. ~~Sheal Runetotem (3034) Denatharion (4218), Sheldras Moontree (5504), Golhine the Hooded (9465) and Jannos Lighthoof (8142) incorrectly start the quest Torwa Pathfinder (9063)~~

3. ~~During the quest Trial of the Lake (28) you can turn in the quest without activating the Shrine Bauble. If the Bauble has not been activated Tajarri should offer the in progress text (https://classic.wowhead.com/quest=28/trial-of-the-lake).~~

4. ~~There is a typo in the completion text for Trial of the Sea Lion should be "your" (see https://classic.wowhead.com/quest=30/trial-of-the-sea-lion)~~

**Shaman**
1. ~~Xanis Flameweaver (5906) should start Call of Fire (1523)~~

2. ~~Siln Skychaser (3030), Kardris Dreamseeker (3344), Sian'tsu (3403) and Haromm (986) incorrectly start the quest Elemental Mastery (8410)~~

3. ~~During the quest Call of Water (1103):~~
  ~~- The mobs should be invisible untill you use the water sapta~~
  ~~- There is some graphical glitching on the stone~~
  ~~- vmangos version: https://youtu.be/vIa643Wmpco~~
  ~~- classic version: see 11:41 https://www.youtube.com/watch?v=VxP6CwWyxIc~~

4. ~~During the quest Call of Air (1532) NPC Prate Cloudseer (5905) is missing the completion text (https://classic.wowhead.com/quest=1532/call-of-air).~~

5. ~~Shaman trainer Haromm (986) is missing multiple trainable spells~~

6. ~~Starting the quest Call of Earth (1520) should enable Earth Sapta (1462) and completing the quest Call of Earth (1520) should disable Earth Sapta (1462)~~

7. ~~Starting the quest Call of Earth (1516) should enable Earth Sapta (1463) and completing the quest Call of Earth (1516) should disable Earth Sapta (1463)~~

8. ~~Swart (3173) incorrectly starts the quest Call of Water~~

[...]

PR #460, opened 2020-04-14
Fix shaman quests and trainers re #436

NickTyrer:
Fixes (shaman)1, 2, 4, 5, 6, 7 and 8 in issue #436


ratkosrb:
The comments need to have a space after -- or it throws a syntax error.

PR #477, opened 2020-04-24
Fix sql errors

NickTyrer:


sql/migrations/20200418172253_world.sql
```diff
@@ -30,21 +30,6 @@ INSERT INTO `npc_trainer` (`entry`, `spell`, `spellcost`, `reqskill`, `reqskillv
 -- remove starshards (all ranks) and elunes grace (all ranks) from Maxan Anvol (1226)
 DELETE FROM `npc_trainer` WHERE `entry` = 1226 AND `spell` IN (19351, 19352, 19353, 19354, 19355, 19356, 19358, 19359, 19360, 19361);
 
--- add desperate prayer (all ranks) to Maxan Anvol (1226)
```
ratkosrb, line 33: Why is that being removed? Trainer should not have those spells or what?

NickTyrer: He already has those spells, it was a mistake to include them.

PR #524, opened 2020-05-07
Added make_migration.py 

tonymmm1:
Added a python program to create migrations that should be cross platform.


GrenderG:
What's wrong with the bash script?

tonymmm1:
@GrenderG I was noticing return carriage issue where ^M was present at the end of every line but this is mostly just to make cross-platform script and for it to be easily changed in the future.

PR #527, opened 2020-05-07
Add conditions for alchemy trainers gossip

NickTyrer:
Adds the following conditions:
1. require skill level 50 to access training from expert trainers
2. require skill level 125 to access training from artisan trainers
3. require skill level 200 to access training from master trainers

ISSUES
1. ~~Alchemist Narett, Kylanna, Rogvar, Serge Hinott and Kylanna Windwhisper - are missing their entry in gossip_menu_option so cant assign condition yet~~
2. ~~Kylanna, Rogvar, Serge Hinott and Kylanna Windwhisper share the same gossip_menu_id, as they are different ranks of trainers can't assign a condition yet.~~


NickTyrer:
1. Alchemist Narett has a training gossip in game
aaa
2. He has a gossip_menu_id assigned
vvvv
3. But no gossip_menu_option
dddd

How does he get his "Train me." gossip in game?

ratkosrb:
Training button is added automatically if there is not one in the gossip menu. It's because he has the trainer npc flag in `creature_template`.

NickTyrer:
@ratkosrb, is it worth implementing this at the moment as the menu_id's are subject to change?

ratkosrb:
I corrected menu ids based on cmangos, which should have data from sniffs.

NickTyrer:
@ratkosrb have aligned this with the latest gossip pr and found a couple of missind conditions.

sql/migrations/20200507205634_world.sql
```diff
@@ -0,0 +1,30 @@
...
+-- tally berryfizz
+UPDATE `gossip_menu_option` SET `condition_id` = 361 WHERE `menu_id` = 4123 AND `option_id` = 5;
+
+-- yelmak
+UPDATE `gossip_menu_option` SET `condition_id` = 361 WHERE `menu_id` = 4126 AND `option_id` = 5;
+
+
+-- require alchemy skill of 225 to unlock master alchemist training gossip
+-- kylanna windwhisper
+UPDATE `gossip_menu_option` SET `condition_id` = 362 WHERE `menu_id` = 4133 AND `option_id` = 5;
```
ratkosrb: Condition 362 is for 125 skill. Did you mean to use 363 which is for 200 skill? This npc is a Master Alchemist.

NickTyrer: @ratkosrb I believe you need to have 225 skill to unlock master alchemist gossip, I may need to add a condition?

PR #528, opened 2020-05-08
fixed trivial mistake in .travis.yml

tonymmm1:


tonymmm1:
This might fix the window build but it fixes the two lines

PR #537, opened 2020-05-10
Fixed page_text of The Third Troll Legend.

GrenderG:
It was using page_text 324, which was wrong (it was the last page of _Sully Balloo's Letter_).


ratkosrb:
WoW Classic client files disagree.
https://wow.tools/dbc/?dbc=itemsparse&build=1.13.5.34138#page=1&search=2007

PageID: 324

GrenderG:
You are right, then, what's wrong, is the page_text entry. Sully Balloo's Letter should be using a different page_text entry, and text of entry 324 should be the same text as 325. Some entries are shifted then, we must find the correct page_text entries.

PR #549, opened 2020-05-16
Add movement scripts to NPC's

NickTyrer:
This PR adds movement scripts to kogan forgestone and tognus flintfire (currently they are static).
proof https://youtu.be/LnxK0_FVHLI
 - I have restored the animation of the mace moving between the furnace and the anvil.
 - I can't add this functionality to the shield as the game object is missing ontop of the anvil.


Daribon:
Script actually seems to be broken in classic.
As I remember it from retail an object should actually be spawned which they hammer on. Another object should spawn in the forge.

I used to have a video showing this, but it appears I've deleted it.

ratkosrb:
Yes there are a number of broken scripts I've noticed. For example the workers in Sentinel Hill walked around and chopped wood during the demo, but they don't do on Classic anymore. There is a guard dueling event in Theramore that only works after a server restart, and then gets stuck a few minutes later.

Daribon:
WowT_dD9jKY6tSU
WowT_PwlB6uw8Jd
It works on BfA, however it's a bit buggy. The timing is off at times, where the object will despawn way before the NPC is finished hitting the object.

Now you might claim that they added the gameobjects to the script in cataclysm, but I have the gameobjects from a 2007 db so it certainly existed in vanilla. (unless they added it in TBC which makes no sense.)

About the theramore event, it's extremly bugged and have been for at least 5 years. I'm amazed they have never fixed it. It constantly gets stuck even in BfA. If you are lucky you might see it go through a full event run once every 4 days or so. Used to have a full recording of it, but seems to have been deleted as well.

NickTyrer:
This could be implemented with tognus hovever the despawn gameobject command doesnt seem to work

PR #550, opened 2020-05-16
Add summon to wren darkspring

NickTyrer:
wren darkspring should summon an imp this PR adds this.
proof
Untitled


NickTyrer:
@ratkosrb I cant get the imp postitiond correctly, i have set the coords bit they seem to be ignored?

ratkosrb:
You can't specify position for pets.

NickTyrer:
Should I leave this one then? The imp spawns in the stairwell and looks a bit janky?

ratkosrb:
It does not matter if the pet is perfectly positioned, as long as the npc has the pet it should have.

PR #552, opened 2020-05-17
hourly bell - new spawn points and code improvements

schell244:
Fix sound not beeing played in some main city areas:
- sound is now played zone wide instad of area wide in main cities (main cities have many area ids, but only one zone id, so previously there have been many spawns in each city)

Reimplemented all spawn points of horde and alliance bell gameobjects:
- all main cities have only one spawn
- added missing spawns to: Undercity, Scarlet Monastery, Lakeshire, Southshore

Removed static zoneId parameter of method PlayDirectSoundToMap()

### DIFF
--- a/sql/migrations/20200517152850_world.sql
+++ b/sql/migrations/20200517152850_world.sql
@@ -1,0 +1,47 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200517152850');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200517152850');
+-- Add your query below.
+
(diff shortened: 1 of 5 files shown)


Kittnz:
And these spawnpoints are from sniffs? cuz the ones that were spawned are from sniffs if i recall and you are not using all those spawn points this is a hack

schell244:
Thank you for your feedback - no, the new spawn points are not from a sniff. The spawn points I added cover all old spawn locations, including the new ones I mentioned.... But if the old spawn points are actual sniff spawns, then we should probably stick to them. Also playing sound in cities zone wide would be redundat in this case, since old spawns locations will cover (almost?) all city areas. So to get the missing spawn location, there is another sniff required I guess.

Kittnz:
Correct, you will notice a lot of these gobs are spawned in the bell areas like churches on top and other area's, in cities they are just all over the place to make sure everywhere you can hear the bells, the only issue we had at the moment was wrong sound id's used for zones/areas, i didn't know scarlet monastery had bells rinning etc.. i might parse all my sniffs again and get all the bells and check if we are missing a few spawns, i appreciate the work tho! 👍

PR #558, opened 2020-05-18
Add movement scripts to Gol'Bolar Quarry

NickTyrer:
This PR adds movement scripts to :
- cook ghilm (also fixes his equiped item)
- frast dokner
- kazan mogosh
- miner grumnal

### DIFF
--- a/sql/migrations/20200518160736_world.sql
+++ b/sql/migrations/20200518160736_world.sql
@@ -1,0 +1,60 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200518160736');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200518160736');
+-- Add your query below.
+
+
+-- add cleaver to cook ghilm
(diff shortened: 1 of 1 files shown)


sql/migrations/20200518160736_world.sql
```diff
@@ -0,0 +1,60 @@
...
+(1698, 3, -5712.14, -1596.03, 383.204, 0.461232, 60000, 0, 0);
+
+-- pathing and scripts for kazan mogosh
+UPDATE `creature` SET `movement_type`= 2 WHERE `id`= 1237;
+INSERT INTO `creature_movement_template` (`entry`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `waittime`, `wander_distance`, `script_id`) VALUES
+(1237, 1, -5671.19, -1577.15, 383.329, 3.4383, 40000, 0, 0),
+(1237, 2, -5668.92, -1574.36, 383.204, 100, 0, 0, 0),
+(1237, 3, -5668.73, -1571.90, 383.204, 100, 0, 0, 0),
+(1237, 4, -5665.08, -1567.92, 383.204, 0.753848, 50000, 0, 0);
+(1237, 5, -5665.08, -1567.92, 383.204, 0.753848, 0, 0, 1237);
```
alexkulya: Here you can use a single cmove script ID: 11 (generic run) to return to the starting point in a runing state. If it does not start, then try to put the script on 2 points.

NickTyrer: Thanks, that worked perfectly.

PR #564, opened 2020-05-22
Add movement scripts to haggle

NickTyrer:
This PR adds movement scripts to haggle and adds some random text he was missing.

proof https://youtu.be/bx42SjSDbPQ


ratkosrb:
9141 - %s sifts through the trash.
9147 - %s yawns.

Blizzard does not hardcode the npc names in the text, they use %s which prints source name.

PR #575, opened 2020-05-25
Fixed deleteitem command and added merged updates to .gitignore.

GrenderG:
`deleteitem` command was not properly showing the amount of deleted stacks.


ratkosrb:
You should add logon and login db updates too.

GrenderG:
Done!

PR #578, opened 2020-05-28
Fixed Stormwind Rendezvous and The Great Masquerade quests.

GrenderG:
· Added missing events for Stormwind Rendezvous quest.
· Fixed faction of Windsor's horse.
· Now orientation of waypoints are being taken into account.
· Added missing emotes / corrected timing in many The Great Masquerade steps.
· Fixed equipment id of Onyxian Guards.
· Now Lady Katrana is correctly switched with Lady Onyxia, this also fixes the issue of her having the wrong scale when respawning.
· Preventing Onyxian Guards and Windsor from respawning just after the ending.


Daribon:
Keep in mind that this event is massively bugged on classic.
On high population realms it tends to break 1/3 of the time.

GrenderG:
Finished. Only this Flare missing which I'm not capable of finding: https://youtu.be/Pv0wCVDRu0I?t=321

PR #598, opened 2020-06-09
Update and add NPC movement scripts - Hillsbrad Foothills

NickTyrer:
This PR adds some new movement scripts and updates existing ones using waypoints from sniffs where possible.


sql/migrations/20200609200850_world.sql
```diff
@@ -0,0 +1,222 @@
...
+(2436, 60, -812.481, -603.175, 14.6819, 100, 0, 0, 0),
+(2436, 61, -812.444, -612.786, 13.9281, 100, 0, 0, 0),
+(2436, 62, -818.662, -623.293, 13.1924, 100, 0, 0, 0),
+(2436, 63, -830.559, -634.047, 14.1155, 100, 0, 0, 0);
+
+
+-- tarren mill
+-- guards should roam
+UPDATE `creature` SET `movement_type` = 1, `wander_distance` = 9 WHERE `guid` IN (15558, 15538, 15535, 15355, 15533, 15354, 15353, 15307, 15306, 15303, 15515, 15252, 15541);
+
```
alexkulya: 15252 maybe should be 15282?

NickTyrer: You are right, thanks for reviewing.

PR #607, opened 2020-06-13
Change std::set to std::vector in mapextract to avoid segfault on Linux

ellingtonsantos:
Change std::set to std::vector in mapextract to avoid segfault on Linux, GCC 10.1.0


ellingtonsantos:
I realize all problems are caused by GCC -O3 flag. All references to std:set are causing segfault. 

I will investigate the cause and post a more appropriated solution.

PR #609, opened 2020-06-16
Removes GCC flags that causes segmentation fault

ellingtonsantos:
Removes GCC flags that causes segmentation fault (core dumped) if binary is compiled with optimizations.


ratkosrb:
I'm not sure removing the flag is actually fixing the segmentation fault. I think its just preventing it from crashing.

```
This option allows trapping instructions to throw C++ exceptions. It allows hardware signals generated by trapping instructions to be converted into C++ exceptions and caught using the standard C++ exception handling mechanism. Examples of such signals are SIGFPE (floating-point exception) and SIGSEGV (segmentation violation). 
```

ellingtonsantos:
Well, I am running this server right now without issues.

With this flag and compiled with -O3, I got segfault always std::set was used.

If I keep the flag but changes to -O0 it works fine too.

ulfgebhardt:
@necropresto What I do not understand about the PR is the SQL changes included in it. Could you elaborate why it is needed or split up the PR in two? <3

ratkosrb:
I think we concluded this was compiler bug, so I'm closing it.

PR #619, opened 2020-06-20
Add Westfall Lumbermill Event

NickTyrer:
This PR restores the event at the lumbermill in westfall


Kittnz:
is this the one at sentinal hill?

NickTyrer:
yes

PR #629, opened 2020-06-22
Update Varimathras orientation and size.

Daribon:
Varimathras orientation was wrong in vmangos.
This corrects those values to be blizzlike.

Orientation source is sniff data.
Source for scale number is wowemu, confirmed that it was correct in classic.


GrenderG:
You forgot a `;`. :P

Daribon:
Fixed.

PR #630, opened 2020-06-22
Update minimum patch value for Apothecary Keever to build 1.7.1.

duckbutt55:
Update minimum patch value for Apothecary Keever to build 1.7.1.

### DIFF
--- a/sql/migrations/20200622152622_world.sql
+++ b/sql/migrations/20200622152622_world.sql
@@ -1,0 +1,2 @@
+INSERT INTO `migrations` VALUES (20200622152622);
+UPDATE `creature` SET `patch_min`='5' WHERE `guid`=31870;


ratkosrb:
1. You are not following the valid migration format. You must use the provided batch script to create migrations, otherwise the sql will throw an error if attempting to run it a second time, and thus prevent easy db updates. The system is designed so that you can just run the combined migrations file "world_db_updates.sql" and it will only apply the migrations you are missing. Not including the procedure code in the sql prevents this from working, as it will attempt to insert the already existing migration id into the migrations table a second time, throwing an error and causing the execution of the whole file to halt, thus forcing you to manually run all missing migrations one by one, which is a colossal waste of time.

2. This npc was not added in patch 1.7. He was there since launch, as can be seen on Allakhazam. And his low creature id would indicate he was probably there in beta too.
https://web.archive.org/web/20041115152012/http://wow.allakhazam.com/db/mob.html?wmob=5734

duckbutt55:
I see exactly what you mean now, sorry for the confusion. I did notice later on that he was referenced all the way back through Allakazam. Closing this for an attempted correction on the script.

PR #631, opened 2020-06-23
Corrected script for Apothecary Keever.

duckbutt55:
Previous configuration caused crash in  1.6.1 client. Tested on both 1.6.1 and 1.7.1, with no crash. Stand state values were incorrect - Keever now stands and kneels as intended.


ratkosrb:
Good job.

duckbutt55:
Thank you! Now has been tested in 1.8.4, 1.9.4, 1.10.2, 1.11.2, and 1.12.1.

PR #632, opened 2020-06-24
Multiple improvements for Corrupted Ashbringer SM event.

GrenderG:
Fixed:
- Wrong equipment for Highlord Mograine (shouldn't use any).
- Missing emotes.
- Wrong texts.
- Wrong timers.
- Some clean up in general.

Still needed:
- Highlord Mograine's final waypoint appears to be guessed, I have added a TODO to fix it once we have the correct position.
- The Cathedral door should be automatically opened, it wasn't working before and I haven't managed to make it work yet.
- Reactions from SM mobs are not working in a Blizzlike way, they should first look at the player, then slowly kneel (not just change the stand state) and then say the random text. Now everything happens at the same time, it should exist some delay between these actions.

Resources:
- https://www.youtube.com/watch?v=24iX8QD3pRA
- https://www.youtube.com/watch?v=QViNP4cnPyM
- https://www.youtube.com/watch?v=m3E6W6-rRGA

### DIFF
--- a/sql/migrations/20200623153117_world.sql
+++ b/sql/migrations/20200623153117_world.sql
@@ -1,0 +1,32 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200623153117');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200623153117');
+-- Add your query below.
+
+-- Highlord Mograine Transform should not use any weapon
+-- https://www.youtube.com/watch?v=QViNP4cnPyM
+-- https://www.youtube.com/watch?v=m3E6W6-rRGA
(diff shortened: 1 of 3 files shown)


ratkosrb:
You should also delete the now unused script_text entries. Try setting go state in order to open the door.
https://github.com/vmangos/core/blob/development/src/game/Objects/GameObject.h#L706

GrenderG:
> You should also delete the now unused script_text entries.

Done, deleted.

>  Try setting go state in order to open the door.

That worked, but sadly seems like the door is not loaded because of the distance when you enter the instance, so it's still not working, it only works if you are near the door.

ratkosrb:
You can do it in OnObjectCreate then. Also don't save pointer to objects, save guid and then use GetGameObject.

GrenderG:
> You can do it in OnObjectCreate then. Also don't save pointer to objects, save guid and then use GetGameObject.

I tried that, but the door should only be opened if the event is active, and the code to check if the event should be activated runs once every 5 seconds, so `OnObjectCreate` can be called before the event is activated event if it's going to be.

PR #642, opened 2020-06-26
Removed non-vanilla auth codes

ghost:
Removed non-vanilla auth codes:
- There was no Battle.net accounts support (only WoW ones) before 3.3.5 patch in WoW
- There was no 10-day free trial before some TBC patch

### DIFF
--- a/src/realmd/AuthCodes.h
+++ b/src/realmd/AuthCodes.h
@@ -78,16 +78,12 @@ enum AuthResult
     WOW_FAIL_PARENTCONTROL          = 0x0F,                 ///< Access to this account has been blocked by parental controls. Your settings may be changed in your account preferences at <site>
     WOW_FAIL_LOCKED_ENFORCED        = 0x10,                 ///< You have applied a lock to your account. You can change your locked status by calling your account lock phone number.
     WOW_FAIL_TRIAL_ENDED            = 0x11,                 ///< Your trial subscription has expired. Please visit <site> to upgrade your account.
-    WOW_FAIL_USE_BATTLENET          = 0x12,                 ///< WOW_FAIL_OTHER This account is now attached to a Battle.net account. Please login with your Battle.net account email address and password.
     //WOW_FAIL_OVERMIND_CONVERTED
     //WOW_FAIL_ANTI_INDULGENCE
     //WOW_FAIL_EXPIRED
-    //WOW_FAIL_NO_GAME_ACCOUNT
     //WOW_FAIL_BILLING_LOCK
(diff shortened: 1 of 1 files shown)


Daribon:
Should remove all those commented lines at the bottom as well, they all appear to be from a later version.

WOW_FAIL_TRIAL_ENDED should be removed as well, trial wasn't added until TBC.
Even found the old european page for it, which still appears to be up: http://media.wow-europe.com/bctrial/splash_en.html

WOW_FAIL_LOCKED_ENFORCED is most likely from another expansion too, I've been unable to find it anywhere on old emulators.

ghost:
> Should remove all those commented lines at the bottom as well, they all appear to be from a later version.
> 
> WOW_FAIL_TRIAL_ENDED should be removed as well, trial wasn't added until TBC.
> Even found the old european page for it, which still appears to be up: http://media.wow-europe.com/bctrial/splash_en.html
> 
> WOW_FAIL_LOCKED_ENFORCED is most likely from another expansion too, I've been unable to find it anywhere on old emulators.

Try to send WOW_FAIL_LOCKED_ENFORCED to client. If it won't work, I will remove it. TrinityCore also [defines](https://github.com/TrinityCore/TrinityCore/blob/3.3.5/src/server/authserver/Authentication/AuthCodes.h) some of commented opcodes.

Are you sure anti-indulgence system was not in vanilla WoW? IIRC, it is mandatory in China and WoW was already released here.

ghost:
What is WOW_FAIL_OVERMIND_CONVERTED? Is it about Battle.net account conversion?

Daribon:
Some WoTLK code for sure.

ghost:
I have removed these opcodes. I'm not sure about anti-indulgence.

I have tested all opcodes on 1.6.1 and 1.12 clients, so this pull request is now ready.

ratkosrb:
The login server is meant to support multiple versions. so removing definitions for newer versions does not make sense.

GrenderG:
But this project should only support up to 1.12.3, no?

ratkosrb:
The realmd is multi expansion and can be used with other cores.

GrenderG:
I know, but I don't see any reason to support post Vanilla in vMaNGOS' realmd.

ratkosrb:
It already supports it. Why remove useful functionality that already exists? It allows you to run servers for multiple expansion on the same realmlist. You could have vmangos vanilla + cmangos tbc + trinitycore wotlk all running with same account database.

GrenderG:
I got your point now. Makes sense. :+1:

PR #644, opened 2020-06-27
Remove trial opcodes

ghost:
It looks like trial accounts were added in TBC


ratkosrb:
Please verify your changes are correct before making pull requests. I just checked and both error messages exist in the 1.12 client.

ghost:
@ratkosrb that's weird as trial accounts do not exist in vanilla. Or do they? At least, trial auth code doesn't work.

GrenderG:
They existed.

Gamemechanicwow:
image

iirc.. at some point there was also 10 days trial giftcodes with physical bought gamecards.

Daribon:
Aha, didn't realize

ghost:
In which version were these opcodes added?

PR #667, opened 2020-07-09
Added Mocking Blow Skill

mnemonic-re:
- Added Mocking Blow
- Added Mocking Blow to Partybots


mnemonic-re:
Already exists!

ulfgebhardt:
Hey @mnemonic-re - i could not figure out where the mocking blow is implemented for the bots - could you elaborate where you found it? <3

ratkosrb:
Taunt spells are used generically, and not by name.
https://github.com/vmangos/core/blob/development/src/game/PlayerBots/PartyBotAI.cpp#L761

Issue #641, opened 2020-06-26
Billing not implemented


ghost:
**Current behavior**: Billing (10-day trial (non-classic), subscription, game time, etc) not implemented

**Expected behavior**: Billing should be implemented

**Commit hash**: 5cb6957

ratkosrb:
This sounds like it goes beyond the scope of this project. We are only recreating the game content. I see no good reason for emulating Blizzard's payment system. I certainly wouldn't want to encourage people to make pay to play servers.

ghost:
> This sounds like it goes beyond the scope of this project. We are only recreating the game content. I see no good reason for emulating Blizzard's payment system. I certainly wouldn't want to encourage people to make pay to play servers.

Without this payment system, the game is not _blizzlike_. Also, IIUC it is not that difficult to implement and I might try to do it (isn't it like a ban with different message sent?), but core takes too much time to build and I don't know is there proper way to update it quickly (please answer in #647 if there is).

Also, were there any warning messages on game time or just disconnection when it ends?

ratkosrb:
Feature request denied. Implementing game time management is tantamount to endorsing for profit servers. If you are going to work on such a thing, then do it in your own fork. This functionality will not be implemented in vmangos.

ReynoldsCahoon:
There could be legitimate reasons to want this outside of implementing a subscription model.

Weren't there 'scheduling' options for accounts? I specifically remember my mother setting this on my younger brother's account. For example if a parent wanted to create a server for their children to play on, and restrict access during certain times or limit the amount of time played per day, would this not be handled through a similar system?

ratkosrb:
go away, ghost

PR #673, opened 2020-07-11
Add initial game time support

ghost:
This pull request adds initial game time support.

Only logon checks are currently implemented.

What do these each of [these](https://github.com/vmangos/core/blob/development/src/game/World.cpp#L271-L273) packets mean? Which flags are available for second one (in client GlueStrings.lua I see there are IGR, subscription, free trial and personal time, there is also support for fallback payment) and what numbers do they have?

### DIFF
--- a/sql/logon.sql
+++ b/sql/logon.sql
@@ -34,6 +34,7 @@ CREATE TABLE IF NOT EXISTS `account` (
   `last_pwd_reset` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
   `online` tinyint(4) NOT NULL DEFAULT '0',
   `expansion` tinyint(3) unsigned NOT NULL DEFAULT '0',
+  `game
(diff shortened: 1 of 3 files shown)


ratkosrb:
This pull request will not be accepted for the reason stated here: https://github.com/vmangos/core/issues/641

ghost:
Game time management is World of Warcraft client feature, not just server one. Why does it even work with zero game time set (even on latest retail patch)?

I still need billing packets to be explained.

ghost:
BillingTimeRemaining is remaining billing time in seconds.

BillingTimeFlags are available [here](https://github.com/TrinityCore/WowPacketParser/blob/master/WowPacketParser/Enums/BillingFlag.cs).

BillingTimeRested is anti-indulgence time in seconds (increased when logged in, decreased when logged out).

PR #664, opened 2020-07-06
Update and add NPC movement scripts - STV

NickTyrer:
This PR adds some new movement scripts and updates existing ones using waypoints from sniffs where possible.

PR #689, opened 2020-07-20
Update and add NPC movement scripts - WPL

### LINKED ISSUE #637: NPC Bug Scarlet Lumberjack
**Current behavior**: 
The NPC's are static in northridge lumber camp

**Expected behavior**: 
Some of the mobs should have the chop wood animation see:
https://youtu.be/laHIvE5hH58

**Client version**: 
1.12.1
**Commit hash**: 
ab11bde
**Operating system**: 
Win 10

NickTyrer:
Fixes #637


ratkosrb:
Does setting talk emote in creature_addon actually do something? Because this is supposed to for emote states, not one-shot emotes.

NickTyrer:
@ratkosrb yes it works.

ratkosrb:
Do they really repeat it over and over with no delay on classic?

NickTyrer:
No, they have a delay. Is there any way to set this through creature_addon or do I need to add an event?

ratkosrb:
No, you have to use EventAI. Are the npcs who do talk emotes in pairs? In this case you can have a OOC timer with NEARBY_CREATURE condition assigned, so they only talk when they have a buddy next to them.

NickTyrer:
Could I implement it like I did here (https://github.com/vmangos/core/pull/664/files# LINE 995)?

ratkosrb:
Yes, a guid condition works too.

PR #692, opened 2020-07-21
Fix Westfall Lumbermill Event

NickTyrer:
This PR adds the creature_movement_script to the correct waypoint and removes an unneeded waypoint.


ratkosrb:
The last waypoint is a delay before repeating the path. You had it made so the delay is in EventAI, but this is not the proper way to do it. You should not assign EventAI to a NPC only to have a timer that starts waypoints periodically. You don't need EventAI here at all. You just need to move the delay to a waypoint. That's what I did.

NickTyrer:
@ratkosrb OK will do that in future. The script has to be on the last waypoint otherwise the npc's orientation will be wrong.

ratkosrb:
Does the script execute correctly if it is on the last wp? I was afraid it would execute after the delay is over, instead of when arriving.

NickTyrer:
@ratkosrb It executes correctly with this PR.

PR #698, opened 2020-07-22
Add all known account flags

ghost:
See https://github.com/TrinityCore/WowPacketParser/blob/master/WowPacketParserModule.BattleNet.V37165/Enums/GameAccountFlags.cs


Daribon:
Many of those are obviously from later expansions.

ratkosrb:

PR #700, opened 2020-07-23
Fix patch downloading

ghost:
This pull request makes patch downloading actually work by sending patch itself after its info


Daribon:
Can confirm this fixes the patch downloading. Even works on the release patch 1.1.1.
realmd_5FEIwalo93
WoW_CLxcP3T17Q


I've spotted some very odd error with realmd crashing 5/6 times when loading patch info for the first time.
Loading patch info from ./patches/4442enUS.mpq
It crashes directly after this, almost all the time for me. Solution for me so far has been to just boot up realmd several times until it works.
This happened to me before this fix as well, so the crash is not something caused by this fix.

ratkosrb:

PR #701, opened 2020-07-23
Remove non-vanilla player flags

ghost:
Comments say they were added in further versions. There were no sanctuaries until TBC.

Won't work without #704.


ratkosrb:

PR #702, opened 2020-07-23
Remove comment on nonexistent XFER opcodes

ghost:
If these opcodes are from pre-release WoW, remove them

### DIFF
--- a/src/realmd/AuthCodes.h
+++ b/src/realmd/AuthCodes.h
@@ -35,7 +35,6 @@ enum eAuthCmd
     CMD_REALM_LIST                  = 0x10,
     CMD_XFER_INITIATE               = 0x30,
     CMD_XFER_DATA                   = 0x31,
-    // these opcodes no longer exist in currently supported client
     CMD_XFER_ACCEPT                 = 0x32,
     CMD_XFER_RESUME                 = 0x33,
     CMD_XFER_CANCEL                 = 0x34


Daribon:
I think I've spotted CMD_XFER_ACCEPT when downloading patch from server:
realmd_5FEIwalo93
Unless I'm making some huge mistake here, unauthorized command 50 length 1 is CMD_XFER_ACCEPT.

I also believe I've seen CMD_XFER_CANCEL, it occurred when I already had the patch downloaded but not installed and logging in, which crashed the client. I might have screwed something up serverside which made the client crash, but still the client did send command 52.
realmd_sYHsFq9cgM

Digging into old mangos code, I found the commit where the xfer code was commented out: https://github.com/cmangos/mangos-classic/commit/9445740f08f34b8c1ea9ce96c0ca1316a13b015f#diff-610897f4e08b4d6d0a2580e8f87fa32a. It was commented on July 13, 2010, that's patch 3.3.5 WoTLK.

Finally, I've also managed to make CMD_XFER_RESUME show up, it shows up if you don't download everything 100% and then restart the client. I believe CMD_XFER_RESUME is broken, since every time I cancelled the download, you get corrupted patch error and then the patch deletes itself.
realmd_qVQlplV8YC

Anyways, in the emulator Ludmilla the following was documented in their implementation of patch sending: 
CMD_XFER_ACCEPT = Client agreed to receive new patch. 
CMD_XFER_RESUME = Client want to continue receiving of patch from specified position.
CMD_XFER_CANCEL = Client want to abort receiving of the patch.

ratkosrb:

PR #721, opened 2020-07-26
Replace 'classic' with 'vanilla' in AuthSocket.cpp

ghost:
"Classic" (not really) != vanilla


ratkosrb:

PR #724, opened 2020-07-26
Add promotion codes support

### LINKED ISSUE #655: Secret pet codes
Implement secret pet codes, like some murloc eggs.

ghost:
This pull request adds support for promotion codes (also known as secret murloc pet codes).

Closes #655

### DIFF
--- a/sql/migrations/20200726215404_world.sql
+++ b/sql/migrations/20200726215404_world.sql
@@ -1,0 +1,26 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20200726215404');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20200726215404');
+-- Add your query below.
+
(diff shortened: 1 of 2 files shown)


ghost:
For some reason client doesn't prompt for code second time in either case. I'm not sure if the packet is valid.

You may also choose to mark used codes (e.g. using database column `used`) instead of deleting them.

ratkosrb:
1. Do not use db queries in packet handlers. Queries are one of the slowest things you can do. A malicious person could send the gossip opcode over and over and lag the server if you do this. Load the table on startup and then only check if the key exists in the set/map.
2. It is not permitted to add gossip menu options without a `broadcast_text` id assigned. Find the correct text in the `broadcast_text` table.
3. The current table structure for `promotion_codes` is inadequate. You have no way of distinguishing what promotion the code is for. Remove the `id` field as it is not needed, and replace it with a `group_id` field that would allow you to group together codes that are for the same promotion. Both the group id and code fields should be part of the primary key.
4. I think you'll have to refactor how the `box_coded` field works, and use it to store the promotion code group id.
5. Please follow the code style of the rest of the code, the opening and closing { } brackets should be on separate lines.
6. Enclose strings in the migration in ' instead of " double quotes.
7. The giving of the reward should not be a hardcoded spell cast. Assign a `gossip_script` id to the option.

ghost:
The only available promotion with codes in vanilla is blue murloc egg. Why should it be grouped then?

ratkosrb:
Seems like a better design to be able to specify what promotion the code is for, instead of tying the whole system to 1 item. Move the whole thing to a c++ creature script for the npc if you don't want to generalize it. There is a pGossipSelectWithCode method in the Script class.

ratkosrb:
I have to close this pull request since the user account is deleted, which means it cannot be updated to address all the issues with it.

PR #735, opened 2020-07-31
Added support for discontinued City Protector titles.

ghost:
City Protector is an apparently discontinued or never-implemented PvP rank that was announced by Blizzard at some point, read more: https://wowwiki.fandom.com/wiki/City_Protector


ratkosrb:
I think it would be better to utilize the `customFlags` field for this, rather than add a whole new field. Just define a new custom flag.

ghost:
Done.

ratkosrb:
The migrations are not needed, just remove them. Also move the "titles" command to the "character" category and rename it to citytitle to make it more obvious what it is.

ghost:
Done.

PR #743, opened 2020-08-02
Update and add NPC movement scripts - Badlands

NickTyrer:


NickTyrer:
Currently missing waypoints for Angor Patrol (only partial are available in current sniff).

ratkosrb:
I'll check it out.

ulfgebhardt:
@NickTyrer Could you obtain the missing data? If not, can we merge the complete parts of it and make an issue for the rest including the incomplete data? <3

NickTyrer:
@ratkosrb no need to sniff this, cmangos has just added lots of movement data to this zone, I will use that.

PR #745, opened 2020-08-04
Fix Westfall and Elwynn Vendors

ratkosrb:


sql/migrations/20200705175656_world.sql
```diff
@@ -0,0 +1,70 @@
...
+
+-- elwynn
+-- Pattern: Blue Linen Vest - Tharynn Bouden
+UPDATE `npc_vendor` SET `maxcount` = 1, `incrtime` = 7200 WHERE `entry` = 66 AND `item` = 6270;
+
+-- Scroll of Strength - Dawn Brightstar
+UPDATE `npc_vendor` SET `maxcount` = 2 WHERE `entry` = 958 AND `item` = 954;
+
+-- Fine Shortbow - Rallic Finn
+UPDATE `npc_vendor` SET `maxcount` = 2 WHERE `entry` = 1198 AND `item` = 11303;
```
ratkosrb: You need to set incrtime here. It cannot be 0 if maxcount is limited.

NickTyrer: Resolved

PR #749, opened 2020-08-04
Fix patch downloading

Simplaf234-ghost:
Just a duplicate of #700 if you don't like deleted accounts. Then I will just leave my account and most likely never return because I decided to quit World of Warcraft.

---

This pull request makes patch downloading actually work by sending patch itself after its info


Simplaf234-ghost:
WTF why is it wrong?

PR #760, opened 2020-08-10
Update and add NPC movement scripts - Teldrassil

NickTyrer:


NickTyrer:
@ratkosrb do you need anything doing here to make this mergable?

NickTyrer:
@ratkosrb waypoints now updated, this should be good to go.

PR #763, opened 2020-08-11
Fix Quest The Shimmering Frond end script and gossip

NickTyrer:
This PR adds the emote none command to the script to prevent the emote work being stuck on.


ratkosrb:
Can you add comments to each command? It's good to have it.

Name - Action

PR #786, opened 2020-09-04
Old NPC spawn pre naxx entrance.

Daribon:
This adds an old creature spawn which was deleted when Naxxramas entrance was added, replacing what was previously in the area.
WoWScrnShot_090420_222607

Source: WoWD-DB


ratkosrb:
I think you want patch_max = 8 here, because it is inclusive. Patch 9 means it will still be there in 1.11 when Naxx is released.

Daribon:
Ah right, will fix that.

PR #787, opened 2020-09-05
Add all patches to realmd

cookreidhabcdefg:
This PR makes realmd work with all client patches, starting from 1.2.


ratkosrb:
1. The core does not currently support those builds, so that's why realmd doesn't either. Patches 1.2.4 to 1.5.1 will be added after core support is officially added to them. Otherwise it would confuse users thinking they can play with those versions.
2. I know who you are. You are banned from the project, and making new accounts does not change that. No contribution from you will ever be accepted.

PR #788, opened 2020-09-06
Fix dual wield penalty when Yellow (Special attack) is queued up

zaphstar:
All off-hand attacks that occur while Heroic Strike is queued should not suffer the dual wield penalty.

Source: https://us.forums.blizzard.com/en/wow/t/wow-classic-%E2%80%9Cnot-a-bug%E2%80%9D-list-updated-feb-19-2020/175887

Offhand attacks that occur while on-next-hit abilities such as Heroic Strike are queued do not suffer the dual wield to-hit penalty.


zaphstar:
After further testing no changes required, closing request.

Issue #806, opened 2020-10-01
[Dungeon] Dire Maul North - Prinz Tortheldrin


schell244:
https://de.classic.wowhead.com/npc=11486/prinz-tortheldrin

After clearing the 5 crystals and killing Immol'thar, Prinz Tortheldrin should be attackable.
Currently he does not get attackable. (2 tests performend)

I think this is the code which should make him attackable(!?): 
https://github.com/vmangos/core/blob/a0bda55d9a310830e1772ba981aa3274107d41d5/src/scripts/kalimdor/feralas/dire_maul/instance_dire_maul.cpp#L315

Maybe the code could be moved to here, since the text emote is always (and instantly) printet after Immol'thar dies:
https://github.com/vmangos/core/blob/a0bda55d9a310830e1772ba981aa3274107d41d5/src/scripts/kalimdor/feralas/dire_maul/instance_dire_maul.cpp#L178

**Client version**: 
1.12.1

**Commit hash**: 
51722403afe426cb737b2a6d4992edf47d3f9045

**Operating system**: 
Win 10

PR #807, opened 2020-10-01
immoltar setdata wasnt called

Gamemechanicwow:
This fix will allow Tortheldrin to be attackable after Immoltar is killed.
#806


ratkosrb:
That is not what the issue was.

PR #838, opened 2020-10-22
Update event_scripts to generic_scripts

NickTyrer:


ratkosrb:
Your migration is from before the table change, so you don't need to update it.

PR #842, opened 2020-10-24
Add Auras to Wickerman Guardian

NickTyrer:


sql/migrations/20201022111130_world.sql
```diff
@@ -9,6 +9,10 @@ INSERT INTO `migrations` VALUES ('20201022111130');
 -- Add your query below.
 
 
+-- Add auras to Wickerman Guardian
+INSERT INTO `creature_template_addon` (`entry`, `auras`) VALUES
+(15195, 12187);
```
ratkosrb: The auras are a string field, not an integer.

sql/migrations/20201022111130_world.sql
```diff
@@ -11,7 +11,7 @@ INSERT INTO `migrations` VALUES ('20201022111130');
 
 -- Add auras to Wickerman Guardian
 INSERT INTO `creature_template_addon` (`entry`, `auras`) VALUES
-(15195, 12187);
+(15195, `12187`);
```
ratkosrb: The ` symbol is for table and field names, not for string. Use the ' symbol.

PR #909, opened 2020-12-16
Respawn Caer Darrow

NickTyrer:
TODO: With the correct auras applied, you cannot see the ghosts unless you have GM mode enabled. Also you caannot access the gossip menus. I am not sure why?

Here is a video for reference: https://youtu.be/kVNuVSdCJBg


ulfgebhardt:
@NickTyrer Is this ready for review? Can you add some screenys/details how to test?

NickTyrer:
@ulfgebhardt The PR is still stuck on the issue mentioned above @ratkosrb is currently looking into it.

ratkosrb:
I'll look into the aura once we do the big parse.

PR #910, opened 2020-12-17
Broadcast texts for wailing caverns

schell244:
Not sure what to do with SAY_BEFORE_MUTANUS?
Previous script text did not have a placeholder ($n), there was "Mutanus the Devourer" harcoded within the string.


ratkosrb:
$n will be replaced with the name of the unit you provide as target to DoScriptText.

schell244:
Currently this text is broadcasted within `Subevent_Phase = 6;` and Mutanus is not summoned at this point (!?) and so can't be used as target - I could move this broadcast to  `Subevent_Phase = 7;` where he's summoned and catch a reference of him there?

schell244:
I've tried the code below and the broadcast worked after Mutanus was spawned, but instead of "Mutanus the Devourer" there was "Unknown" in the text.

```                       
case 6:
SummonAttacker(MOB_MUTANUS_DEVOURER, Position[3][0], Position[3][1], Position[3][2]);
if (Creature* Mutanus = Naralex->FindNearestCreature(MOB_MUTANUS_DEVOURER, 100.0f))
{
DoScriptText(SAY_BEFORE_MUTANUS, m_creature, Mutanus);
}
 Subevent_Phase = 7;
Event_Timer = 2000;
break;
```

ratkosrb:
Can you try with DoSimulateScriptTextForMap ?

schell244:
I have tried the following code:
```
if (Creature* Mutanus = Naralex->FindNearestCreature(MOB_MUTANUS_DEVOURER, 100.0f))
{
    DoOrSimulateScriptTextForMap(SAY_BEFORE_MUTANUS, m_creature->GetEntry(), m_creature->GetMap(), Mutanus);
}
```
Got the following error in server.log:

DoSimulateScriptTextForMap entry 1276 has not supported chat type 0.
(boradcast text type seems to be required to be from type CHAT_TYPE_ZONE_YELL for using DoSimulateScriptTextForMap )

ratkosrb:
Does it have to be a Say, and not a Yell?

schell244:
When using `DoOrSimulateScriptTextForMap()` I don't pass the chat type value as parameter, it gets determined by broadcast_text entry itself. `DoOrSimulateScriptTextForMap()` seems to work exclusive for CHAT_TYPE_ZONE_YELL texts.

schell244:
`DoScriptText` finally worked together with `FindNearestCreature` by calling `FindNearestCreature` with a little delay after mutanus gets spawned (I added another `Subevent_Phase` for that).

PR #912, opened 2020-12-20
Correct Race Req's For Quests

NickTyrer:
This PR adds the correct race req's for quest chains:

https://classic.wowhead.com/quest=6344/nessa-shadowsong
https://classic.wowhead.com/quest=6387/honor-students
https://classic.wowhead.com/quest=6321/supplying-the-sepulcher
https://classic.wowhead.com/quest=6181/a-swift-message


ratkosrb:
Those changes are already in place.

PR #920, opened 2020-12-24
install instructions (linux)

ulfgebhardt:
This includes install instructions and requirements + install instruction for those on linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,55 @@ On arch linux it can be installed with `yay -S ace`
 1. Make build directory: `mkdir build`
 2. Make install directory: `mkdir run`
 3. Go to build directory: `cd build`
-4. CMake: `cmake -DCMAKE_INSTALL_PREFIX=../run -DDEBUG=0 -DUSE_LIBCURL=1 ..`
+4. CMake: `cmake -DCMAKE_INSTALL_PREFIX=../run ..`
 5. Make: `make`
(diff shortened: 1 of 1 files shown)


Daribon:
Shouldn't this be on the wiki instead https://github.com/vmangos/wiki?

ulfgebhardt:
It is common practice to have install instructions in the `README.md` of a repo to get started quickly.
Nevertheless those instructs seem to be missing (at least for linux) in the wiki aswell.

I personally dislike external wikis - it makes no sense to have the documentation separated from the code - it is in general very hard to find the documentation and the distance to the code make it go out of sync aswell - but thats only my two cents.

ratkosrb:
Thanks for trying to contribute, but we already have a linux installation guide in the wiki, which is where guides belong in my opinion. Other wow emulator projects like cmangos and trinity have their installation guides in a separate wiki as well. And the wiki is even linked to by the readme.

https://github.com/vmangos/wiki/wiki/Compiling-on-Ubuntu
https://github.com/vmangos/wiki/wiki/Getting-it-working

And if we were to put the install guide in the readme, why would it be just the linux guide in particular, when most people use windows. It would give the impression this is a linux only project, which is not true. So overall this change does not make sense to me. If you feel our existing install guide is missing something, feel free to update the wiki, that's where guides belong.

Issue #923, opened 2020-12-28
🐛 [Bug Report] NPC in RP-Transition should not give quests


ulfgebhardt:
## 🐛 Bugreport

An NPC enacting some RP-Lines should not give you quests during this RP.

As Example: `Denalan's Earth`. Completing this quest will make the NPC walk over to the flower pot and have some RP lines till he returns  to his post. On Retail-Classic during this transition you cannot take or turn in quests (this ensures everyone has his/her own RP-transition).
In the attached screenshot you can see that between the two RP lines quests have been accepted.

![image](https://user-images.githubusercontent.com/1238238/101294968-e65b9980-381a-11eb-98c1-dacd863b21a6.png)

**This is also an Issue for vmangos.**

Original reference: https://github.com/cmangos/issues/issues/2332
Solution: https://github.com/cmangos/wotlk-db/commit/a8afae616e7e25b6e65ceddb528c055d91901523

### Expected behavior

On retail within the transition its not possible to interact with the NPC till its finished. I cannot tell if ~~this behaviour can be generalized~~ or is quest specific

### Version & Environment

Client Version: "2.4.3" (TBC)

CMaNGOS Repo & Commit Hash: [TBC](https://github.com/cmangos/mangos-tbc/tree/daf52413)

Database Repo & Commit Hash: [TBC](https://github.com/cmangos/tbc-db/tree/96d6b19b)

Operating System: Linux Debian

ulfgebhardt:
This was merged

PR #924, opened 2020-12-28
Denalan's Earth (997) - Improve quest_end_script

NickTyrer:
This PR adds the commands to remove the gossip and questgiver flags before the script runs


ulfgebhardt:
Thank you for doing this @NickTyrer ! <3
This fixes https://github.com/vmangos/core/issues/923

PR #926, opened 2020-12-29
Quivers should not increase haste for ammoless ranged weapons.

ghost:


Daribon:
Can confirm that on classic quivers do not increase speed of for example wands.

Issue #830, opened 2020-10-16
[Quest] Counterattack!


schell244:
**The Quest script needs some rework:**
 ~~use broadcast texts instead of script texts~~ https://github.com/vmangos/core/commit/72bfa1ba86d29d539b3570abb2c40d4627082ebe
- add waypoints to horde guard npc's
- container object [kromzars banner] - currently it can only be looted by one player. I think it should be somehow a "timed summon" and while its summoned/spawed it can be looted from all players who have the quest active
- some creature ai scripts can probaly be moved to DB (npc_kolkar_invader, npc_axe_thrower, npc_warlord_kromzar)

**Reference classic video**:
https://www.youtube.com/watch?v=5k6760g_M9E

**Some comments from wowhead:**
_Don't let the defenders fall.  If you see the NPCs fighting the centaurs, support them; don't let them die.  I think (not positive) that if too many defenders die, you lose the battle and Krom'Zar doesn't come._
_After the battle, he won't leave a corpse.  Instead, look for a banner._ 

**Research TODO:**
What happens when player (or all players within group) who started the quest (-script) get killed?
I currently assume than there should be a timer for each phase - if kill count is not reached whithin a certain time, the script should be interrupted.

https://classic.wowhead.com/quest=4021/counterattack
https://classic.wowhead.com/npc=9456/warlord-kromzar
https://classic.wowhead.com/object=164690/kromzars-banner
https://classic.wowhead.com/item=11227/piece-of-kromzars-banner

It seems that CMangos has properly implemented that script.
https://github.com/cmangos/mangos-classic/blob/master/src/game/AI/ScriptDevAI/scripts/kalimdor/the_barrens.cpp

This was their initial implementation, but it got reworked (see above):
https://github.com/cmangos/mangos-classic/commit/ac8cc7e36730b6a84f469c318ea4c29211a0daf0#diff-978c79d464c9d3a5fc16fa19227dfcc238c34d726114e448cb030d6997647e94

CMangos DB waypoints:
https://github.com/cmangos/classic-db/commit/ee9c247bdc1be2767fcd54da90562fea2d9f73d1

**Client version**: 
1.12.1
**Commit hash**: 
3d73832
**Operating system**: 
Win10

schell244:
broadcast texts for script:

4926 = Beware, $N! Look to the west!
4757 = A defender has fallen!
1254 = %s charges! (Emote)
8439 = Half way there... (Yell)
4922 = Defend the bunkers!
[...]

PR #927, opened 2020-12-30
Quest: Counterattack! - end event when player is dead

schell244:
I have no idea how this is handled on classic - it's meant to be a small improvement for https://github.com/vmangos/core/issues/830

If i'm wrong feel free to close this PR.


schell244:
I'll do some more research first - not happy with this

Issue #499, opened 2020-05-01
LOS check ignores some gameobjects


schell244:
Npcs can start attack/cast spells through fence:
https://classic.wowhead.com/npc=2272/dalaran-theurgist

<img width="714" alt="79684398-4c19f780-8231-11ea-8af3-63f9284e64a3" src="https://user-images.githubusercontent.com/9536270/80803144-b7f54c00-8bb1-11ea-8bbd-e7dd3c6fe69d.PNG">

I did some research about LOS checks and found that some gameobjects are (intentionally) ignored.
https://github.com/cmangos/mangos-classic/pull/239

The bug mentioned above does no longer occur, once you remove the following lines at 
MapTree.cpp, line39-42:

```
        // Nostalrius: pas de LoS pour certains models (arbres, ...)
         if (los && prims[entry].flags & MOD_NO_BREAK_LOS)
            return false;
```

I assume that this is also responsible for many incorrect pulls (through house ceilings, etc...)
More testing needs to be done on this.

**Client version**: 
1.12.1
**Commit hash**: 
04a2826
**Operating system**: 
Win 10

ratkosrb:
Does this fence cause LoS on Classic?

schell244:
Enemies start attack (see player) through fence ->no
Enemies and player can attack through fence -> yes

Daribon:
Almost all .m2 objects ignore spell LoS in classic. I believe doors are the only .m2 objects which do follow LoS.

schell244:
So I guess the correct way would be to distinguish between attack and enemy detection LOS and then decide whether to ignore .m2 objects?

schell244:
I noticed that sometimes also the cursor does ignore LOS. 
For example player can detect/select an enemy by clicking on a wall:

<img width="510" alt="LOS" src="https://user-images.githubusercontent.com/9536270/92573800-935b6000-f286-11ea-98e0-0baae2144348.PNG">

I compared some of the vmap code to CMangos and noticed that VMangos uses an additional flag (appart from M2) for ignoring LOS:

https://github.com/vmangos/core/blob/d8a1dc68cadba45d03046f469dc5150128898354/src/game/vmap/ModelInstance.h#L35-L41

https://github.com/cmangos/mangos-classic/blob/9b0203a8da00d8eaa3794cacd6cd313072c7625b/src/game/vmap/ModelInstance.h#L35-L40

ratkosrb:
[...]

PR #945, opened 2021-01-01
Make object line of sight collision configurable

schell244:
## 🍰 Pullrequest
Some gameobjects (like trees, fences, ...) are intentionally ignored when calculating "line of sight" for spells and attacks. Currently on VMangos those gameobjects are ignored for aggro calculating as well. 

With this PR, it's possible to enable/disable line of sight collision for those gameobjects by editing the mangosd.conf:
`vmap.enableM2ObjectLos `
This settings is disbaled by default.

(Furthermore I've removed a parameter from mangosd.conf/code which was not used at all -> pet.los) 

### Proof
- None

### Issues
- as suggested in https://github.com/vmangos/core/issues/499

### How2Test
- spells/distance attacks:
 get in combat with a mage an walk behind a tree while mage is casting a spell:
vmap.enableM2ObjectLos = 0 -> mage can cast throgh tree
vmap.enableM2ObjectLos = 1 -> mage interrupts cast and walks around tree until player is in line of sight
- aggro:
vmap.enableM2ObjectLos = 0 -> enemies can see player through tree/some walls/fences and start attack
vmap.enableM2ObjectLos = 1 -> enemies cannot see player and don't start attack until player walks in actual line of sight

Good location for testing both settings: .tele dalaran check fence like in picture:
80803144-b7f54c00-8bb1-11ea-8bbd-e7dd3c6fe69d

### Todo / Checklist
- the current default setting (vmap.enableM2ObjectLos = 0) does ignore gameobjects for enemy aggro als well, 
which is probaly incorrect. I assume when claculating los, there should be distinguished between agrro and attack.


ratkosrb:
I've decided to take a different approach to this and fix the linked issue directly, instead of having M2 object LoS be decided by a config option. The proposed changes do not actually solve the problem, as M2 object LoS should only be considered for proximity aggro, not for all LoS checks. It should still be possible to cast spells through fences. A config option would affect all LoS checks which is not what we want.

PR #961, opened 2021-01-02
Fix $[uU] references within ptBR translation files.

ellingtonsantos:
## 🍰 Pullrequest
Fix wrong $[uU] references to $[gG] gender macro within ptBR translation files.

### Proof
- None

### Issues
- Gender dependent text displayed wrong

### How2Test
- Talk to any NPC with gender text

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20201211153803_world.sql
+++ b/sql/migrations/20201211153803_world.sql
@@ -1,0 +1,345 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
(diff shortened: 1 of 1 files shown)


Daribon:
Is this really accurate?
Wowhead lists https://pt.classic.wowhead.com/quest=1339/ as using $[uU] while in your edit you change it to $[gG].
Entire ptBR broadcast text table is parsed from wow classic.
Does the same errors exists in wow classic?

ratkosrb:
The locale data is bruteforced from classic, and is therefore correct. However, Classic has an official Portuguese client, while Vanilla does not. Perhaps $[uU] works in Classic, but it cannot work in any client supported by vmangos, so therefore I would say its permissible to change the data in this situation.

ellingtonsantos:
@ratkosrb is right and I am following this VMaNGOS directive: only official translations from wow classic

Issue #946, opened 2021-01-02
🐞 [Bug] Spell 29519 (Silithyst) should create Silithyst Mound (181597) on Cancel Aura.


Oroxzy:
https://github.com/Oroxzy/core/commit/d89feb7891673cd862ff5da530a7bff31045f032
https://github.com/Oroxzy/core/commit/a99a7fae390d73986e308e499c9536ed63fbeb9b

https://youtu.be/_YpumVmwybc?t=160

Oroxzy:
I also remember where we all farmed dust and put it in front of the machine for people who grinded their ranks.

ulfgebhardt:
As I see you have a fix ready in those two Commits you linked - is that correct @Oroxzy ?
Would you be willing to PR this?

Issue #608, opened 2020-06-15
Spell "Silithyst" has no deceleration effect.


gamelgs:
Spell "Silithyst" has no deceleration effect.

Spell ID 29519

**Client version**: 
1.12.1  5875
**Commit hash**: 
7d0ef5388e402d7e0239c691b164dd005bbfeb5a
**Operating system**: 
win10

ulfgebhardt:
Can you elaborate on this @gamelgs ? <3

Oroxzy:
![silithy](https://user-images.githubusercontent.com/357606/103467629-4d546b80-4d51-11eb-8565-c4afe4b91fb0.png)

idk what he mean.

NickTyrer:
@ulfgebhardt I think he means that you should not be able to go faster than 100% when you have the debuf, so things like rocket boots should not work and maybe they do in vmangos?

Oroxzy:
https://www.youtube.com/watch?v=_YpumVmwybc

can anyone test it on classic? on vmangos i can use Swiftness Potion and also rocket boots. Check also if mounting up gets blocked, or does the player just drop the dust on mount up.

NickTyrer:
@Oroxzy just tested on classic, any speed enhancing buff (pot, shapeshifting, engi) does not work. You get the speed buff, but you do not speed up. For example you can still shapeshift but you do not speed up.

Oroxzy:
Ok thanks @NickTyrer. Now if you deliver the dust you get the buff 29534 (Traces of Silithyst). Check if:

- [x] possible to pickup new dust with this buff? (in vmangos you can)
- [ ] speed limited too? (in vmangos it is)
- [x] possible to pickup a mound? (in vmangos you can)

**if we take a closer look to Traces of Silithyst: https://classicdb.ch/?spell=29534**

**Effect 1: (86) Activate Object (5)**
what does it and which object is meant?
- 181617 (Sillithus Flag Stand, Horde)
- 181619 (ToWoW - Flag Cap Counter, Horde)
- 181603 (Sillithus Flag Stand, Alliance)
- 181618 (ToWoW - Flag Cap Counter, Alliance)
- 181598 (Silithyst Geyser) (you can stand near it and .gobj select, .gobj toggle and it disappears, so maybe its this)
- 181597 (Silithyst Mound)

on .cast 29534 console says:
```
[...]

PR #962, opened 2021-01-02
Spell 29519 (Silithyst) should create Silithyst Mound (181597) on Cancel Aura

Oroxzy:
Spell 29519 (Silithyst) should create Silithyst Mound (181597) on Cancel Aura


ulfgebhardt:
Fixes: https://github.com/vmangos/core/issues/946

Does this also fix #608 ?

PR #968, opened 2021-01-04
Silithus world pvp fixes

Oroxzy:
**fixes:**
#608 

- remove speed limit form Traces of Silithyst (29534).
- adds speed limit to Silithyst (29519).
- some clean up and DB fixes.

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -2732,11 +2732,18 @@ void Aura::HandleForceReaction(bool apply, bool Real)
             player->StopAttackFaction(scarlet_crusade_faction_id);
     }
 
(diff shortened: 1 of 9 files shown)


src/game/Spells/SpellAuras.cpp
```diff
@@ -2732,11 +2732,18 @@ void Aura::HandleForceReaction(bool apply, bool Real)
             player->StopAttackFaction(scarlet_crusade_faction_id);
     }
 
-    if (!apply && player->GetZoneId() == 1377 && GetId() == 29519 && (m_removeMode == AURA_REMOVE_BY_CANCEL || m_removeMode == AURA_REMOVE_BY_DEATH))
+    if (player->GetZoneId() == 1377 && GetId() == 29519)
```
ratkosrb: I would put the GetId check first, cause the other function is more expensive.

Issue #969, opened 2021-01-05
🚀 [Feature] - update mysql querys within core  


schell244:
## 🚀 Feature
To make vmangos compatible with newer mysql versions,
all keywords and reserved words contained in core queries should be escaped.

List of keywords can be found here:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html

### Benefits, Purpose
 compatibility for latest mysql versions

[Update]
- Lot's of queries are escaped with https://github.com/vmangos/core/commit/179f2bdb113f73aaf25ff5846c54789b2adf1002

PR #973, opened 2021-01-05
escaped some core mysql queries

schell244:
## 🍰 Pullrequest
escaped queries which contained some reserved words: account, action, active

### Issues
- relates https://github.com/vmangos/core/issues/969


ulfgebhardt:
fixes #969 (?)

schell244:
There are a lot more keywords to check (https://dev.mysql.com/doc/refman/8.0/en/keywords.html), this just covers a few. 
I'll continue soon and if I find more, I'll open another PR.

PR #976, opened 2021-01-06
Fix patch handler to update intermediate client versions

ellingtonsantos:
## 🍰 Pullrequest
Fix patch handler to update intermediate client version and create new config option to set patches dir.
For example: client 1.11.0.5428 isn't a "valid client". If you have a patch 5428enGB.mpq in patch folder, client will download and apply this patch.

### Proof
- None

### Issues
- None

### How2Test
- Install client 1.11.0 enGB
- Put 5428enUS.mpq into patches dir
- Logon

### Todo / Checklist
- [ ] Allow updates from "valid versions"


src/shared/Progression.h
```diff
@@ -27,8 +27,10 @@
 #define CLIENT_BUILD_1_8_4 4878
 #define CLIENT_BUILD_1_9_4 5086
 #define CLIENT_BUILD_1_10_2 5302
+#define CLIENT_BUILD_1_11_0 5428
```
ratkosrb: Why are you adding 1.11.0 ? We do not support this build. Only the last version of each major patch, not the minor patches.

src/realmd/AuthSocket.cpp
```diff
@@ -1115,137 +1100,117 @@ bool AuthSocket::_HandleRealmList()
...
-                pkt << uint32(i->second.icon);              // realm type
-                pkt << uint8(realmflags);                   // realmflags
-                pkt << name;                                // name
-                pkt << i->second.address;                   // address
-                pkt << float(i->second.populationLevel);
-                pkt << uint8(AmountOfCharacters);
-                pkt << uint8(i->second.timezone);           // realm category
-                pkt << uint8(0x00);                         // unk, may be realm number/id?
-            }
+  if (_build < CLIENT_BUILD_2_0_0)
```
ratkosrb: 1, The tabulation looks weird here, it should be 4 spaces.
2. You've removed the build switch case, does that mean realmd will now let you login with any client version? This is not desired behavior, it should give the wrong version error when trying to connect with vanilla clients that mangosd does not yet support, or will never support. So that means any version older than 1.6 currently, or non-final major patch version (like 1.12.0 instead of 1.12.1 for example).

src/shared/Progression.h
```diff
@@ -27,8 +27,10 @@
...
 #define CLIENT_BUILD_1_11_2 5464
 #define CLIENT_BUILD_1_12_1 5875
+#define CLIENT_BUILD_2_0_0 5610
```
ratkosrb: Don't add TBC builds, again somebody might try to set SUPPORTED_CLIENT_BUILD to it, and we do not support TBC.

src/shared/Progression.h
ellingtonsantos: All non supported build will be catch on patch downloader. If patch isn't available, client will show invalid client version error. I added this build to not user a "magic number" in code.

src/shared/Progression.h
ellingtonsantos: There are code to 2.0 clients to logon. As I said before, it is just to not use "magic numbers" inside code and to identify it.

src/realmd/AuthSocket.cpp
ellingtonsantos: 1. I will fix it.
2. It is need to patch handler work. We need to allow unsupported clients to login so they can get patches. If we block login, as before, client will never accept CMD_XFER_INITIATE. Patch code, in this implementation, looks for update a intermediate version (1.11.0 ou 1.11.1 to 1.11.2, a supported build version).

ulfgebhardt: Can I suggest that we use a make command like `-DPATCHABLE=1` or `-DCUSTOMCLIENT=1`. The default for this should be 0 ofc. This will then not pollute the `CLIENT_BUILD_X_XX_X`  definitions.

It still has to interact with the version check ofc - but you then can use the preprocessor defined in make.

This way both systems are separated from each other and its not a hack

src/shared/Progression.h
ulfgebhardt: I think `CLIENT_BUILD_2_0_0 ` should be defined as the actual TBC Build number, but not available in the make so people don't get the wrong idea that it actually is supported. All custom versions should be between `CLIENT_BUILD_1_12_1 = 5875` and `CLIENT_BUILD_2_0_0 = ???? `.

ellingtonsantos: I will fix CLIENT_BUILD_2_0_0 to right one: 6080. I put an alpha version as mistake.

ellingtonsantos: My reference: https://archive.org/download/World_of_Warcraft_Client_and_Installation_Archive/ISO/

src/realmd/AuthSocket.cpp
ellingtonsantos: At this moment, my code only updates intermediate versions if mpq patch is available. If not available, it gives a invalid version error as expected. In future I want to implement code for custom builds and I will follow your instructions.

src/shared/Progression.h
ellingtonsantos: I deleted Progression.h entries. I didn't know it is generated for cmake and it is in .gitignore.
Now my code passes on github's checks.

ulfgebhardt:
Awesome - this should make it easier to have custom content servers <3

Issue #866, opened 2020-11-20
"Haunting Spirits" curse in SFK


T1ti:
**Current behavior**: 
-The spawnrate of the Spirits is extremly high, I got **20** Haunting Spirit spawns over 5 minutes.
-The spirits drop money
**Expected behavior**: 
https://youtu.be/npays0yEpoE?t=654 video from classic
-I saw **4** spawns over 5 minutes.
Spawns can be seen at 35, 145, 195, 200 seconds after curse application, (yes a spawn exactly 5sec after another one).
-spirits don't seem to have any loot.

**Steps to reproduce**:

1. let a haunted servitor cast the curse on you

**Client version**: 
1.12

Issue #865, opened 2020-11-20
"Haunting Phantoms" curse in Stratholm


T1ti:
**Current behavior**: 
The spawnrate of the phantoms is pretty high, I got 7 phantoms spawns over 5 minutes.
Also I noticed a phantom always appears when the curse is applied.(maybe not always but it happened every time to me)
The phantoms have loot , insane loot actually. tons of world drops and 7% chance to drop a bag)
-The phantoms have 2000 health

**Expected behavior**: 
This is the only video I could find of the curse on classic
https://youtu.be/5TsWsXZM32M?t=877
-There is no phantom spawn on curse application (or maybe just low chance)
-There are only TWO phantoms spawns over 5 minutes : at  170second and 200second. 
-The phantoms have NO loot
-phantoms have very low health, hard to say how much but the  health addon on the video calculates about 300

**Steps to reproduce**:

1. let a spectral citizen apply the curse on you

**Client version**: 
1.12

T1ti:
loot part is fixed in 7954585493fe5d9e45deeef27f10a664884bd458

PR #977, opened 2021-01-06
Add rejuvenation spell when eating deviate fish, improve spell effect "Party Time"

schell244:
Credit goes to TrinityCore https://github.com/TrinityCore/TrinityCore/pull/25802/files

## 🍰 Pullrequest
Added missing random spell effect when eating a deviate fish 
https://classic.wowhead.com/spell=8070/rejuvenation

### How2Test
- eat a raw deviate fish 

### Todo / Checklist
-  I think currenlty Aura::PeriodicDummyTick() has some issues/ is not working ( this affects spell "Party Time"). 
Related issues: https://github.com/vmangos/core/issues/866, https://github.com/vmangos/core/issues/865

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -6234,7 +6234,14 @@ void Aura::PeriodicDummyTick()
                 }
                 case 8067:                                  // Party Time!
                 {
-                    target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH, EMOTE_ONESHOT_DANCE));
+                    if (!target->HasAura(8067))
+                    {
+                        if (target->IsMoving())
+                            target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH));
+                        else
+                            target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH, EMOTE_ONESHOT_DANCESPECIAL));
+                        // TODO: TrintiyCore: _player->m_Events.AddEventAtOffset(this, RAND(5s, 10s, 15s));
+                    }
                     return;
                 }
                 case 7057:                                  // Haunting Spirits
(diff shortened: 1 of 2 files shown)


ulfgebhardt:
Related source on wowhead https://de.classic.wowhead.com/item=6522/deviatfisch#english-comments the comment with 75 upvotes lists the spells.
According to the comment there it should be `invigorate` not `rejuvenation`
https://classic.wowhead.com/spell=8065/invigorate

src/game/Spells/SpellAuras.cpp
```diff
@@ -6234,7 +6234,14 @@ void Aura::PeriodicDummyTick()
                 }
                 case 8067:                                  // Party Time!
                 {
-                    target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH, EMOTE_ONESHOT_DANCE));
+                    if (!target->HasAura(8067))
```
ratkosrb: This check will never be true, since you are in the handler of spell 8067.

src/game/Spells/SpellAuras.cpp
```diff
@@ -6234,7 +6234,14 @@ void Aura::PeriodicDummyTick()
...
+                    {
+                        if (target->IsMoving())
+                            target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH));
+                        else
+                            target->HandleEmoteCommand(PickRandomValue(EMOTE_ONESHOT_APPLAUD, EMOTE_ONESHOT_CHEER, EMOTE_ONESHOT_CHICKEN, EMOTE_ONESHOT_LAUGH, EMOTE_ONESHOT_DANCESPECIAL));
+                        // TODO: TrintiyCore: _player->m_Events.AddEventAtOffset(this, RAND(5s, 10s, 15s));
+                    }
```
ratkosrb: What are you trying to do here?

src/game/Spells/SpellAuras.cpp
schell244: I've removed this redundant check.

src/game/Spells/SpellAuras.cpp
schell244: Removed this as well. I think trinity added a random time before the "party emote event" gets triggered on player.

schell244:
Due to https://vanilla-wow.fandom.com/wiki/Deviate_Fish `rejuvenation` should be one of the random spells as well. `invigorate` is already contained in the spell list.

src/game/Spells/SpellAuras.cpp
schell244: Due to https://vanilla-wow.fandom.com/wiki/Deviate_Fish, the party emote lasts 2 minutes and gets triggered every 10–15 seconds. That's probaly what they what trinity did with `AddEventAtOffset`.

schell244:
Just tested changes of this PR. The new added random spell `rejuvenation `works fine, also the "Party Time" spell works - intervall between emotes seems to be always 10 seconds. 
The emote EMOTE_ONESHOT_DANCESPECIAL (triggered by "Party Time" spell when player is not moving) has no visual effect.
This can be tested with .npc playemote 401.
Will I just set it back to EMOTE_ONESHOT_DANCE? Does anyone know if there is some special dance emote in classic?

schell244:
I think EMOTE_ONESHOT_DANCESPECIAL is not available in vanilla client, so I've restored EMOTE_ONESHOT_DANCE.
PR is now ready from my side

PR #978, opened 2021-01-07
Fixes to Theradras

Gamemechanicwow:
## 🍰 Pullrequest
Makes Theradras immune to nature damage
And only cast Dustfield if there is a player in range

### Proof
Princess Theradras should be immune to all nature spells according to Classic WoW and a comment posted by Mumba on 2005-11-04. Here's a few examples from Classic WoW:

1. https://youtu.be/1wpA97AycA8. The hunter in the video is using a bow called Verdant Keeper's Aim (as seen at 0:46 in the video) which has a Chance to strike your ranged target with Keeper's Sting for 15 to 21 Nature damage. At 7:29, 8:44 and 9:53 in the video his bow procs and the boss is immune every time. Watch for the green wrath-like visual effect on his character's hands followed by Immune text on the target frame.
2. https://youtu.be/x4Bb-wjnvaw?t=34435. Watch the fight and notice the boss is immune to the Deadly Poison effect on the character's main hand weapon (as seen applied at 9:30:35).
3. https://youtu.be/30ZU7RD-wSQ?t=37559. The druid casts Faerie Fire (Feral) and the boss is immune.

Princess Theradras should not cast Dust Field if nobody is in melee range of her according to Classic WoW. Here's a few examples:

1. https://youtu.be/1wpA97AycA8?t=405. Here the hunter starts explaining how he got Princess Theradras to cast Dust Field using his pet.
2. https://youtu.be/ORL-CpNVFzc?t=140. Watch the fight and notice Princess Theradras only casts Dust Field if she reaches her target.

### Issues

- fixes https://github.com/the-hyjal-project/bugtracker/issues/176

### How2Test
- None

### DIFF
--- a/sql/migrations/20210107040717_world.sql
+++ b/sql/migrations/20210107040717_world.sql
@@ -1,0 +1,19 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20210107040717');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20210107040717');
+-- Add your query below.
+
+-- Princess Theradras is immune to nature damage
+UPDATE `creature_template` SET `school_immune_mask`='8' WHERE  `entry`=12201 AND `patch`=0;
+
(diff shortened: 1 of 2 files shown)


ratkosrb:
This will cast the spell multiple times, for each target in melee range.

ratkosrb:
Decided to use the opportunity to re-script this boss in db.

PR #1007, opened 2021-01-18
Fix Mooncloth pattern pre 1.3

ghost:
## 🍰 Pullrequest
This pull request does the following:

1. Changes Pattern: Mooncloth to the correct name pre 1.3
2. Changes the vendor that sells Pattern: Mooncloth pre 1.3

### Proof
The name of the pattern is Pattern: Purify Mooncloth pre 1.3
https://web.archive.org/web/20050218155128/http://wow.allakhazam.com/db/item.html?witem=14526

The vendor that sells the pattern is different pre 1.3
https://wowwiki.fandom.com/wiki/Patch_1.3.0#Professions
"Mooncloth recipe moved to a different vendor in Everlook."

You can see in this Allakhazam archive that the vendor Evie Whirlbrew sells it until 1.3 when it is changed to Qia
https://web.archive.org/web/20050216032853/http://wow.allakhazam.com/db/mob.html?wmob=11188


ulfgebhardt:
Your work looks really awesome @thenewblock - welcome on board. Please don't get frustrated on the slow merges. Normally takes a few weeks, months or years till stuff goes into the core ❤️

PR #1022, opened 2021-01-27
Use cmake pch instead of custom FindPCH

WobLight:
## 🍰 Pullrequest
Using cmake pch (target_precompile_headers) instead of FindPCH allows to build using precompiled headers with gcc.
This help compilation times.

***This increases minimum cmake version (3.16)***, more work is required if we want backward compatibility.


WobLight:
there's some work pending, don't merge yet.

WobLight:
changes done

PR #1032, opened 2021-01-28
Add legacy cmake support for PCH.

ratkosrb:
Followup to https://github.com/vmangos/core/commit/1b3274d27828a67d28a35e7629dcfe18a78b2a13


WobLight:
Is the legacy implementation needed at all? Afaik it only worked on windows, are there cases where updating cmake would be troublesome on it?

WobLight:
I think the empty `.cpp` files with the pch include have no meaning,  but I can't test this.

ratkosrb:
I use version 3.12 and don't want to increase requirements. Having both implementations allows it to work for everyone. I want the project to work on as many systems are possible.

PR #1033, opened 2021-01-28
CMake: fix default PCH

WobLight:
## 🍰 Pullrequest
CMake: fix default PCH

This would also be fixed by #1032


ratkosrb:
Resolved in https://github.com/vmangos/core/pull/1032

PR #1047, opened 2021-02-04
fixed some german broadcast locales

### LINKED ISSUE #1046: 🐛 [Bug] broadcast text locales - some (format?) issues for german texts
## 🐛 Bug report
I noticed some spelling issues within some german texts.
Looks like "D. $n" was used as replace.

broadcast text errors

### Expected behavior
Texts should not be cut.

### Steps to reproduce
1. open db
2. run query: 
``` sql 
SELECT entry,male_text_loc3 FROM locales_broadcast_text WHERE male_text_loc3 LIKE '%D. $n%'; 
```

Commit Hash:  
9dba4cc

schell244:
## 🍰 Pullrequest
Corrected some spelling/format issues within german broadcast locales.

### Proof
- None 

### Issues
- fixes https://github.com/vmangos/core/issues/1046

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20210204225833_world.sql
+++ b/sql/migrations/20210204225833_world.sql
@@ -1,0 +1,33 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20210204225833');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20210204225833');
+-- Add your query below.
+
+UPDATE `locales_broadcast_text` SET `male_text_loc3`='Ich brauche Hilfe. $n hat mich gesehen.' WHERE `entry`=488;  
+UPDATE `locales_broadcast_text` SET `male_text_loc3`='$n will mich töten! Hilfe!' WHERE `entry`=490;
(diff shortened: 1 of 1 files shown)


ratkosrb:
Correcting broadcast texts? But that data is all sniffed.

schell244:
I've noticed some of those spelling issues a while ago, yesterday I checked these texts on trinity core.
It seemd than they had the same issues, like one or two texts also contained this "D." but they have corrected most of them.
So when when those texts are broadcasted, it basically looks like this:

"Hilfe! D. Scheckiger Worg greift uns an!"
"Help! D. Mottled Worg is attacking us!" 

gamebreaking of course.... 

quest: https://classic.wowhead.com/quest=435/escorting-erland
text:   1074
$n:     https://classic.wowhead.com/npc=1766/mottled-worg

ratkosrb:
You need to provide some proof the sub string "D." gets removed in classic servers. Otherwise we should not touch sniffed data. The goal is to emulate what Blizzard does, and that does not include fixing their spelling mistakes.

schell244:
So I found a video where someone was doing the quest Escorting Erland where I saw the bug on vmangos.
Sadly a different aggro text was broadcasted in the video (I think there are 3 radom texts or so) which did not contain a "D." substring. 
But at 3:05, the video showes a text which contained a cut text part:
https://www.youtube.com/watch?v=HOSYHXg3pdc 
escorting erland 2

The " Ei. " part is exactly the same in vmangos broadcast text. 

So Blizzard definately seem to broadcast some cut/broken text elements - which makes this PR a custom patch.

PR #1078, opened 2021-02-20
[WIP] Documentation

ulfgebhardt:
## 🍰 Pullrequest

- Update `Readme.md`

image
image

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,25 @@
 
 
 # Progressive Vanilla
-This project is an independent continuation of the Elysium / LightsHope codebases, focused on delivering the most complete and accurate content progression system possible, including support for the patch appropriate game clients.
+*vMaNGOS* is an open-source Vanilla WoW emulation project. It is an independent continuation of the *Elysium* / *LightsHope* codebases, focused on delivering the most complete and accurate content progression system possible, including support for the patch appropriate game clients.
 
-### Currently supported builds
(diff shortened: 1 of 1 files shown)


ratkosrb:
You've just copied the "About the project" page from the wiki and pasted it in the readme...

The readme should be short, we do not need to go into such detail there. People who want to learn more can go read additional info in the wiki, this is where documentation belongs.

https://github.com/vmangos/wiki/wiki/About-the-project

ulfgebhardt:
More important than something to be short is it to be structured well and in a easy to find place. While the wiki is not easy to find, the README.md certainly is. The structuring on the other hand feels extraordinary to me ;)

And I do understand the argument that in case of DMCA the wiki is gone as well, which would be bad. This does not apply to docu which is directly included in the code, since there will always be several copies of it available (which is actual superior to the current wiki, which could also be subject the a DMCA if the whole orga is targeted)

Furthermore I believe the suggested change does not make the README.md too long; I simply merge some texts from the wiki into it to include all important infos. The big part about progression is  justified I guess since its the focus of this mangos-fork.

Everything else are rather minor changes which the majority attributes the origin of the code as well as the competitor projects.

GrenderG:
I like this change. However, I'd make a shorter **Progression** section and add a link to the bottom of it saying something like `To learn more about the project, visit https://github.com/vmangos/wiki/wiki/About-the-project` or similar, instead of copy pasting it entirely from the wiki.

ratkosrb:
I disagree with documentation being part of the core repo. Stop trying to throw everything into 1 pile. The readme contains a direct link to the wiki, how is that hard to find?

tonymmm1:
@ulfgebhardt I would like for there to be some documentation on the mangosd commands that are available from the server console/in-game. As the older mangos admin documentation is not up to date.

PR #1081, opened 2021-02-22
Old dragonkin elites should be elites in pre patch 1.12.0.

Daribon:
### Proof

- Joana old leveling videos, from patch Patch 1.9.2. For example https://youtu.be/DHPascB50tg?t=1987
firefox_CXWweUd6jO
- https://youtu.be/nTOL6lzTLMs?t=1234
firefox_0amg3ZPwHJ
- Allakhazam, for example https://web.archive.org/web/2018*/https://wow.allakhazam.com/db/mob.html?wmob=746

### Issues

- None

### How2Test

- Go on pre patch 1.12.0 and they should become elites.

### Todo / Checklist

- [ ] Update stats


ratkosrb:
Have you updated their stats too? Elite mobs should have higher stats.

Daribon:
Nah, that's left to do.
I have no clue on the supposed damage and HP values they are supposed to have as elites.

T1ti:
> Nah, that's left to do.
> I have no clue on the supposed damage and HP values they are supposed to have as elites.

most of the time elites are just 300% hp and dmg, especially if they aren't unique

PR #1083, opened 2021-02-22
[WIP] Added old levels for Duskwood bosses.

Daribon:
## 🍰 Pullrequest

Updates level values for Duskwood bosses below patch 1.6.
### Proof

https://wow.gamepedia.com/Patch_1.6.0#World_Environment
> Morbent Fel's level was slightly reduced to better fit the level range of Duskwood.
https://web.archive.org/web/20041231112935/http://wow.allakhazam.com/db/mob.html?wmob=1200

The other two bosses aren't found in any patch note, however when looking at allakhazam their level seems to be have been changed at the same time as Morben Fel.

> Stalvan Mistmantle 
https://web.archive.org/web/20050413234203/http://wow.allakhazam.com/db/mob.html?wmob=315
> Zzarc' Vul
https://web.archive.org/web/20050106062657/http://wow.allakhazam.com/db/mob.html?wmob=300

There is also this old leveling guide, which confirms both Zzarc’Vul and Stalvan Mistmantle level values :
https://www.paullynch.org/wowblog/?p=18

- None

### Issues

- None

### How2Test

- Use a patch below 1.6 and go to the creature locations to see if level changed.

### Todo / Checklist

- [ ] Update stats

### DIFF
--- a/sql/migrations/20210222173358_world.sql
+++ b/sql/migrations/20210222173358_world.sql
@@ -1,0 +1,29 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20210222173358');
+IF v=0 THEN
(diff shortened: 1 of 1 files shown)


GrenderG:
I think we should merge this already and just update old stats if we ever find them.

PR #1096, opened 2021-03-03
Fix Starshards and Curse of Agony

WobLight:
- tick damage ramp should not affect spell power increment
- fractional part is dithered locally as presently there's no dithering on damage

I'm planning to use that dithering function to fix rounding in spell damage/heal calculation everywhere so review carefully.


WobLight:
Anyone who can re-run CI?

WobLight:
mm I'll try merge with dev

ratkosrb:
The CI seems to be bugged atm, it fails on development branch too.

WobLight:
I have a big dithering pr ready, but this should get merged first

PR #1097, opened 2021-03-03
Misc DB Fixes

NickTyrer:
Two instances of the gameobject where spawned in the same locations causing them to be difficult to interact with


ratkosrb:
I see many spawns of gameobject id 19535 in sniffs, so why are you deleting them? This is not correct.

NickTyrer:
What about gameobject id 13891?

NickTyrer:
I have now updated the PR updating the gobject flags rather than deleting them.

sql/migrations/20210303085717_world.sql
```diff
@@ -44,6 +44,9 @@ UPDATE `gameobject_loot_template` SET `maxcount` = 3 WHERE `entry` = 1415 AND `i
 UPDATE `spell_affect` SET `SpellFamilyMask` = 6447547418 WHERE `entry` = 18218; 
 UPDATE `spell_affect` SET `SpellFamilyMask` = 6447547418 WHERE `entry` = 18219;
 
+-- Marsh Flesheater and Bloodscalp Berserker should not flee at low hp
+DELETE FROM `creature_ai_scripts` WHERE `id` IN (59702, 75102);
```
ratkosrb, line 50: You must delete the event too, not just the script.

NickTyrer: There are other scripts assigned to the events.

PR #1100, opened 2021-03-05
Implement dithering of fractional values

WobLight:
Currently, floating point values are truncated, this is not blizzlike and causes inconsistencies with blizzard tooltips (e.g. Thors has 15 damage, with +75% talent, tooltip will say 26 to 27 damage but currently it would always deal 26, due to truncation).

Changes:
- all damage (and healing), including auras `amount` are now float instead of integers
  - this also include the value stored in characters database for auras
  - this affects a lot of code paths and hacky stuff around, so might have subtle side effects
- removed some hacky pointer to integers, now using `optional` instead. A custom implementation of `optional` is provided or `c++17` can be used
  - to allow `c++17` compatibility `register` keywords are being purged from the code
- `eastern_plaguelands.cpp` converted to utf-8 encoding

What is presently tested:
- various spells dithers correctly with talents/spell power
  - Moonfire, both instant and dot
  - Healing Touch
  - Rejuvenation
  - Thorns
  - Starshards


GrenderG:
Can you please translate French comments to English when you see them. :pray:

src/game/Spells/SpellEffects.cpp
```diff
@@ -360,7 +360,7 @@ void Spell::EffectSchoolDMG(SpellEffectIndex effect_idx)
                     // percent from health with min
                     case 25599:                             // Thundercrash
                     {
-                        damage = unitTarget->GetHealth() / 2;
+                        damage = unitTarget->GetHealth() / 2.;
```
ratkosrb: 2.0f or it counts as a double

WobLight:
Another round of checks done. Help testing for regression would be much appreciated, there are lots of code paths...

ghost:
Ready to merge

PR #1106, opened 2021-03-08
Rework stealth calculation

WobLight:
Make stealth detection calculation depend on creature detection range and clean up calculation. This will advantage lower level players against creatures, max level detect range should stay about the same.


WobLight:
I'm also thinking of fixing Hunter's Mark, which ability to detect stealth is too strong (it shouldn't be unlimited range, but I don't know how much it's supposed to be).

PR #1127, opened 2021-03-18
Remove 'Teaches you how to...' description from all locales.

GrenderG:
## 🍰 Pullrequest
These items had `Teaches you how to...` text set as description in the different locales,  that's not Blizzlike in Vanilla.


Kittnz:
Can you give an example how it is in classic and in vanilla? Any evidence?

GrenderG:
> Can you give an example how it is in classic and in vanilla? Any evidence?

Well, the `locales_item` table was filled using data from newer expansions where the Mount and Pet tab already existed and you needed to "learn" how to summon them. This was obviously not a thing in Vanilla and Classic. For evidence, you can check in _classic.wowhead.com_ and switch between languages to see that there's no description.

I also remember those descriptions not being a thing in Vanilla when the Spanish language was released, but I don't have any screenshot of that. :sweat_smile:

PR #1128, opened 2021-03-19
Allow the usage of race specific languages even if crossfaction chat is enabled

GrenderG:
## 🍰 Pullrequest
With this change players will be able to use race specific languages even if crossfaction chat is enabled. Crossfaction chat will only have effect if they are using default faction languages (Common or Orcish).

### DIFF
--- a/src/game/Handlers/ChatHandler.cpp
+++ b/src/game/Handlers/ChatHandler.cpp
@@ -175,8 +175,7 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket& recv_data)
             lang = LANG_UNIVERSAL;
         else
         {
-            // Send message in universal language if crossfaction chat is enabled and player is using default faction
-            // languages.
+            // Send message in universal language if crossfaction chat is enabled and player is using default faction languages.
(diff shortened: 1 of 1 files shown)


src/game/Handlers/ChatHandler.cpp
```diff
@@ -175,8 +175,9 @@ void WorldSession::HandleMessagechatOpcode(WorldPacket& recv_data)
             lang = LANG_UNIVERSAL;
         else
         {
-            // send in universal language in two side iteration allowed mode
-            if (sWorld.getConfig(CONFIG_BOOL_ALLOW_TWO_SIDE_INTERACTION_CHAT))
+            // Send message in universal language if crossfaction chat is enabled and player is using default faction
+            // languages.
```
ratkosrb: why is the last word on a separate line

GrenderG: Because the line had more than 120 characters.
image

ratkosrb: so? whats special about that number

GrenderG: Coding conventions to avoid big lines. Originally 80 characters was the default limit, but today 120 is what most people use. It even comes marked by default (the vertical line you see in the screenshot) in CLion.

PR #1129, opened 2021-03-19
Vaelastraz shouldn't despawn after 1 hour in Patch 1.8 or greater.

GrenderG:
## 🍰 Pullrequest
Vaelastraz shouldn't despawn after 1 hour in Patch 1.8 or greater.

### Proof
https://wow.gamepedia.com/Patch_1.8.0

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_vaelastrasz.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_vaelastrasz.cpp
@@ -206,8 +206,12 @@ struct boss_vaelAI : public ScriptedAI
         if (!m_bEngaged)
         {
             m_bEngaged = true;
-            m_creature->SetRespawnDelay(43200); // 12h 43200
-            m_creature->ForcedDespawn(3600000); // 1h 3600000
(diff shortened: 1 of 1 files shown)


src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_vaelastrasz.cpp
```diff
@@ -206,8 +206,12 @@ struct boss_vaelAI : public ScriptedAI
         if (!m_bEngaged)
         {
             m_bEngaged = true;
-            m_creature->SetRespawnDelay(43200); // 12h 43200
-            m_creature->ForcedDespawn(3600000); // 1h 3600000
+            // From 1.8: There is no longer a one-hour time restriction on the Vaelastraz the Corrupt encounter in Blackwing Lair.
+            if (sWorld.GetWowPatch() < WOW_PATCH_110)
```
ratkosrb: Why are you using 1.10 here, when it says 1.8 above.

GrenderG: Because Grender smart. 🤗

PR #1132, opened 2021-03-23
Fix taunt to prevent exponential threat gains between two taunt casters.

Velite12:
Correct the addThreat calculation so that the caster of a taunt spell receives threat equal to the current tank's threat.

Previously, your threat was added with the current tank's threat which was causing exponential threat gain by using taunt between two warriors/druids.

This corrects taunt to work as specified in the comments. On taunt, you will now gain (current tank's threat - your current threat) aggro so that you become equal to the current tank's threat.

### DIFF
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -4066,7 +4066,8 @@ void Spell::EffectTaunt(SpellEffectIndex eff_idx)
     // Also use this effect to set the taunter's threat to the taunted creature's highest value
     if (unitTarget->CanHaveThreatList() && unitTarget->GetThreatManager().getCurrentVictim())
     {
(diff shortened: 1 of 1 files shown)


src/game/Spells/SpellEffects.cpp
```diff
@@ -4066,7 +4066,8 @@ void Spell::EffectTaunt(SpellEffectIndex eff_idx)
     // Also use this effect to set the taunter's threat to the taunted creature's highest value
     if (unitTarget->CanHaveThreatList() && unitTarget->GetThreatManager().getCurrentVictim())
     {
-        unitTarget->GetThreatManager().addThreat(m_casterUnit, unitTarget->GetThreatManager().getCurrentVictim()->getThreat());
+        //Add the difference of the Current highest value threat and the taunter's threat
```
GrenderG: Please add an extra space between `//` and `Add` and remove uppercase from `Current`. :pray:

Velite12: Made a second commit to update, hope that's alright.

PR #1160, opened 2021-04-11
Fix more quest completion issues.

GrenderG:
## 🍰 Pullrequest
This is an attempt to fix quest issues:
- [x] Sometimes you can't complete a quest even if you have all items.
- [ ] Item count is not updated when you deposit the item on bank.

### DIFF
--- a/src/game/Objects/Player.h
+++ b/src/game/Objects/Player.h
@@ -1053,9 +1053,7 @@ class Player final: public Unit
         void SendProficiency(ItemClass itemClass, uint32 itemSubclassMask) const;
         Item* BankItem(ItemPosCountVec const& dest, Item* pItem, bool update)
         {
-            Item* storedItem = StoreItem(dest, pItem, update);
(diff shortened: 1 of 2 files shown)


src/game/Objects/Player.h
```diff
@@ -1053,7 +1053,9 @@ class Player final: public Unit
         void SendProficiency(ItemClass itemClass, uint32 itemSubclassMask) const;
         Item* BankItem(ItemPosCountVec const& dest, Item* pItem, bool update)
         {
-            return StoreItem(dest, pItem, update);
+            Item* storedItem = StoreItem(dest, pItem, update);
+            ItemRemovedQuestCheck(pItem->GetEntry(), pItem->GetCount());
```
GrenderG: This is **not** making the item count update on item banking. Investigate.

GrenderG:
I removed my attempt to make objective count change when banking items since:
1. I couldn't make it work.
2. @Gamemechanicwow said he's not sure if that's a Classic-only thing or was happening in Vanilla too, so needs a bit of research.

This PR is ready for merge.

PR #1171, opened 2021-04-18
Draft: fix condition of defence skill increase 

schell244:
## 🍰 Pullrequest
- fix condition of defence skill increase (sometimes defence got updated by two points)
- use the same increase chance formula for weapon and defence

### Proof
- None

### Issues
- None

### How2Test
- create a new player
- get attacked by an enemy npc 
- defence skill increase should only happen if player dodges/parrys/blocks/resists an enemy attack

### Todo / Checklist
- [ ] check if this is classic behavior


schell244:
closed since not classic behavior

PR #1178, opened 2021-05-01
added new sound and timer for lighthouse bell

schell244:
## 🍰 Pullrequest
use sound https://classic.wowhead.com/sound=7197/lighthousefoghorn instead of https://classic.wowhead.com/sound=6594/belltollalliance for westfall lighthouse hourly bell gameobject.

### Proof
I just found that sound id and remembered a bell gameobject spawn at westfall lighthouse.
I don't know if this is classic behaviour, but sound 6594 feels wrong at this location.

### Issues
- None

### How2Test
- move player to westfall lighthouse
- .event start 78

### Todo / Checklist
~~[x] classic behaviour?~~


GrenderG:
Needs proof, can you check in the Classic PTR? No active subscription needed.

schell244:
Sadly Classic PTR seems to be no longer available. Would be nice if someone who has an active subscription could help out with this. I could not find any other usage of sound 7197 within the core.

GrenderG:
Just checked in Classic and I can confirm this is the sound being played. Any other lighthouses where this sound might be missing?

schell244:
Thanks @GrenderG ! I'll iterate all bell spawn points and check if there's another lighthouse. Not sure if the current spawn points in db cover all classic spawns.

schell244:
Found another lighthouse with a bell spawn at alcaz island (.go object 9104).

src/scripts/world/go_scripts.cpp
```diff
@@ -419,6 +422,15 @@ struct go_bells : public GameObjectAI
                     case ASHENVALE_ZONE:
                         _soundId = BELLTOLLNIGHTELF;   // nightelf bell sound 
                         break;
+                    case WESTFALL_ZONE:
+                        _soundId = LIGHTHOUSEFOFHORN;   // lighthouse fog horn
+                        break;
+                    case DUSTWALLOW_MARSH_ZONE:
+                        if(me->GetAreaId() == 2079) // Alcaz Island - lighhouse fog horn
```
GrenderG: Please add an space after the `if` :pray:

GrenderG:
There's also a bell missing in the lighthouse next to Theramore Isle, I assume this one has that sound too, but it's in the same area (513) as the normal Theramore bell... so maybe the way sound ids are being assigned needs a bit of refactoring:
```
         guid: 87737
           id: 176573
          map: 1
      zone_id: 15
      area_id: 513
   position_x: -3667.09
   position_y: -4754
   position_z: 1.80047
  orientation: 2.26892
    rotation0: 0
    rotation1: 0
    rotation2: 0.906307
    rotation3: 0.422619
 is_temporary: 0
 creator_guid: 0
   creator_id: 0
 creator_type: 
   display_id: 3972
        level: 0
      faction: 114
        flags: 0
dynamic_flags: 0
path_progress: 65535
        state: 1
         type: 5
       artkit: 0
 custom_param: 0
     sniff_id: 186
  sniff_build: 36325
```

schell244:
I've actually confused Theramore and Menethil in the comments. Fixed now. Maybe whole Theramore should be covered with the lighhouse sound? Since there's no other 'bell tower'. In this case the if statement could be removed and lighhouse sound could be used for all alliance bell at zone 15 (Dustwallow Marsh)?

GrenderG:
> 
> 
> I've actually confused Theramore and Menethil in the comments. Fixed now. Maybe whole Theramore should be covered with the lighhouse sound? Since there's no other 'bell tower'. In this case the if statement could be removed and lighhouse sound could be used for all alliance bell at zone 15 (Dustwallow Marsh)?

There are 2 Lighthouses (one near Theramore and another one in Alcaz Island) and a Bell tower inside Theramore. I have also noticed that in Classic, the lighthouse sound is played every 2 minutes (:53, :55, :57, :59, :01...).

schell244:
Ok, in this case there's indeed a different logic required in order to distinguish between lighhouse/bell. Also those lighhouse objects need their own time schedule.

schell244:
I've added the missing spawn for Theramore lighthouse bell/horn you've found and also added some logic which checks for the object guid to decide which sound should be used. With the current changes the sound get's played once every hour - so the correct time schedule is still missing. Any suggestions?

schell244:
Time schedule for lighthouse sound is fixed now as well.

PR #1180, opened 2021-05-02
Added HK ranks to Battle bots

alexkulya:
Added HK ranks to Battle bots.
Testing, no errors.
Maybee need add some gear conditions...

Screenshot_1


ratkosrb:
Will that not fuck up the honor calculations for real players later? I think it would be better to just set the rank in the updatefield directly, and not touch honor mgr.

alexkulya:
I'll test it more thoroughly and see what can be done.

PR #1238, opened 2021-06-20
Add Firecrackers Script

NickTyrer:
## 🍰 Pullrequest
This PR will set off random firecrackers during the lunar festival event

### How2Test
.event start 7

### Todo / Checklist
FIx errors


NickTyrer:
I think this functionality should be handled by the core as the radius used in generic scripts is not large enough to cover an entire city.

ratkosrb:
Just assign the script to a NPC near the gameobjects?

NickTyrer:
The gameobjects are everywhere, you would have to assign it to many npc's throughout the world.

Kittnz:
This isn't how it works. The firecrackers at random do the despawn animation when they despawn. And when they respawn it will spawn one of the 2 different firecracker gobs.

NickTyrer:
I dont think that functionality is possible, unless you know a way?

Oroxzy:
similar to this?
https://github.com/vmangos/core/blob/development/src/scripts/world/fireworks_show.cpp

Kittnz:
No, different @Oroxzy .
@NickTyrer  You can do it with pools, i think it's the only way to do it. @brotalnia WYT?

ratkosrb:
I guess you'll need a c++ script assigned to those gameobject entries to despawn them. And for each spawn point you need a pool of all the different entries that can spawn there.

Kittnz:
Wouldn't it be easier to just add a event script to it? make it despawn every x seconds ?

NickTyrer:
I can add the pooling but I wont be able to do the c++

NickTyrer:
I have now added the relevant pools.

PR #1241, opened 2021-06-20
fix stat modifier issue

### LINKED ISSUE #1236: 🐛 [Bug] damage values at character UI sometimes get displayed with wrong color
## 🐛 Bug report
Sometimes meele and range damage values are displayed in red while the damage modifier is at 100%.

This issue does soemtimes occur after an aura gets removed which had positiv impact to the damage modifier. 

Unbenannt

### Expected behavior
When damage modifier is at 100%, the damage should get displayed with white color.

### Steps to reproduce
1. create a warrior
2. put 1 skill point on enrage
3. get hit by an enemy and wait for enrage to get triggered
4. wait for enrage to fade -> damage is at 100% an displayed in red

Notes:
1. Warrior: changing between battle and defence stance won't fix the issue, damage will still get displayed red at 100% at battle stance.
2. Relogging does fix it.
3. Warrior: there seem to be a difference how many skill points you put on enrage. 1 Skill point seems to always trigger it, while all skill points do not.
4. Battleground damage buffs do always trigger it

### Version & Environment

Client Version:
  - 1.12.1.5875

Commit Hash:  
79efe80ae39d94a5e52b71179583509b1df75899

OS Client: Win 10
OS Server: Win 10

schell244:
## 🍰 Pullrequest
Fix stat modifier calculation and the damage modifier shown in client.

Inspired by azerothcore and trinitycore:
<del>
https://github.com/azerothcore/azerothcore-wotlk/blob/36a6d04156214b207d7b1f286ff8f365c733a557/src/server/game/Entities/Unit/Unit.cpp#L14462
</del> (does work, but not related to the issue)

https://github.com/azerothcore/azerothcore-wotlk/blob/ea5b1b384658fc071ae509f6e26bb680bdb3c8ef/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L4860

https://github.com/TrinityCore/TrinityCore/blob/c08f086a02729f3b7e4c99f4741063e45301ebf8/src/server/game/Spells/Auras/SpellAuraEffects.cpp#L4187

### Issues
- fix https://github.com/vmangos/core/issues/1236

### Todo / Checklist
- [x] None

### DIFF
--- a/src/game/Objects/Object.h
+++ b/src/game/Objects/Object.h
@@ -368,8 +368,9 @@ class Object
 
         void ApplyPercentModFloatValue(uint16 index, float val, bool apply)
         {
-            val = val != -100.0f ? val : -99.9f ;
-            SetFloatValue(index, GetFloatValue(index) * (apply?(100.0f+val)/100.0f : 100.0f / (100.0f+val)));
+            float value = GetFloatValue(index);
+            ApplyPercentModFloatVar(value, val, apply);
+            SetFloatValue(index, value);
         }
 
         void SetFlag(uint16 index, uint32 newFlag);
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -5293,7 +5293,7 @@ void Aura::HandleModDamagePercentDone(bool apply, bool Real)
             // For show in client
             if (target->GetTypeId() == TYPEID_PLAYER)
             {
-                // TODO: fix positive aura gets removed -> damage is displayed at 100%
+                // TODO: fix positive aura gets removed -> damage is displayed red at 100%
                 target->ApplyPercentModFloatValue(PLAYER_FIELD_MOD_DAMAGE_DONE_PCT, m_modifier.m_amount, apply);
             }         
         }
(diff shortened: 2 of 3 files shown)


ratkosrb:
Have you tested this to make sure it does not break anything else with stat calculations?

schell244:
I've added some logs and compared some modified values before and after the change. 
The displayed values (modifier and damage) have been the same for the following cases:
warrior -> switching stance
warrior -> adding/removing enrage
I also could not reproduce the display issue when using `ApplyPercentModFloatVar`.
I think it's becasue `ApplyPercentModFloatVar` will prevent the modifier to become negative, when value -100 is used.
I will do a few more tests with this change next week to make sure it's safe.

schell244:
I just did some more tests. Sadly the formula changes won't fix the display issue (but using `ApplyPercentModFloatVar` seems to be a valid replacment for the current hack within `HandleStatModifier`.)
While testing I've found a way how to always reproduce the issue #1236 and the culprit seems to be somewhere within the aura handling/removal code.

schell244:
It's finally fixed now. I did some tests and the issue will no longer occur, also the modifer values are correct.

schell244:
I just found an issue with wands. With this change, it seems that the damage modifier gets applied and removed twice for wands (tested with https://classic.wowhead.com/spell=23505/berserking). Meele and "normal" range values get displayed correct. I think it's due to to wands damage being handled twice as range and spell. More reseach is needed to solve this.
Closing this PR for now.

PR #1264, opened 2021-07-03
Love is in the Air Event

NickTyrer:
TODO:
1. When using the item Unbestowed Friendship Bracelet (22259), both players should get the buff The Power of Friendship (26664) Buff. Currently only the casting player gets the buff when both should.
2.  Popularity Contest is not implemented (some gameobjects and gossips are commented in the PR)


Oroxzy:
gameobjects respawn, creatures respawn and creature template fixes:
https://gist.github.com/Oroxzy/bf6f20dc582f05a3fc33eedf57d39231

alot of stuff is in here too:

`SELECT * FROM game_event_creature_data WHERE  event=8`

maybe take a closer look what it does :D

while i was doing midsummer event spawns i saw this in the replaycore, overlapping gameobjects from 2 events .. so the little heart thing is from the bonus week if the faction leader wins. Maybe we need todo one event for every leader to spawn and despawn them.

> The faction leader who wins the popularity contest will be rewarded with having Kwee remain in residence for an extra week beyond the end of the Love is in the Air event. If you /kiss Kwee, he will cast a +200 HP buff on you that lasts 1 hour. This is why he's valuable to keep a week beyond the event, so vote early and often!

some gameobjects are lasting until the 
grafik

PR #1279, opened 2021-07-17
Add Some Missing Creature Auras

NickTyrer:
## 🍰 Pullrequest
This PR adds some missing auras to creatures based on cmangos and sniff data


ratkosrb:
What's the status on this?

NickTyrer:
I diddent realise how many auras where missing or applied through event_ai in vmangos, this just scratches the surface. It will be a while before this one is done.

PR #1289, opened 2021-07-27
Lunar Festival Event

Oroxzy:
**TODO:**

- [x] Cluster/Firework Launcher pools.
- [x] Fix how Fireworks should work.
- [x] Add additional Firecracker pools.
- [x] Respawn all GameObjects (used Massparse 20/21).
- [x] Respawn all Creatures.
- [x] Lunar Teleportation scripting.
- [x] Add Spelllist, BroadcastTexts and GuardAI to Lunar Festival Sentinel.
- [x] Update all creature_templates.

**Firework and Cluster Launchers seems to be pooled:**
grafik

**Teleporting from Ironforge to Moonglade:**
grafik

**The Effect #1 Gameobject target is not in the sniffs (unsniffable like many of them.. also in scourge event and midsummer) but someone already added it to the DB. Actually its a check for the spell to make sure you are in the Spotlight**
grafik

**Fireworks**
As you can see here, Fireworks nearly instant fires up after using the item: https://www.youtube.com/watch?v=JDxo6W6a9wM&t=0s
grafik


NickTyrer:
Reminder for myself to update lunar festival emissary ai scripts when creature guids are set.

Oroxzy:
Why?

Oroxzy:
2

> Reminder for myself to update lunar festival emissary ai scripts when creature guids are set.

Why?

ratkosrb:
I think he is talking about the quest givers. Only a specific npc in each city offers the quest, while the others do not, even though they are of the same entry, so a condition to check the guids is used for enabling quest giver flag on the npcs.

ratkosrb:
```
Make the Teleport mechanic Blizzlike.
```
We have sniffs of all the teleport to moonglade scripts. Nicky helped me sniff that, and I think he already worked on it, or am I wrong?

```
Figure out how the Firecrackers really works.
```
Don't they work correctly now with the c++ script I added? Seemed fine to me. They play the animation and despawn every minute, and a different firecracker spawns after a few seconds.

sql/migrations/20210727201033_world.sql
```diff
@@ -75,7 +75,7 @@ INSERT INTO `spell_target_position` (`id`, `target_map`, `target_position_x`, `t
 (26456, 0, 1642.41, 239.9, 62.59, 3.01, 5086, 5875);
 
 -- Lunar Festival Sentinel texts are chat_type 12.
-UPDATE `broadcast_text` SET `chat_type`='12' WHERE `entry` IN (10948,10949,10950,10951,10952,10953,10954,10955);
+UPDATE `broadcast_text` SET `chat_type`=12 WHERE `entry` IN (10948,10949,10950,10951,10952,10953,10954,10955);
```
ratkosrb: Chat type 12??? The maximum is 6.
https://github.com/vmangos/core/blob/808ec9801870ae926fa037f100bca4d2c3efa2d8/src/game/Objects/CreatureDefines.h#L383

Issue #1302, opened 2021-08-08
🐛 [Bug] Giant Softshell Clam (177784)


NickTyrer:
## 🐛 Bug report
- Currently they have a respawn time of 120 hours
- You should only be able to interact with toe object if you are on the quest https://classic.wowhead.com/quest=6142/clam-bait
- According to the comments on wowhead Enraged Reef Crawlers eat the clams if they are near to them see https://youtu.be/sahhsmKoecA?t=184

Commit Hash:  60c8057

PR #1308, opened 2021-08-10
Misc DB Fixes

NickTyrer:
## 🍰 Pullrequest


NickTyrer:
Closes #1302

sql/migrations/20210810191400_world.sql
```diff
@@ -0,0 +1,647 @@
...
+INSERT INTO `creature_ai_scripts` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) VALUES (1094303, 0, 15, 26047, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Decrepit Guardian - Cast Birth');
+INSERT INTO `creature_ai_scripts` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) VALUES (1094303, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 6282, 0, 0, 0, 0, 0, 0, 0, 0, 'Decrepit Guardian - Talk');
+
+-- Redo Devilsaur Movement (credit cmangos)
+DELETE FROM `creature` WHERE `guid` IN (23743, 23744, 23745, 23741, 95009, 24172);
```
Kittnz, line 49: I think we have this sniffed tho

sql/migrations/20210810191400_world.sql
```diff
@@ -617,6 +617,27 @@ UPDATE `creature_template` SET `gossip_menu_id` = 5081, `npc_flags` = 3 WHERE `e
...
 (5081, 6288);
 
+-- Seersa Copperpinch, Guchie Jinglepocket and Jaycrue Copperpinch
+UPDATE `creature_template` SET `gossip_menu_id` = 5181 WHERE `entry` IN (13432, 13436, 13430);
+UPDATE `creature_template` SET `npc_flags` = 7 WHERE `entry` = 13432;
+UPDATE `creature_template` SET `npc_flags` = 5 WHERE `entry` = 13436;
+UPDATE `creature_template` SET `npc_flags` = 5 WHERE `entry` = 13430;
+
+-- Scarshield Infiltrator
+UPDATE `creature_template` SET `gossip_menu_id` = 2630, `npc_flags` = 1 WHERE `entry`=10299;
```
ratkosrb: This npc changes entry before you can talk to him.

ratkosrb: And npc_flags are 0 for it in sniff, so why are you changing them to 1?

NickTyrer: @ratkosrb doesnt the npc have to have that flag set if it has a gossip menu?

ratkosrb: This npc does NOT have gossip. If the reason you are adding gossip to it is because you see gossip menus for its entry inside the `creature_gossip` table in the sniff db, you are missing out on a key fact about how that works. The entry in this table, is the entry that is part of the creature's guid, for which a gossip menu was sent. The entry from the guid is the original entry with which it spawns. This value does not change. This npc however can never be interacted with in its normal form. As you can see in the sniffs it has npc_flags = 0 which means you can't talk to it. When a player goes near the npc and has the Unadorned Seal of Ascension item in his inventory, the npc changes its entry to Vaelan (10296), which you can see in the EventAI script for the npc. And actually the menu already exists and is fully implemented, but its using a placeholder id (above 10k). All you need to do is update the existing placeholder gossip menu ids for Vaelan to the ones in the sniff. You can see that they start from 12039. Look in the gossip menu table in vmangos and match them with the real ids from sniffs.

Example of what you have to do:
```
UPDATE `creature_template` SET `gossip_menu_id`=2630 WHERE `gossip_menu_id`=12039;
UPDATE `gossip_menu` SET `entry`=2630 WHERE `entry`=12039;
UPDATE `gossip_menu_option` SET `menu_id`=2630 WHERE `menu_id`=12039;
UPDATE `gossip_menu_option` SET `action_menu_id`=2630 WHERE `action_menu_id`=12039;
```

sql/migrations/20210810191400_world.sql
```diff
@@ -590,6 +590,9 @@ UPDATE `creature_loot_template` SET `ChanceOrQuestChance` = -50 WHERE `item` = 3
 -- Correct Diplay ID For Energetic Rod
 UPDATE `item_template` SET `display_id` = 30915 WHERE `entry` = 18321;
 
+-- Remove Teebus Blazing Longsword From Lockboxes
+DELETE FROM `reference_loot_template` WHERE `entry` = 1014 AND `item`=1728;
```
ratkosrb: It could drop from Heavy Junkbox according to thottbot snapshot from 2006.
https://web.archive.org/web/20090313074825/http://thottbot.com:80/i16885#D::A

ratkosrb: Oh no, wait. That's 2009, somehow I read the last digit upside down...

NickTyrer: I think it was added sometime in tbc to lockboxes

ratkosrb: Ok

sql/migrations/20210810191400_world.sql
```diff
@@ -0,0 +1,780 @@
...
+UPDATE `creature_loot_template` SET `ChanceOrQuestChance` = -30 WHERE `item` = 3236;
+
+-- Fix drop rate of items used in quest 451 (A Recipe For Death) in Silverpine Forest (credit cmangos)
+UPDATE `creature_loot_template` SET `ChanceOrQuestChance` = -50 WHERE `item` = 3257;
+
+-- Correct Diplay ID For Energetic Rod
+UPDATE `item_template` SET `display_id` = 30915 WHERE `entry` = 18321;
+
+-- Remove Teebus Blazing Longsword From Lockboxes
+DELETE FROM `reference_loot_template` WHERE `entry` = 1014 AND `item`=1728;
```
GrenderG: Should other weapons be removed too? Like Alcor's Sunrazor, Shadowblade and Blanchard's Stout. According to wowhead they aren't dropping either.

sql/migrations/20210810191400_world.sql
```diff
@@ -588,8 +588,8 @@ UPDATE `creature_loot_template` SET `ChanceOrQuestChance` = -50 WHERE `item` = 3
 -- Correct Diplay ID For Energetic Rod
 UPDATE `item_template` SET `display_id` = 30915 WHERE `entry` = 18321;
 
--- Remove Teebus Blazing Longsword From Lockboxes
-DELETE FROM `reference_loot_template` WHERE `entry` = 1014 AND `item`=1728;
+-- Remove Teebus Blazing Longsword, Alcors Sunrazor, Blanchards Stout and Shadowblade From Lockboxes
+DELETE FROM `reference_loot_template` WHERE `entry` = 1014 AND `item` IN (1728, 14555, 13046, 2163);
```
ratkosrb: Should you not remove the reference loot template at this point?

sql/migrations/20210810191400_world.sql
```diff
@@ -891,6 +891,7 @@ UPDATE `creature` SET `position_x` = -3893.58642578125, `position_y` = -1627.259
 -- -----------------------------------------------------------------------------------------
 
 -- Add Serpent Statue Script (NOTE: spell 19473 is not being cast by gameobject for some unknown reason also cannot use gameobject guid 12609 when it respawns)
+-- Does vmangos support Gameobject Casting?
```
ratkosrb: Yes, it does.

sql/migrations/20210810191400_world.sql
```diff
@@ -2952,6 +2952,19 @@ UPDATE `quest_template` SET `PrevQuestId` = 0 WHERE  `entry` = 691;
...
+-- -----------------------------------------------------------------------------------------
+
+-- Add Missing Text to Gilnid
+DELETE FROM `creature_ai_scripts` WHERE `id`=176301;
+INSERT INTO `creature_ai_scripts` (`id`, `delay`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) VALUES (176301, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1147, 1146, 1145, 0, 0, 0, 0, 0, 0, 'Gilnid - Say Text');
+
+-- -----------------------------------------------------------------------------------------
+
+-- Arikara should be Tameable and Cast Spell on Spawn (Credit Cmangos)
+UPDATE `creature_template` SET `unit_class` = 1, `ai_name` = 'EventAI' WHERE `entry` = 10882;
```
ratkosrb: The unit class for this npc is already 1?

Issue #1311, opened 2021-08-15
🐛 [Bug] Stormwind Guard (GUID: 79875) NPC


LEO33SoFo:
## 🐛 Bug report

The _Stormwind Guard_ located at _Elwynn Forest_ (35.29, 53.66) does not patrol.
### Expected behavior

I do not know if that _Stormwind Guard_ patrols in _Classic_, but logically he should patrol the road from _Goldshire_ to the entrance of _Stormwind City_.
### Steps to reproduce

1. `.go creature 79875`
2. Check his behavior.

### Version & Environment

Client Version: 
  - 1.12.1.6005

Commit Hash:  
7b83d9952f32559b72f7094e389631f4a30ef7b0

OS Client and Server:
  - Windows XP Professional

### Crashlog

- None

LEO33SoFo:
The _Twinstar Kronos_ private server had this [issue](https://vanilla-twinhead.twinstar.cz/?issue=3322) fixed.
I somewhat fixed this by adding the relevant waypoints based on the vanilla _TwinHead_ [database](https://vanilla-twinhead.twinstar.cz/?npc=1423).

```sql
UPDATE `creature` SET `position_x`='-9452.957031', `position_y`='64.345367', `position_z`='56.010929', `orientation`='6.246572', `movement_type`='2' WHERE (`guid`='79875') LIMIT 1;

INSERT INTO `mangos`.`creature_movement` (`id`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `waittime`, `wander_distance`, `script_id`) VALUES ('79875', '1', '-9433.38', '74.0578', '56.0882', '100', '0', '0', '0');
INSERT INTO `mangos`.`creature_movement` (`id`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `waittime`, `wander_distance`, `script_id`) VALUES ('79875', '2', '-9412.31', '90.9749', '57.4718', '100', '0', '0', '0');
INSERT INTO `mangos`.`creature_movement` (`id`, `point`, `position_x`, `position_y`, `position_z`, `orientation`, `waittime`, `wander_distance`, `script_id`) VALUES ('79875', '3', '-9368.46', '108.686', '60.8707', '100', '0', '0', '0');
[...]

PR #1315, opened 2021-08-26
Add Some Missing Creature Spawns

NickTyrer:
### Todo / Checklist
- [ ] Add Spawns From Cmangos
- [ ] Check Existing Spawns that are pooled and have movement_type 2 are using creature_movement not creature_movement_template


Kittnz:
Do not forget, cmangos also has a lot of manual spawned mobs. I suggest you work from sniff data instead.

NickTyrer:
The sniff data doesnt have the correct spawn points if the creature is roaming, unless the sniffer saw the mob spawn. If we know the creature spawns in multiple locations I think we should use cmangos data.

ratkosrb:
The cmangos db does not have the exact spawn coordinates of wandering mobs either, there is no way to get those unless you see the mob spawn as you said. You can just look at the positions the mob has been seen at, and pick a spot in the middle. Their guess is no more correct than ours. And it does not really matter when the mob is wandering, people will never see it at the exact position it spawns, it makes no difference if the spawn point is half a yard off, since the mob does not sit still anyways.

NickTyrer:
So what is the problem using cmangos data if we know the mob spawns near an area?

ratkosrb:
I listed you the problems already when we talked in discord.
- There were many manual spawns added in mangos back in the day, and even people adding completely made up waypoints to mobs, such as [this](https://github.com/vmangos/core/issues/1311) case. So you cannot just query missing or different spawns in their db and trust that to be authentic. Same way you cannot guarantee existing data we have is correct without checking sniffs.
- The vmangos and cmangos database have a common root, and they were down ported from TBC, which was downported from WotLK.
- The main developers of cmangos are focused on TBC, and Vanilla is merely an afterthought. The vast majority of commits in the classic db are simply ported from the TBC and WotLK repositories.
- None of the cmangos developers had any original Vanilla sniffs, so any work done prior to 2019 when Classic released, is based on TBC and WotLK sniffs.
- There were multiple changes made to the Vanilla world in TBC and WotLK. Elite areas were nerfed. New quests and npcs were added. Specific npcs were moved to a different place. Some mobs were removed and replaced with different ones. You cannot trust data from TBC to be correct for Classic. There are multiple places with different spawns i can mention off the top of my head that I've spent time to revert to their correct Vanilla state (quest givers in ashenvale, battlemasters in most cities, the majority of dustwallow marsh, many spawns in blasted lands...). If you were to start blindly replacing our spawns with cmangos classic-db data, you will revert most of these to their TBC state!

If you want to port spawns from cmangos, you must contact the people who added them and inquire what their source was on a case by case basis. If the work was based on 1.13 data, then you can port it. If it was based on TBC and WotLK sniffs they had, then you have to actually verify everything to make sure it's correct for Vanilla. Either go to that spot in the official Classic-Era (**NOT Classic-TBC**) servers, or look in the sniff db if we have any spawns near those coordinates.

NickTyrer:
Ok, I give up.

PR #1324, opened 2021-09-09
Some fixes for EquipRandomGearInEmptySlots.

GrenderG:
## 🍰 Pullrequest
- Allow healer bots to use any type of gear (this way paladin healers can use cloth, leather and mail, for example).
- Prevent non casters from equipping holdables instead of offhand weapons.

### DIFF
--- a/src/game/PlayerBots/CombatBotBaseAI.cpp
+++ b/src/game/PlayerBots/CombatBotBaseAI.cpp
@@ -2584,22 +2584,22 @@ void CombatBotBaseAI::EquipRandomGearInEmptySlots()
         if (!pProto)
             continue;
 
-        // only items that have already been obtained by someone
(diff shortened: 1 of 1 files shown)


src/game/PlayerBots/CombatBotBaseAI.cpp
```diff
@@ -2644,7 +2644,7 @@ void CombatBotBaseAI::EquipRandomGearInEmptySlots()
                         continue;
 
                     // Only equip holdables on mana users
-                    if (pProto->InventoryType == INVTYPE_HOLDABLE && me->IsCaster())
+                    if (pProto->InventoryType == INVTYPE_HOLDABLE && !me->IsCaster())
```
ratkosrb, line 2647: should check the role for healer or ranged dps instead

PR #1328, opened 2021-09-11
Store personal XP rates into DB

ghost:


Vanheden:
Love it <3

ratkosrb:
I've decided against this. I don't want to have fields for custom features in the db, and taking up space for each character.

PR #1329, opened 2021-09-11
New check for fishing channeling

ghost:
Sometimes it's not possible to start fishing because of "invisible"  walls and roofs due to wrong extracted maps.

### How2Test

- Try to fish in the canals of stormwind before and after this PR.

All credits to @Gamemechanicwow .


GrenderG:
I miss @Gamemechanicwow. :(

PR #1331, opened 2021-09-11
REPLACE instead of INSERT for some querys

ghost:


ratkosrb:
@Nolin7777 Can you resolve the conflicts? It's not letting me merge this.

Issue #223, opened 2019-06-25
Issues with Pyrewood Ambush quest


GrenderG:
**Current behavior**: 
- Humans summoned by this quest are humans even if done at night.
- Quest is not correctly marked as failed in the quest log if Deathstalker Faerleia dies.
- When a player fails the quest and Deathstalker Faerleia respawns, she doesn't have the quest giver flag. This doesn't always happen, but definitely does some times.

**Expected behavior**: 
- Humans summoned by this quest should be worgens if done at night.
- Quest should be correctly marked as failed in the quest log when the quest giver dies.
- The quest giver must always have the quest giver flag set again.

ulfgebhardt:
@GrenderG can you provide more details, like the link to the quest in wowhead. This is not clear to me.

GrenderG:
Wowhead: https://classic.wowhead.com/quest=452/pyrewood-ambush
YouTube video where you can see NPCs being worgens: https://www.youtube.com/watch?v=4Mj5tH2goIc

PR #1338, opened 2021-09-17
Quest: Pyrewood Ambush - fix some issues

schell244:
## 🍰 Pullrequest
- fix quest giver flag sometimes not being restored
- properly fail quest for player/group when Deathstalker Faerleia gets killed.

### Issues
- relates https://github.com/vmangos/core/issues/223
- this does not cover _Humans summoned by this quest should be worgens if done at night._

### How2Test
- accept quest https://classic.wowhead.com/quest=452/pyrewood-ambush
- fail quest (Deathstalker Faerleia gets killed)
 1. check if quest is properly failed (questlog failed /no turn in possible)
 2. check if Deathstalker Faerleia will have a quest giver flag again after respawn

### DIFF
--- a/src/scripts/eastern_kingdoms/silverpine_forest/silverpine_forest.cpp
+++ b/src/scripts/eastern_kingdoms/silverpine_forest/silverpine_forest.cpp
@@ -27,6 +27,7 @@ npc_deathstalker_faerleia
 EndContentData */
 
 #include "scriptPCH.h"
+#include "Group.h"
 
 /*#####
 ## npc_deathstalker_erland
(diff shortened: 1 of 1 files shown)


src/scripts/eastern_kingdoms/silverpine_forest/silverpine_forest.cpp
```diff
@@ -249,16 +253,34 @@ struct npc_deathstalker_faerleiaAI : ScriptedAI
         m_bEventStarted = false;
     }
 
-    void JustRespawned() override
+    void JustRespawned() override {}
```
ratkosrb: Why override it at all if you're not doing anything on respawn? This line should be removed.

PR #1341, opened 2021-09-21
Arcane Dust shouldn't be available on vendors.

GrenderG:
## 🍰 Pullrequest
For some reason this item was available at some vendors.

### Proof
https://classic.wowhead.com/item=17019/arcane-dust

### DIFF
--- a/sql/migrations/20210921152511_world.sql
+++ b/sql/migrations/20210921152511_world.sql
@@ -1,0 +1,19 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20210921152511');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20210921152511');
+-- Add your query below.
+
+-- Arcane Dust shouldn't be available on vendors.
+DELETE FROM `npc_vendor` WHERE `item` = 17019;
+
+-- End of migration.
+END IF;
+END??
+delimiter ; 
+CALL add_migration();
+DROP PROCEDURE IF EXISTS add_migration;


ratkosrb:
It must have been sold by vendors in early vanilla.
https://web.archive.org/web/20050426143917/http://wow.allakhazam.com/db/item.html?witem=17019

GrenderG:
Interesting, I guess it was sold until https://web.archive.org/web/20050430161220/http://wow.allakhazam.com/db/spell.html?wspell=23027 was removed from the game.

Daribon:
Can confirm that this is something that most likely only existed during beta.

Here is another example of an item that used to be sold by reagent vendors back in beta, that was used by an old spell:
firefox_akb6PtETEA
https://web.archive.org/web/20050115172438/http://wow.allakhazam.com/db/item.html?witem=17027

Also to clarify that an item being updated does not mean that it was last seen being sold at that date.
For example at May 24, 2006 archive it says This item is not attainable:
https://web.archive.org/web/20060524022043/http://wow.allakhazam.com/db/item.html?witem=17019
It says updated: 2005-04-16 00:58:03

At the archived date Dec 31, 2006 it says: This item is not attainable
https://web.archive.org/web/20061231222736/http://wow.allakhazam.com/db/item.html?witem=17019
But it says updated: 2006-10-19 00:27:47

However it might be that someone had it in its inventory when uploading data and that's how it got updated, which would imply that it existed in early patches to buy from vendors.

As for items sold by vendor on Allakhazam, sadly it doesn't display last updated.
At least one good thing is that considering those items were pretty much sold by every single reagent vendor in the game at some point, any image displaying a reagent vendor from early patch would suffice to confirm its existence.

ratkosrb:
So it seems at some point in the past, I deduced this item was still sold in game before patch 1.10. And wowhead is showing classic data which is using 1.12.0 data, so it does not disprove that this item was potentially in the game before 1.10. Thus I've decided to close this pull request.
https://github.com/vmangos/core/blob/a01720f547f0d042b7989085a1a3558cc6e92a64/sql/old_migrations/20190304084505_world.sql#L224

PR #1359, opened 2021-10-11
Loatheb should not instantly remove Vampiric Embrace before TBC prepatch.

Velite12:
Loatheb does not automatically remove Vampiric Embrace before TBC, source: https://wowpedia.fandom.com/wiki/Loatheb_(Classic). He currently does so I have disabled him from doing so in Vanilla, as this was a hotfix introduced in TBC.

I am removing the TBC prepatch behavior, that is currently in the core, with this pull request.

  Shadow priests are excellent for healing groups with current talents (as of 4 January 2007) as shown by Death and Taxes, 
  who successfully managed to kill him with five people (http://www.dtguilds.com/forums/showthread.php?t=5534). 
  Movie: http://www.youtube.com/watch?v=raaLqFelbCk. 
  This was hotfixed later, and Loatheb dispels the Vampiric Embrace off of himself shortly after applied. 
  Vampiric Embrace triggers the Corrupted Mind debuff.


Daribon:
4 January 2007 is from tbc pre patch.
2.0.1: Vampiric Embrace is now considered a Magic effect instead of a Curse effect.

Velite12:
Well, that is after 1.12.1, the last patch of Vanilla, so it's officially 2.0 (tbc), I can make it clear but it doesn't really change anything.

Loatheb will still remove vampiric embrace but only when he decides to remove curses, which is the correct behavior

https://youtu.be/2ZeGkz6d9W0

There are a plethora of videos from classic showing shadow priests sustaining vampiric embrace on Loatheb too.

I get your point though, Blizzard implemented the current behavior in TBC since it became a magic effect and was exploited.

ratkosrb:
I'm confused. Does this pull request make it behave like it did during Vanilla, or during TBC?

Velite12:
The pull request makes it behave like in Vanilla. It currently behaves like TBC. The description is very clear about it.

"He currently does so I have disabled him from doing so in Vanilla". This is a Vanilla core after all...

Loatheb in vanilla should not instantly remove vampiric embrace. Period.

Velite12:
I'll make the title abit clear on what I am doing, I am removing the TBC prepatch behavior, that is currently in the core, with this pull request, which will make it behave like Vanilla.

PR #1379, opened 2021-11-02
Development DB Setup Check and full Dump

coolzoom:
## 🍰 Pull request
This action will trigger a fresh database setup process in GitHub runner to check the database installation,
- pull the latest full world database from https://github.com/brotalnia/database
- merge and apply all updates inside sql folder
after the steps are done, it will generate the full dump including world/realm/logs/characters then update to the release section as a development database dump.

this can be used for database install/conflict check, also simplify the initial setup process for the users.

example:
https://github.com/coolzoom/w1x-vmcore/releases
image
image


sql/logon.sql
```diff
@@ -28,8 +28,8 @@ CREATE TABLE IF NOT EXISTS `account` (
   `failed_logins` int(11) unsigned NOT NULL DEFAULT '0',
   `locked` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `lock_country` varchar(2) NOT NULL DEFAULT '00',
-  `last_login` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
-  `last_pwd_reset` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
+  `last_login` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `last_pwd_reset` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
ratkosrb: SQL Error (1293): Incorrect table definition; there can be only one TIMESTAMP column with CURRENT_TIMESTAMP in DEFAULT or ON UPDATE clause

coolzoom: let check on this, this "zero date" cause an error in the github runner (mysql5.7) while applying the migration so i attempted to change it.

coolzoom: I changed the sql back to the original zero date format, this action is fixed by using mysql5.6, which should work for mysql5.5 as well. the zero date inside logon.sql would cause issues for mysql5.7 and above. 
https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_zero_date
but should be another workaround.

PR #1380, opened 2021-11-02
Fix crash on stop server and compilation on OS X

smellinet:
## 🍰 Pullrequest

Fix Crash mangosd  server during the stop : missing virtual constructor
Fix compilation on OS X with clang.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


ratkosrb:
This broke build on Linux and Windows, since size_t is the same as uint64 there, and we can't have 2 constructors with exact same parameters. I've added a check to only include the size_t constructor on apple platforms.

https://github.com/vmangos/core/commit/4ee28d70ba017d63f6cd14d97f6b3fe612ed9ab1

PR #1387, opened 2021-11-17
Full Database check and dump

coolzoom:
## 🍰 Pullrequest

same request as pr#1379 https://github.com/vmangos/core/pull/1379
just a fresh submit with a fixed cleaner commit


ratkosrb:
Was the issue with the date field resolved?

coolzoom:
> Was the issue with the date field resolved?

Yes, all SQL files work fine for mysql5.5 and 5.6, for MySQL >=5.7 just need to use a valid date-time. please review the logon.sql and see whether if it is acceptable.
see https://newbedev.com/mysql-incorrect-datetime-value-0000-00-00-00-00-00

sql/logon.sql
```diff
@@ -28,8 +28,8 @@ CREATE TABLE IF NOT EXISTS `account` (
   `failed_logins` int(11) unsigned NOT NULL DEFAULT '0',
   `locked` tinyint(3) unsigned NOT NULL DEFAULT '0',
   `lock_country` varchar(2) NOT NULL DEFAULT '00',
-  `last_login` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
-  `last_pwd_reset` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00',
+  `last_login` timestamp NOT NULL DEFAULT '1970-01-01 00:00:01',
+  `last_pwd_reset` timestamp NOT NULL DEFAULT '1970-01-01 00:00:01',
```
ratkosrb: This still throws an error in MySQL 5.7 if your computer's timezone is not UTC. Instead can you do what it says here and change the sql mode github actions uses? You have to remove the "NO_ZERO_DATE" and "NO_ZERO_IN_DATE" modes.

https://dev.mysql.com/doc/refman/8.0/en/datetime.html

```
MySQL does not accept TIMESTAMP values that include a zero in the day or month column or values that are not a valid date. The sole exception to this rule is the special “zero” value '0000-00-00 00:00:00', if the SQL mode permits this value. The precise behavior depends on which if any of strict SQL mode and the NO_ZERO_DATE SQL mode are enabled; see Section 5.1.11, “Server SQL Modes”. 
```

ratkosrb: Actually with mysql 5.6 it says NO_ZERO_DATE is not even enabled by default, so it should work with zero dates.

coolzoom: yes, mysql5.6 and blow have no issue with the zero date. this is enabled starting from 5.7 and is rather annoying. looks like in the github runner have to change the mysql.ini file to disable it, so i just changed the sql file instead. if we only need to check for the database, using 5.6 is enough imo. so revert the sql change is safer.

PR #1389, opened 2021-11-17
Automatic Data Extractor and Release

coolzoom:
## 🍰 Pullrequest
here is just a 'works for me' action that may or may not help or fit the master vmangos repo, so i just made a pr see if you or some people have interests. it would trigger dbc/maps extract process after the extractor or map-related code changed. publish a new data package to the release section. not used often, but useful when it is needed

- client using small client from @Daribon  https://github.com/coolzoom/smallvanilla/  https://www.reddit.com/r/wowservers/comments/bunrcr/smallest_1121_client_ever_only_356_gb/
- trigger only when extractor or map-related file changes.
- use Github runner so no personal computer time is needed. 
- multithread extractor script using cmangos mangos-classic so credit to cmangos team.  https://github.com/cmangos/mangos-classic/blob/master/contrib/extractor_scripts/MoveMapGen.sh
- 4.5hours time consume, still better compare to my computer(~8hours)
- it does consume around 5 hours github free action time, but since we dont trigger it often, so should be fine

image

example release
image

example run history:
https://github.com/coolzoom/w1x-vmcore/runs/4238358797?check_suite_focus=true

### DIFF
--- a/.github/workflows/datagen.yml
+++ b/.github/workflows/datagen.yml
@@ -1,0 +1,194 @@
+name: Client Data Extractor
+
+on:
+  push:
+    paths:
+      - '.github/workflows/datagen.yml'
+      - 'contrib/extractor/*.*'
+      - 'contrib/vmap_assembler/*.*'
+      - 'contrib/vmap_extractor/*.*'
+      - 'contrib/mmap/*.*'
+      - 'src/game/Maps/*.*'
+      - 'src/game/vmap/*.*'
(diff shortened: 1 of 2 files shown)


ratkosrb:
We can't do this, I'm pretty sure the client files are copyrighted. Uploading the whole game to github is piracy.

coolzoom:
alright, oh yes, didn’t realize that.

PR #1398, opened 2021-11-22
Update Rogue Trainer Gossips

NickTyrer:

### DIFF
--- a/sql/migrations/20211122153748_world.sql
+++ b/sql/migrations/20211122153748_world.sql
@@ -63,6 +63,8 @@ DELETE FROM `gossip_menu_option` WHERE `menu_id` = 436 AND `id` = 1;
 
 -- Creature 4163
 UPDATE `gossip_menu` SET `text_id` = 4795 WHERE `entry` = 4576 AND `text_id` = 4794;
+INSERT INTO `gossip_menu` (`entry`, `text_id`) VALUES
+(4576, 4795);
 
 -- Creature 4214
 UPDATE `creature_template` SET `gossip_menu_id` = 4577 WHERE `entry` = 4214;
--- a/sql/migrations/20211122153748_world.sql
+++ b/sql/migrations/20211122153748_world.sql
@@ -108,10 +108,10 @@ INSERT INTO `gossip_menu_option` (`menu_id`, `id`, `option_icon`, `option_text`,
 
 -- Creature 4582
 UPDATE `gossip_menu` SET `text_id` = 5993 WHERE `entry` = 4542 AND `text_id` = 5996;
-UPDATE `gossip_menu_option` SET `option_broadcast_text` = 7563, `option_text` = 'I would like to train.' WHERE `menu_id` = 4542 AND `id` = 0;
+UPDATE `gossip_menu_option` SET `option_broadcast_text` = 2548, `option_text` = 'I would like to train.' WHERE `menu_id` = 4542 AND `id` = 0;
 
 -- Creature 4584
-UPDATE `gossip_menu_option` SET `option_broadcast_text` = 7563, `option_text` = 'I would like to train.' WHERE `menu_id` = 4541 AND `id` = 0;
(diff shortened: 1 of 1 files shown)


sql/migrations/20211122153748_world.sql
```diff
@@ -0,0 +1,136 @@
...
+
+-- Creature 3594
+DELETE FROM `gossip_menu_option` WHERE `menu_id` = 4690 AND `id` = 1;
+
+-- Creature 3599
+UPDATE `gossip_menu` SET `text_id` = 934 WHERE `entry` = 436 AND `text_id` = 4794;
+DELETE FROM `gossip_menu_option` WHERE `menu_id` = 436 AND `id` = 1;
+
+-- Creature 4163
+UPDATE `gossip_menu` SET `text_id` = 4795 WHERE `entry` = 4576 AND `text_id` = 4794;
```
ratkosrb: You are replacing text id 4794 with 4795, but both of those are part of gossip menu 4576 according to sniff.
Untitled

sql/migrations/20211122153748_world.sql
```diff
@@ -0,0 +1,136 @@
...
+(4575, 4, 0, '<Take the letter>', 8548, 1, 1, -1, 0, 16, 0, 0, '', 0, 198),
+(4575, 9, 0, 'Here, I\'d like to give you this token of my love.', 11723, 1, 1, -1, 0, 410, 0, 0, NULL, 0, 462),
+(4575, 10, 0, 'Here, I\'d like to give you this token of my love.', 11723, 1, 1, -1, 0, 411, 0, 0, NULL, 0, 461);
+
+-- Creature 4582
+UPDATE `gossip_menu` SET `text_id` = 5993 WHERE `entry` = 4542 AND `text_id` = 5996;
+UPDATE `gossip_menu_option` SET `option_broadcast_text` = 7563, `option_text` = 'I would like to train.' WHERE `menu_id` = 4542 AND `id` = 0;
+
+-- Creature 4584
+UPDATE `gossip_menu_option` SET `option_broadcast_text` = 7563, `option_text` = 'I would like to train.' WHERE `menu_id` = 4541 AND `id` = 0;
```
ratkosrb: Broadcast Text ID 7563 has an emote assigned, so I think this text is used in npc_text, and for the gossip option you should use 2548.

sql/migrations/20211122153748_world.sql
NickTyrer: text 4795 seems to be her default rogue greeting, not sure how 4794 is activated.

sql/migrations/20211122153748_world.sql
NickTyrer: Not sure what you mean

ratkosrb: Look at broadcast text id 7563. It has emote_id1 = 1. Blizzard only assigns emotes to broadcast texts that are used in npc_text. So i suggest yo should assign broadcast id 2548 instead to this gossip option, which is the exact same text, but with no emote (gossip option cannot have emote, its clicked by player).

sql/migrations/20211122153748_world.sql
```diff
@@ -64,7 +64,7 @@ DELETE FROM `gossip_menu_option` WHERE `menu_id` = 436 AND `id` = 1;
 -- Creature 4163
 UPDATE `gossip_menu` SET `text_id` = 4795 WHERE `entry` = 4576 AND `text_id` = 4794;
 INSERT INTO `gossip_menu` (`entry`, `text_id`) VALUES
-(4576, 4795);
+(4576, 4794);
```
ratkosrb: Don't you actually want to replace text id 4793 here? Cause in sniff i see 4794 and 4795 for this menu, but we currently have 4793 and 4794.

NickTyrer: Yes, 4795 is the default greeting for rogues not 4794 (I dont know under what condition 4794 should unlock).

PR #1405, opened 2021-12-02
Add Charlga Razorflank Missing Sounds

NickTyrer:
## 🍰 Pullrequest


ratkosrb:
The sounds are already assigned to the broadcast text entries, so why add a separate script action to play them? They will be played twice now.

SELECT *
FROM mangos.broadcast_text
WHERE entry IN (6179, 6181, 6182, 6183, 6180)

PR #1413, opened 2021-12-05
Add Missing Lordaeron Citizen Spawns

NickTyrer:


ratkosrb:
@NickTyrer  I have redone this using sniff data. There were several issues:
- You added 38 spawns, but there are only 13 total. That means you had duplicates and added same spawn multiple times (the mobs wander). I have killed them in Classic and waited for them to respawn to get the correct spawn positions.
- You had respawn time set to 300 seconds, but i measured it to be min 360 max 480.
- You had the wander distance set to 5 yards, but its clearly more than that even if you observe them move by eye in Classic.
- You used the wrong aura. In sniff its spell id 4986. I dunno where you got 16380 from, but its 4986 in both Classic-Era and Classic-TBC.
- You changed the detection_range, but those are not values from sniffs. The value 20 which you copied is the placeholder default value in cmangos that is assigned to all but 14 creatures. I have measured it in Classic, and the default detection range for normal mobs outside of starting areas is 18 yards. You should not submit updates for the detection range based on cmangos data when its 20 there, that's a placeholder value.

Issue #1334, opened 2021-09-12
🐞 [Bug] Cannibalize


schell244:
## 🐞 Bug report - Spell

When activating spell Cannibalize while a valid corpse is around, the client will show 'Interrupted' message, but the spell works.

https://classic.wowhead.com/spell=20577/cannibalize 

![WoWScrnShot_091221_112923](https://user-images.githubusercontent.com/9536270/132983187-1ae5d013-04b2-4b5f-9355-8feeba1f79b4.jpg)

### Expected behavior
 Cannibalize with a valid corpse -> no message
 Cannibalize while no valid copse around -> "There are no nearby corpses to eat" (works)

### Related line in code:
https://github.com/vmangos/core/blob/2c831394672aa02bfbce14005cb2b9d953610310/src/game/Spells/Spell.cpp#L3623
switch (m_spellState) -> SPELL_STATE_DELAYED -> SendInterrupted(0);

### Version & Environment

Client Version: 
  - 1.12.1.5875

Commit Hash:  
619be98

Operating System:
  - Win 10

Wall-core:
This also seemingly affects spells like Tranquility

schell244:
Just noticed that [SPELL_ATTR_EX4_CAN_CAST_WHILE_CASTING](https://github.com/vmangos/core/blob/8f883d991183416e3d59e506501537728137a3f1/src/game/Spells/SpellDefines.h#L864) is not implemented in vmangos.
Flag is definded but not used anywhere in core.
[CMangos](https://github.com/cmangos/mangos-classic/blob/b2734d913f87410b153a8e6775ed50eb3ee0cdef/src/game/Spells/SpellDefines.h#L181) has this flag implemented and the issue mentioned above does not occur there.
I Have not checked yet if one of those spells below is using this flag (due to flags on wowhead this flag is not included)
[Cannibalize Dummy](https://www.wowhead.com/classic/spell=20577/cannibalize)
[Cannibalize Buff/Aura](https://www.wowhead.com/classic/spell=20578/cannibalize)

[Update] Sadly not relevant to fix this:
20577 flags: `SPELL_ATTR_EX_NO_AUTOCAST_AI, SPELL_ATTR_EX2_ALLOW_DEAD_TARGET`
20578 flags: `SPELL_ATTR_EX_NAME_IN_CHANNEL_BAR, SPELL_ATTR_EX_IS_SELF_CHANNELED`

ratkosrb:
[...]

PR #1422, opened 2021-12-18
Core/Spell - move corpse check for cannibalize to Spell::CheckCast

schell244:
## 🍰 Pullrequest
Improvements for spell https://classic.wowhead.com/spell=20577

- move cannibalize corpse check into `SpellCastResult Spell::CheckCast(bool strict)` like cmangos and remove cooldown hack

- ~~prevent cannibalize sending interrupted message, when cast is succesful (I have no idea how to fix this appart from checking for the spell id, any suggestions are welcome)~~ since checking for the spell id was not the fix for the actual problem, I've removed that change

### Proof
https://github.com/cmangos/mangos-classic/blob/5f8b6afb732237f8973d2c475eb10d8eae4839fc/src/game/Spells/Spell.cpp#L4877

https://github.com/cmangos/mangos-classic/blob/3b534fd702aac54be554a0465d7dcda05665c9ce/src/game/Spells/SpellEffects.cpp#L845

### Issues
- relates https://github.com/vmangos/core/issues/1334

### How2Test
- create undead character
- use cannibalize on a valid dead corpse
- ~~cannibalize is casted -> player should not see interrupted message~~
- cannibalize should work like before


schell244:
Closing this for now, I might do further research on this soon.

PR #1424, opened 2021-12-19
Naxxramas/Patchwerk - rewrite target selection for spell Hateful Strike

schell244:
## 🍰 Pullrequest
Rewrite target selection for spell Hateful Strike [https://classic.wowhead.com/spell=28308/hateful-strike] used by Patchwerk according to linked sources.
Remove old constexpr for meele distance check
Set Hateful Strike timer to 1,5 - 2 seconds, instead of 1,2 second

### Proof
- https://classic.wowhead.com/guides/patchwerk-naxxramas-raid-strategy
- https://youtu.be/fn9I8afey7w?t=822
- https://youtu.be/bmpVXEQYIcg?t=53

### Issues
- None

### How2Test
- Encounter Patchwerk in Naxxramas with a group
- Patchwerk should use Hateful Strike on a target if it matches the following conditions:
 1) checks for players between second, third, and fourth on threat list
 2) only players within melee range
 3) player with the most health (which has passed two contitions above) OR Hateful Strike will target main tank if second, third, and fourth players form threat list are not within meele range.

### Todo / Checklist
- [x] tested ingame
- [ ] confirm that new Hateful Strike timer is correct
- [ ] Hateful Strike meele range correct? (See no longer used constexpr which is removed with this PR)
- [x] What happes when second, third, and fourth players form threat list are not within meele range. Will main tank be targeted from Hateful Strike? - Due to https://old.reddit.com/r/classicwow/comments/jw3vlt/patchwerk_hateful_strike_clarification/ and the damge values https://youtu.be/pjUv7HfOPbE?t=563 main tank will be targeted.
- [ ] can Hateful Strike target anything other than players?

### DIFF
--- a/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
+++ b/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
@@ -139,22 +139,21 @@ struct boss_patchwerkAI : public ScriptedAI
             if (threatListPosition > 0)
             {
                 // Only target players
-                if (!iter->getUnitGuid().IsPlayer())
-                    continue;
-
-                Player* pTempTarget = m_creature->GetMap()->GetPlayer(iter->getUnitGuid());
-                if (!pTempTarget)
-                    continue;
-
-                // Player must be within map and within melee range
-                if(!m_creature->IsInMap(pTempTarget) || !m_creature->CanReachWithMeleeSpellAttack(pTempTarget))
-                    continue;
-
-                // Check if target has higher hp than anyone checked so far
-                if (pTempTarget->GetHealth() > uiHighestHP)
+                if (iter->getUnitGuid().IsPlayer())
                 {
-                    pTarget = pTempTarget;
-                    uiHighestHP = pTarget->GetHealth();
+                    if (Player* pTempTarget = m_creature->GetMap()->GetPlayer(iter->getUnitGuid()))
+                    {
+                        // Check if selected player is within melee range
+                        if (m_creature->IsInMap(pTempTarget) && m_creature->CanReachWithMeleeSpellAttack(pTempTarget))
+                        {
+                            // Check if target has higher hp than anyone checked so far
+                            if (pTempTarget->GetHealth() > uiHighestHP)
+                            {
+                                pTarget = pTempTarget;
+                                uiHighestHP = pTarget->GetHealth();
+                            }
+                        }
+                    }
                 }
             }
             threatListPosition++;
--- a/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
+++ b/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
@@ -40,8 +40,6 @@ enum PatchwerkData
     SPELL_SLIMEBOLT       = 32309  // Added in patch 1.12
 };
 
-constexpr float MELEE_DISTANCE = 5.0; 
-
 enum ePatchwerkEvents
 {
     EVENT_BERSERK       = 1,
--- a/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
+++ b/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
@@ -128,7 +128,7 @@ struct boss_patchwerkAI : public ScriptedAI
         for (const auto iter : tList)
         {
             // Do not check further than fourth form threat list
-            if (threatListPosition > 4)
+            if (threatListPosition > 3)
                 break;
 
             // Skip current main tank for inital target search
--- a/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
+++ b/src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
@@ -17,7 +17,7 @@
 /* ScriptData
 SDName: Boss_Patchwerk
 SD%Complete: 80
-SDComment: TODO: confirm how hateful strike work
+SDComment:
 SDCategory: Naxxramas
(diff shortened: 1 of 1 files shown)


ratkosrb:
It already works how it should, I updated it based on information from Classic.
https://github.com/vmangos/core/commit/ca8f5d3e1b9ec6f4e95e8b964078ddc659fea2a4

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
```diff
@@ -119,68 +119,55 @@ struct boss_patchwerkAI : public ScriptedAI
...
+        uint32 uiHighestHP       = 0;
         uint8 threatListPosition = 0;
 
         ThreatList const& tList = m_creature->GetThreatManager().getThreatList();
         for (const auto iter : tList)
         {
-            // Only top 4 players on threat in melee range are targetted.
-            if (threatListPosition > 3)
+            // Only second, third, and fourth form threat list can be targeted
+            if (threatListPosition > 4)
```
ratkosrb: Now it will target top 5, since we count from 0...

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
```diff
@@ -119,68 +119,55 @@ struct boss_patchwerkAI : public ScriptedAI
...
-                if (pTempTarget->GetHealth() > uiHighestHP)
+                // Only target players
+                if (iter->getUnitGuid().IsPlayer())
                 {
-                    pTarget = pTempTarget;
-                    uiHighestHP = pTarget->GetHealth();
+                    if (Player* pTempTarget = m_creature->GetMap()->GetPlayer(iter->getUnitGuid()))
+                    {
+                        // Check if selected player is within melee range
+                        if (m_creature->IsInMap(pTempTarget) && m_creature->CanReachWithMeleeSpellAttack(pTempTarget))
```
ratkosrb, line 143: What's the point of moving all the checks into nested ifs? Just makes the code more indented.

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
schell244: Due to the sources I've linked, it should only target players between second, third, and fourth from threat list. So index 0 should be skipped. (There's a if which does that in below conditions).

ratkosrb: It does not sound plausible to me that it cant hit the main tank, otherwise the strategy would be to only send 1 tank and no melee dps, and the fight would be piss easy, literally tank and spank.

ratkosrb: https://old.reddit.com/r/classicwow/comments/jw3vlt/patchwerk_hateful_strike_clarification/

schell244: Interresting. At https://classic.wowhead.com/guides/patchwerk-naxxramas-raid-strategy they claim main tank can't be targeted: _Understanding these three mechanics is key. It will never hit the person with the most threat, so that means the Main Tank is only taking damage from auto attacks._

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
schell244: I did it because `threatListPosition++;`  would have to be added above every `continue;` statement, othewise the index check won't work properly.

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
schell244: Just found this video where all meeles are dead except the main tank https://youtu.be/pjUv7HfOPbE?t=563 but I can't tell if the tank only gets meele hits (maybe with crits) or also hateful strikes.

schell244: After watching the video and doing some tests with vmangos, Patchwerk must be using hateful strike on the main tank, if no one else (from position 2-4 on thread list) is available in meele range. I've tested the fight on vmangos without the tank being targeted with hateful strike, and the damage on my test character never reached those high values like what the tank in the video had to cope with. The frequency of high damage occuring in the video seems to fit to a 1,5-2 seconds timer, like mentioned at wowhead.

src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/boss_patchwerk.cpp
```diff
@@ -111,76 +108,71 @@ struct boss_patchwerkAI : public ScriptedAI
...
-            // Skipping maintank, only using him if there is no other viable target 
-            // todo: not sure if this is correct. Should we target the MT over the offtanks, if the offtanks have less hp?
-            if (iter->getUnitGuid() != mainTankGuid)
+            // Skip current main tank for inital target search
+            if (threatListPosition > 0)
             {
-                // target has higher hp than anyone checked so far
-                if (pTempTarget->GetHealth() > uiHighestHP)
+                // Only players
+                if (iter->getUnitGuid().IsPlayer())
```
scottnice, line 138: Can pets on the threat list be targeted?

schell244: I've just added this to the remaing TODO's. 
(With this change Hateful Strike can only hit a pet, if pet is main tank and no player from threat list place 2,3,4 is within meele range)

ghost:
Status?

schell244:
Closing this for now due to lacking proofs/information. Script already works pretty good though.

PR #1446, opened 2022-01-14
Updated database scripts to fix voodoo feathers quest

### LINKED ISSUE #1442: 🐝 [Bug] Missing Quest End Script Voodoo Feathers (8425)
## 🐝 Bug report - Quest
The questgiver should say some text and give you a buff
see: https://classic.wowhead.com/quest=8425/voodoo-feathers#comments

Commit Hash:  1fd3946

scottnice:
## 🍰 Pullrequest

Updated database scripts to fix voodoo feathers quest
After the voodoo feathers quest is completed the fallen hero of the horde will now correctly transform into non-ghost form, cast battleshout, say some text, kneel, stand, and transform back into a ghost with an animation

### Proof

- https://youtu.be/la2GTYLeiak?t=473

### Issues

- fixes #1442

### How2Test

- Turn in quest

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20220114190344_world.sql
+++ b/sql/migrations/20220114190344_world.sql
@@ -1,0 +1,35 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20220114190344');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20220114190344');
+-- Add your query below.
+
+DELETE FROM `quest_end_scripts` WHERE `id` = 7572;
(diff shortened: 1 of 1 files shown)


NickTyrer:
Looking at the video, you should remove the gossip and questgiver flags at the start of the script and add them at the end.

scottnice:
> Looking at the video, you should remove the gossip and questgiver flags at the start of the script and add them at the end.

Updated

ratkosrb:
Some issues that I fixed in the migration:
- It's best to use the explicit "Add Flags" and "Remove Flags" mode for the "Modify Flags" command. There's a chance things get messed up with the "Toggle" option.
- When you want creature to cast a spell on itself, always add the "Target Self" flag, otherwise it will throw an error if the player who turned in the quest disappears before the script ends.
- The id of quest scripts should match the id of the quest itself, not use arbitrary numbers. The core even warns you about that on startup.

scottnice:
> Some issues that I fixed in the migration:
> 
>     * It's best to use the explicit "Add Flags" and "Remove Flags" mode for the "Modify Flags" command. There's a chance things get messed up with the "Toggle" option.
> 
>     * When you want creature to cast a spell on itself, always add the "Target Self" flag, otherwise it will throw an error if the player who turned in the quest disappears before the script ends.
> 
>     * The id of quest scripts should match the id of the quest itself, not use arbitrary numbers. The core even warns you about that on startup.

@ratkosrb thank you for the info.

PR #1452, opened 2022-01-23
Stay with Windows Server 2019 +  VS 2019

coolzoom:
windows-latest has been updated to Windows Server 2022 and VS 2022 today.
https://github.com/actions/virtual-environments/issues/4856
new commits check flow would fail for "vs2019 not found".

There is not a stable ACE for vs2022 yet so we should stay at Win2019+VS2019 right now.

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -15,7 +15,7 @@ on:
    
 jobs:
   build:
-    runs-on: windows-latest
+    runs-on: windows-2019
     
     steps:
     #git checkout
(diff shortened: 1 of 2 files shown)


al3xc1985:
You can rewrite the whole system and remove ace and this kind of problems will dissapear

ratkosrb:
lol

al3xc1985:
Joking. Don:t get it serious

Zeebler:
Is it possible you guys can update the guide to compiling on windows and with window server 2019 + vs2019?

ratkosrb:
The steps would be the same. Just have to use a compatible version of the dependencies.

PR #1464, opened 2022-02-19
Love is in the Air Event Additions / Fixes

Oroxzy:
**ToDo:**
- [x] Implement all Winning events / spawns.
- [x] Respawn Main (8) event objects due some missing and wrong ones.
- [x] Check all Amorous affected NPCs.

**Facts:**
- Each city has its own Trap GameObject (not sniffable but from TC bruteforce) "Event Trap". Its targeted by a spell which Peddle is casting if a Player turns in a gift, so its obvious that those GameObject are handling the whole outcome for the Winner events. In vmangos the whole handling is set in Peddles c++ script. I've added the GameObjects but i do not use them right now.
- I sniffed all Winner events in each City separately and after Love is In the Air, so there is no doubt every GameObject is correct and not mixed up with Objects from the main event (unless Blizzard screwed it up again).
- Peddle still had the Quest flag and NPCs in Cities the Amorous auras during the Winning event on some Classic ERA Realms. Im not sure why it was not on all realms. I think it was a bug and it should not be possible to bring in gifts after the main event has ended.
- If a Classic Realm had 0 votes, the Winner event spawns in Stormwind and Thunderbluff (idk if its a bug or not). But this was on every realm with 0 votes.
unknown
- The winning event is meant to stay only a week after the end of Love is in the Air, it was a bug in vanilla where it stayed for 1 year until the start of the next Love is in the Air. 

**WoW Classic ERA EU/RU Final standings of the Popularity Contest**
grafik


ratkosrb:
```
The winning event is meant to stay only a week after the end of Love is in the Air, it was a bug in vanilla where it stayed for 1 year until the start of the next Love is in the Air.
```
That "bug" was not fixed in Vanilla. So it should work like this in vmangos, not the changed Classic behavior.

PR #1470, opened 2022-02-28
A Game of Towers - Curing Shrine Fix

Oroxzy:
Makes Curing Shrine usable again after 0.5 seconds (from TC bruteforced GameObjects)

### DIFF
--- a/sql/migrations/20220228181818_world.sql
+++ b/sql/migrations/20220228181818_world.sql
@@ -8,7 +8,7 @@ IF v=0 THEN
 INSERT INTO `migrations` VALUES ('20220228181818');
 -- Add your query below.
 
--- Makes Curing Shrine usable again after 0.5 seconds (from TC brutforced GameObjects)
+-- Makes Curing Shrine usable again after 0.5 seconds (from TC bruteforced GameObjects)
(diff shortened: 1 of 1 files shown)


sql/migrations/20220228181818_world.sql
```diff
@@ -0,0 +1,20 @@
...
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20220228181818');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20220228181818');
+-- Add your query below.
+
+-- Makes Curing Shrine usable again after 0.5 seconds (from TC bruteforced GameObjects)
+UPDATE `gameobject_template` SET `data3`=500 WHERE `entry`=181955;
+UPDATE `gameobject_template` SET `data3`=500 WHERE `entry`=181682;
```
ratkosrb, line 13: Is the gobject not there in mass parse or wdb parses?

Oroxzy: it is, with the same value as before.

ratkosrb: So you are replacing the classic data with mists of pandaria data from tc?

pjaszko:
image
how i can fix this?

PR #1476, opened 2022-03-11
Aura: Use m_miscvalue in HandleModDamagePercentDone for correct application

schell244:
## 🍰 Pullrequest
- Credit: cmangos https://github.com/cmangos/mangos-classic/commit/3d1c0723b398ea22e704cf43a8bcc5c2c4d30a87
- additionally some improvements for redability

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -5292,14 +5292,27 @@ void Aura::HandleModDamageDone(bool apply, bool Real)
 void Aura::HandleModDamagePercentDone(bool apply, bool Real)
 {
     DEBUG_FILTER_LOG(LOG_FILTER_SPELL_CAST, "AURA MOD DAMAGE type:%u negative:%u", m_modifier.m_miscvalue, m_positive ? 0 : 1);
+
     Unit* target = GetTarget();
+    if (!target)
+    {
+        return;
+    }
 
     // apply item specific bonuses for already equipped weapon
-    if (Real && target->GetTypeId() == TYPEID_PLAYER)
+    if (Real && target->IsPlayer())
     {
-        for (int i = 0; i < MAX_ATTACK; ++i)
-            if (Item* pItem = ((Player*)target)->GetWeaponForAttack(WeaponAttackType(i), true, false))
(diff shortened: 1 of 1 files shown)


src/game/Spells/SpellAuras.cpp
```diff
@@ -5332,7 +5347,9 @@ void Aura::HandleModDamagePercentDone(bool apply, bool Real)
 
     // Skip non magic case for speedup
     if ((m_modifier.m_miscvalue & SPELL_SCHOOL_MASK_MAGIC) == 0)
+    {
         return;
+    }
```
ratkosrb: Do not add brackets around single statements. This is not our code style.

src/game/Spells/SpellAuras.cpp
```diff
@@ -5292,14 +5292,27 @@ void Aura::HandleModDamageDone(bool apply, bool Real)
...
 
     // apply item specific bonuses for already equipped weapon
-    if (Real && target->GetTypeId() == TYPEID_PLAYER)
+    if (Real && target->IsPlayer())
     {
-        for (int i = 0; i < MAX_ATTACK; ++i)
-            if (Item* pItem = ((Player*)target)->GetWeaponForAttack(WeaponAttackType(i), true, false))
-                ((Player*)target)->_ApplyWeaponDependentAuraDamageMod(pItem, WeaponAttackType(i), this, apply);
+        if (Player* player = target->ToPlayer())
```
ratkosrb, line 5301: No need to check IsPlayer above if using ToPlayer(). It will return null if its not a player, and it will not enter the if block, since assigning null results in false.

src/game/Spells/SpellAuras.cpp
```diff
@@ -5292,14 +5292,27 @@ void Aura::HandleModDamageDone(bool apply, bool Real)
 void Aura::HandleModDamagePercentDone(bool apply, bool Real)
 {
     DEBUG_FILTER_LOG(LOG_FILTER_SPELL_CAST, "AURA MOD DAMAGE type:%u negative:%u", m_modifier.m_miscvalue, m_positive ? 0 : 1);
+
     Unit* target = GetTarget();
+    if (!target)
+    {
+        return;
```
ratkosrb: Aura target can never be null. This check is pointless.

PR #1478, opened 2022-03-11
Improve handling of stat modification

schell244:
## 🍰 Pullrequest
Remove hack for stat modifier calcualtion. 
Credit: Azerothcore

### Proof
- https://github.com/azerothcore/azerothcore-wotlk/blob/a775ddcb5b935be3358644be27df282fc29a9936/src/server/game/Entities/Unit/Unit.cpp#L14465
- https://github.com/azerothcore/azerothcore-wotlk/blob/a775ddcb5b935be3358644be27df282fc29a9936/src/server/game/Entities/Player/Player.cpp#L4907

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Objects/Player.cpp
+++ b/src/game/Objects/Player.cpp
@@ -5333,8 +5333,6 @@ void Player::HandleBaseModValue(BaseModGroup modGroup, BaseModType modType, floa
         return;
     }
 
-    float val = 1.0f;
-
     switch (modType)
     {
         case FLAT_MOD:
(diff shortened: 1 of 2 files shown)


ratkosrb:
And what does this do?

schell244:
This is just a cosmetic change. This function does basically the same (only difference is that -100.0f treated as -99,99f to prevent zero division)
```
inline void ApplyPercentModFloatVar(float& var, float val, bool apply)
{
    if (val == -100.0f)     // prevent set var to zero
        val = -99.99f;
    var *= (apply?(100.0f+val)/100.0f : 100.0f / (100.0f+val));
}
```

ratkosrb:
So by removing it, you allow for server to crash when division by zero occurs?

ghost:
Status?

schell244:
Works fine for me, but just code cosmetics - won't notice any change ingame.
(I saw that change at azerothcore while comparing code to look for a solution for https://github.com/vmangos/core/issues/1236)

ratkosrb:
Lets see what happens.

PR #1485, opened 2022-03-23
fix BattleGroundMgr::CreateClientVisibleInstanceId() bugs

4demec:
current implementation's for-loop never incremented its *itr, and also incremented lastId too much
this means that bgs created will have the numbers 1, 3, 3, 3... etcetera, and any bgs after the second will fail to be listed, since the "ids" set can only have unique entries.

**How2Test:**
Create at least 4 battlegrounds of the same type, in the same bracket using at least 4 different accounts.
Use partybots to fill each battleground to max players, or modify the max_players_per_team column in the battleground_templates database table to be = 1.
Alternatively, use the .debug bg ingame GM command, but this will only allow you to assert 1) in my experience, as battlegrounds don't seem to show up in the list that way.
1) Assert that each battleground instance is enumerated correctly in the "enter" popup
2) Assert that each battleground instance is enumerated correctly in the given battlemaster's list of active battlegrounds.


ratkosrb:
You were interrupting the loop on the first iteration, so the ids were only going up to 2.
https://github.com/vmangos/core/commit/49a9c295d980a54f4515886bae734bf48937e1f8

PR #1489, opened 2022-04-05
Remove extra cooking table from razor hill

Makkonii:
- Removed one of two cooking tables that were inside each other on razor hill at .go 316.052399 -4666.528320 17.332701 1

### Proof
-  Razor Hill used to contain 2 cooking tables on vmangos using .gobject near got the guid of both tables.
- Using ReplayCore  with  combined_sniffdb parses that there should only be one cooking table and what GUID it has and it matched one of the cooking tables on vmangos.
- Also checked on Season of mastery that only one cooking table is present.

### How2Test
.go 316.052399 -4666.528320 17.332701 1
.gobject near
Now contains only cooking table with guid of 18076.

### DIFF
--- a/sql/migrations/20220405061736_world.sql
+++ b/sql/migrations/20220405061736_world.sql
@@ -1,0 +1,16 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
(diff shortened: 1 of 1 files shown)


nellyeo (approved):
Still exists on latest build and fixes the issue without any conflict

PR #1497, opened 2022-05-01
On completion of the Arena event in BRD the mobs spectating the Arena will become neutral

### LINKED ISSUE #1495: 🦂 [Bug] All NPC's above the Ring of Law should become neutral.
## 🦂 Bug report - Dungeon
Spectators above the Ring of Law in Blackrock Depths should become neutral after the event. Most NPC's are currently hostile.
[Picture of one of my players seeing hostile NPC.](https://i.ibb.co/0GS7wt1/brdringoflaw.png)

### Expected behavior
All bystanders of Ring of Law should become neutral after completing the event.
[Here is a guide to the event, ](https://www.wow-pro.com/48-60-james-blackrock-depths-guide-brd/).scroll to Section 2 Ring of Law, where it states:
```
[...]it also puts every arena spectator in non-agressive mode, which is handy because you will have to cross the upper part of the arena.
```
Some mobs are named `Arena Spectator`, but it does apply to ALL mobs above the ring, spectating the event.
Anvilrage Officer
Anvilrage Soldier
Shadowforge Citizen
Shadowforge Peasant
Arena Spectator

### Steps to reproduce

No GM commands needed. Teleport to BRD. Go inside past the jail and to the Ring of Law and complete the event.

### Version & Environment
Client Version: 
1.12.1.5875
Have not tested the other versions.

OS Client: Windows, expected to happen on other clients aswell. But not tested.
OS Server:  Arch Linux

### Crashlog
- None

scottnice:
On completion of the Arena event in BRD the mobs spectating the Arena will become neutral. This code is similar to the code from CMangos.

## 🍰 Pullrequest
On completion of the Arena event in BRD the mobs spectating the Arena will become neutral. If they are killed they will respawn with original faction settings. This code is similar to the code from CMangos. https://github.com/cmangos/mangos-classic/blob/master/src/game/AI/ScriptDevAI/scripts/eastern_kingdoms/blackrock_depths/instance_blackrock_depths.cpp
### Proof

- None

### Issues
- fixes #1495 

### How2Test

- Complete ring of law event and see faction of mobs spectating. Also restart server to make sure they maintain their faction.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/blackrock_depths/blackrock_depths.h
+++ b/src/scripts/eastern_kingdoms/burning_steppes/blackrock_depths/blackrock_depths.h
@@ -63,6 +63,7 @@ enum
     EVENT_BAR_PATRONS          = 100,
 
     FACTION_DARK_IRON          = 54,
+    FACTION_ARENA_NEUTRAL      = 15,
 
     NPC_EMPEROR                = 9019,
     NPC_PRINCESS               = 8929,
@@ -128,8 +129,12 @@ enum
 
     NPC_FIREGUARD_DESTROYER    = 8911,
     NPC_ANVILRAGE_OFFICER      = 8895,
+    NPC_ANVILRAGE_SOLDIER      = 8893,
+    NPC_ANVILRAGE_MEDIC        = 8894,
 
     NPC_SHADOWFORGE_SENATOR    = 8904,
+    NPC_SHADOWFORGE_PEASANT    = 8896,
+    NPC_SHADOWFORGE_CITIZEN    = 8902,
     NPC_GRIMSTONE              = 10096,
     NPC_THELDREN               = 16059,
 
@@ -182,5 +187,16 @@ static float const aBarPatrolPositions[2][4] = {
 };
 
 static uint32 const aBarPatrolId[3] = {NPC_FIREGUARD_DESTROYER, NPC_ANVILRAGE_OFFICER, NPC_ANVILRAGE_OFFICER};
+
+struct ArenaCylinder
+{
+    float m_fCenterX;
+    float m_fCenterY;
+    float m_fCenterZ;
+    uint32 m_uiRadius;
+    uint32 m_uiHeight;
+};
+
+static const ArenaCylinder aArenaCrowdVolume = { 595.78f, -188.65f, -38.63f, 69, 10 };
 
 #endif
--- a/sql/migrations/20220502012000_world.sql
+++ b/sql/migrations/20220502012000_world.sql
@@ -1,0 +1,21 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20220502012000');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20220502012000');
+-- Add your query below.
+
+-- set shadowforge senators and arena spectator faction to dark iron dwarves based on sniff data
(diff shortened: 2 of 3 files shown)


src/scripts/eastern_kingdoms/burning_steppes/blackrock_depths/blackrock_depths.h
```diff
@@ -63,6 +63,7 @@ enum
     EVENT_BAR_PATRONS          = 100,
 
     FACTION_DARK_IRON          = 54,
+    FACTION_DARK_IRON_NEUTRAL  = 674,
```
ratkosrb: In my sniffs, these mobs use only factions 15 and 54. Perhaps 15 is the neutral faction?

scottnice: Cmangos also used 15. If the sniffs say 15 I can change it to that. But looking at the database faction_template 674 uses the same faction id as template 54 but with a hostile mask removed.

image

scottnice: Do the sniffs show any neutral mobs before the arena is cleared? Currently the spectators and the senators are neutral from the start. From what I see on classic, all spectators should be hostile until the arena is cleared.

scottnice: Updated to 15

scottnice: Updated to use faction 15

scottnice: @ratkosrb sniff data showed that Shadowforge Senators and Arena Spectators were only encountered with a faction id of neutral inside of the Arena cyclinder. See image and query below. This query was run against the sniffs_combined3 database. 

`
SET @X = 595.78;
SET @Y = -188.65;
SET @Z = -38.63;
SET @RADIUS = 69*69;
SET @HEIGHT = 10;

SELECT *, COUNT(*) AS count FROM (
SELECT
case 
when c.id = 8904 then 'Shadowforge Senator'
when c.id = 8916 then 'Arena Spectator'
END AS NAME,
 c.id, c.faction, 
(POWER(c.position_x - @X, 2) + POWER(c.position_y - @Y, 2)) < @RADIUS AND ( c.position_z >= @Z AND c.position_z <= (@Z + @HEIGHT)) AS inArena FROM creature c 
WHERE c.id IN (8904, 8916)) AS a
GROUP BY a.id, a.faction, a.inArena`

image

Notice that the vast majority of sniffs show that Senators and Spectators are hostile dark iron 54 faction. Even inside of the arena the majority of sniffs for these mobs show as dark iron 54.

WockaFlockaRist:
Tested this, and there are still hostile mobs. All Shadowforge Senators and Shadowforge Spectators become neutral, but there are Anvilrage Officers, Anvilrage Medics and Anvilrage Soldiers present for the event.

scottnice:
@WockaFlockaRist are you sure? That was the behavior before the change. The patch here specifically turns neutral all soldiers officers and medics.

WockaFlockaRist:
Yes, I am new to vmangos, but I am very sure I have tested this correctly. Maybe you can poke holes in the way I am testing this. I have a vmangos server, with automatic systemd startup. I cloned the latest development branch `0fa10d10685d1` and compiled it on my server under a new directory. I changed the working directory of the systemd unit file to point to the newest compiled version. Where this commit `2c0854594e85` is part of the history. I started the server, completed the Ring of Law event and there are still hostile mobs after completing the Ring of Law event.

Is there any other way I should test this? Or is this sufficient?

PR #1511, opened 2022-05-19
Fixed compiler warnings under VS2022

ghost:
## 🍰 Pullrequest
Fixed compiler warnings under Visual Studio 2022 (toolset v143). All of the warnings were related to signed/unsigned comparison mismatch.

### How2Test
- Run compilation with VS2022

Note: There are some additional warnings in third party libraries (mostly in libmpq and g3dlite) but I didn't fix those. Perhaps ignoring specific warnings for those projects would be ideal?

### DIFF
--- a/src/game/Conditions.cpp
+++ b/src/game/Conditions.cpp
@@ -220,7 +220,7 @@ bool inline ConditionEntry::Evaluate(WorldObject const* target, Map const* map, 
         }
         case CONDITION_SAVED_VARIABLE:
         {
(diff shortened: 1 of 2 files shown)


ratkosrb:
We use east const in this project. Also wouldn't it be better to cast them to int64 not int32, cause other ways the unsigned values will turn into negative ones?

ghost:
Yeah, that would be better, I made the changes.

PR #1516, opened 2022-05-28
Fix quest Extinguishing the Idol after it was failed once

schell244:
## 🍰 Pullrequest
Reset all variables of quest https://classic.wowhead.com/quest=3525/extinguishing-the-idol when it was failed once.

### Proof
- None

### Issues
- None

### How2Test
- accept quest 3525
- follow Belnistrasz until he starts his ritual
- let Belnistrasz die and fail quest
- .respawn Belnistrasz 
- reaccept quest -> quest should work and progress like before

### Todo / Checklist
- [X] None


ratkosrb:
How is it possible to try it again after failing? I thought the npc will despawn.

schell244:
You're right, normal players will probably not encounter this bug. I failed the quest and did .respawn on is corpse and re-accepted the quest on his spawn point. Fix is more for debugging (I was looking at his broadcast texts)

PR #1519, opened 2022-06-02
Logging rewrite

namreeb:
## 🍰 Pullrequest

Groundwork for improved logging, which is in turn the groundwork for better GM tooling when investigating players

### Todo / Checklist
- [X] Remove all but one single file logging
- [X] Remove all but one single database logging of player actions
- [x] Cleanup config file settings pertaining to logging
- [x] Verify all logging related settings are properly enforced
- [x] Audit log level of all log messages
- [x] Remove redundant information from all log messages


ratkosrb:
The build is failing on Linux according to the CI.

namreeb:
> The build is failing on Linux according to the CI.

Fixed.

sql/migrations/20220601082200_logs.sql
```diff
@@ -0,0 +1,48 @@
...
+DROP TABLE IF EXISTS `logs_chat`;
+DROP TABLE IF EXISTS `logs_movement`;
+DROP TABLE IF EXISTS `logs_spamdetect`;
+DROP TABLE IF EXISTS `logs_warden`;
+
+-- New unified table
+DROP TABLE IF EXISTS `logs_player`;
+CREATE TABLE `logs_player` (
+  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
+  `time` DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
ratkosrb: **SQL Error (1067): Invalid default value for 'time'**

The plan in vmangos is to never raise the minimum requirements, and this part does not work in mysql 5.5, which is the last mysql version to run on Windows XP. 

Suggested fix:
```
`time` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

ratkosrb:
I'd like to see the ERROR part un-hardcoded. Can you add a LOG_ERROR type, and when using that or LOG_DBERROR make it automatically prefix the message with ERROR, instead of having us manually write it as part of the string. Then errors can be distinguished from other messages in code. It would be nice to be able to show all errors, but not other messages.

namreeb:
> I'd like to see the ERROR part un-hardcoded. Can you add a LOG_ERROR type, and when using that or LOG_DBERROR make it automatically prefix the message with ERROR, instead of having us manually write it as part of the string. Then errors can be distinguished from other messages in code. It would be nice to be able to show all errors, but not other messages.

Right.  I mentioned this on discord, but it makes more sense to discuss here.  I think it makes sense to restore, but I'm not quite sure how.  How does it compare to `LOG_LVL_MINIMAL`?  What we have right now is:

```cpp
enum LogLevel
{
    LOG_LVL_MINIMAL = 0,                                    // unconditional and errors
    LOG_LVL_BASIC   = 1,
    LOG_LVL_DETAIL  = 2,
    LOG_LVL_DEBUG   = 3
};
```

Note the comment.  I feel like it should maybe change to this:

```cpp
enum LogLevel
{
    LOG_LVL_ERROR = 0,
    LOG_LVL_MINIMAL,
    LOG_LVL_BASIC,
    LOG_LVL_DETAIL,
    LOG_LVL_DEBUG
};
```

If we do something like this, it would be easy to automatically prepend the "ERROR: " string (although note that that has a performance impact since it is at least one extra allocation and copy)..

sql/migrations/20220601082200_logs.sql
namreeb: Resolved by [e8687a6](https://github.com/vmangos/core/pull/1519/commits/e8687a62d119f6a32cc02c4b7c5f4f277336ab2f).

ratkosrb:
Another option is to just add a sLog.Error method that will append ERROR and then pass it again to sLog.Out.

namreeb:
> Another option is to just add a sLog.Error method that will append ERROR and then pass it again to sLog.Out.

Removing that sort of redundancy and unnecessary complexity is one of the goals of this MR.  Is that in some way preferrable to adding a `LOG_LVL_ERROR`?

ratkosrb:
@namreeb I was just trying to give you ideas. Do whatever you think is best.

namreeb:
> @namreeb I was just trying to give you ideas. Do whatever you think is best.

Alrighty!  Check out https://github.com/vmangos/core/pull/1519/commits/4d840aef6ce4da561f9958591a969da135851fee

Edit: In particular, [log.cpp](https://github.com/vmangos/core/commit/4d840aef6ce4da561f9958591a969da135851fee#diff-93eef7c860a081602c2a653fabf9e763c4384496c22aeb72cb110c0ff2d03991) contains the change.  I think it makes more sense to conditionally do two writes rather than rebuilding the string.

namreeb:
> Tabs to spaces.

Fixed in https://github.com/vmangos/core/pull/1519/commits/0e245a41db2a6d94f5703433db4d5ff3343ea166

namreeb:
This is probably ready for review now @brotalnia.  The last thing I want to do is go and find calls to `sLog.Player()` which manually include the player's position, which is no longer necessary.

src/game/Objects/Pet.h
```diff
@@ -198,9 +198,12 @@ class Pet : public Creature
         int32 GetBonusDamage() const { return m_bonusdamage; }
         void SetBonusDamage(int32 damage) { m_bonusdamage = damage; }
 
+        bool UpdateStats(Stats stat) override;
```
ratkosrb: Why does it show these functions have been re-added to Pet class?

ratkosrb: Actually the whole CLS implementation commit seems to be reverted by your pull request :O

https://github.com/vmangos/core/commit/358ab70d8892a9bf9a4cc3f0c57294b9338ce9c7

namreeb: This is probably a merge error when I rebased my branch off the current development branch.  Ugh.  I guess I'll have to go through the whole thing manually.

namreeb: This should be fixed now.

namreeb:
@ratkosrb I think this is ready for review again.  There are still things that would be nice to do.  For example the notion of "LogFilter" is sort of redundant with LogType, so it might be nice to combine them.  But I'd rather get this merged and retain the option to do that later.  That should mean less wrestling with git as I take my sweet time!

namreeb:
I don't understand the pipeline failure.  It cites outError in Unit.cpp, like 10457, but you can clearly see [here](https://github.com/namreeb/vmangos/blob/development/src/game/Objects/Unit.cpp#L10457) that its not there, or anywhere in that file.

src/game/Battlegrounds/BattleGroundAV.cpp
```diff
@@ -795,7 +795,7 @@ void BattleGroundAV::UpdateScore(BattleGroundTeamIndex teamIdx, int32 points)
 
 void BattleGroundAV::ResetTamedEvent(uint32 teamIdx)
 {
-DEBUG_LOG("BattleGroundAV: TeamIdx %i ", teamIdx);
+sLog.Out(LOG_BASIC, LOG_LVL_DEBUG, "BattleGroundAV: TeamIdx %i ", teamIdx);
```
ratkosrb: indentation

src/game/Handlers/CharacterHandler.cpp
```diff
@@ -334,10 +334,6 @@ void WorldSession::HandleCharCreateOpcode(WorldPacket& recv_data)
 
         data << (uint8)CHAR_CREATE_SUCCESS;
         SendPacket(&data);
-
-        std::string IP_str = GetRemoteAddress();
-        BASIC_LOG("Account: %d (IP: %s) Create Character:[%s] (guid: %u)", GetAccountId(), IP_str.c_str(), name.c_str(), guidLow);
-        sLog.out(LOG_CHAR, "Account: %d (IP: %s) Create Character:[%s] (guid: %u)", GetAccountId(), IP_str.c_str(), name.c_str(), guidLow);
```
ratkosrb, line 341: Why removing this>

src/game/HonorMgr.cpp
```diff
@@ -261,47 +261,47 @@ void HonorMaintenancer::DoMaintenance()
...
     ToggleMaintenanceMarker();
     SetMaintenanceDays(GetNextMaintenanceDay());
 }
 
 void HonorMaintenancer::CreateCalculationReport()
 {
-    std::string timestamp = Log::GetTimestampStr();
-    std::string filename = "HCR_" + timestamp + ".txt";
+    std::string timestamp = sLog.GetTimestampStr();
+    std::string filename = "HCR" + timestamp + ".txt";
```
ratkosrb: Why remove the underscore from the file name? It's harder to read without it.

src/shared/Log.h
```diff
@@ -158,140 +154,89 @@ class Log : public MaNGOS::Singleton<Log, MaNGOS::ClassLevelLockable<Log, std::m
...
+        // include timestamp in console output
         bool m_includeTime;
-        bool m_wardenDebug;
-        Color m_colors[LOG_TYPE_MAX];
         uint32 m_logFilter;
 
         // cache values for after initilization use (like gm log per account case)
         std::string m_logsDir;
-        std::string m_logsTimestamp;
+        const std::string m_logsTimestamp;
```
ratkosrb: East const!

src/game/Handlers/CharacterHandler.cpp
namreeb: It's redundant with the log from Player::SaveNewPlayer.

namreeb:
Okay, merge conflicts resolved and fixes made based on feedback.

coolzoom:
this pr works for the single gm log however breaks the per account gm log

1, gm log file has the redundant config
https://github.com/vmangos/core/blob/e0ffbc0f76ef5f9e55503760ce3bb267e704935b/src/shared/Log.cpp#L81
https://github.com/vmangos/core/blob/e0ffbc0f76ef5f9e55503760ce3bb267e704935b/src/shared/Log.cpp#L84
https://github.com/vmangos/core/blob/e0ffbc0f76ef5f9e55503760ce3bb267e704935b/src/shared/Log.cpp#L116
https://github.com/vmangos/core/blob/e0ffbc0f76ef5f9e55503760ce3bb267e704935b/src/shared/Log.cpp#L117

2, openGmlogPerAccount never got called
https://github.com/vmangos/core/blob/e0ffbc0f76ef5f9e55503760ce3bb267e704935b/src/shared/Log.cpp#L323

PR #1541, opened 2022-07-06
GM Settings - add cheat god to world config options

schell244:
## 🍰 Pullrequest
Set god mode state for gm's after login due to world config settings.
Default: enabled (like it was before)

### How2Test
- compile, update world config
- check if switching value in world config works (player can be killed or not)

### Todo / Checklist
- [X] None


ratkosrb:
Why is it a uint32 config option instead of a bool?

ratkosrb:
Why did u put it after VALUE_COUNT, that's supposed to represent the total count of bool config options. It should always be at the end.

PR #1580, opened 2022-08-26
Change default value of realmd.revision to VMangos

schell244:
## 🍰 Pullrequest
Previously default value was Trinitycore
Unbenannt

### Todo / Checklist
- [X] None


Kittnz:
Change it aswell in https://github.com/vmangos/core/blob/development/sql/logon.sql

PR #1608, opened 2022-09-07
Correct graveyard factions and and invalid graveyards.

Daribon:
## 🍰 Pullrequest

Corrects all graveyard factions in the game that didn't have correct faction set.
Last remaining invalid graveyards were deleted and correct graveyard added for zones affected.
### Proof

I checked every single graveyard on classic by suicide a ton of times on both factions.
My tests resulted in the information below, which this commit is based on:

|ID |AreaName_lang                             |Loc[0]    |Loc[1]    |Loc[2]    |Continent|Facing  |faction |
|---|------------------------------------------|----------|----------|----------|---------|--------|--------|
|3  |Duskwood, Darkshire                       |-10774.263|-1189.6719|33.149426 |0        |220.378 |neutral |
|4  |Westfall, Sentinel Hill                   |-10546.9  |1197.24   |31.7263   |0        |143.1   |neutral |
|6  |Loch Modan, Thelsamar                     |-5351.229 |-2881.5823|340.94235 |0        |273.825 |neutral |
|7  |Wetlands, Crossroads                      |-3289.1238|-2435.9912|18.59659  |0        |144.778 |neutral |
|8  |Badlands, Graveyard NE                    |-6289.911 |-3493.108 |251.48277 |0        |160     |neutral |
|10 |The Barrens, The Crossroads               |-592.60144|-2523.4924|91.78803  |1        |70.3032 |neutral |
|31 |Desolace, Ghost Walker Post               |-1443.488 |1973.3696 |85.490715 |1        |0       |neutral |
|32 |Durotar, Razor Hill                       |233.458   |-4793.73  |10.1881   |1        |12.1862 |neutral |
|34 |Mulgore, Red Cloud Mesa                   |-2944.5564|-153.21503|65.786026 |1        |108.225 |horde   |
|35 |Darkshore, Auberdine                      |6739.19   |209.9926  |23.284561 |1        |279.001 |neutral |
|36 |Deadwind Pass, Morgan's Plot              |-11110.377|-1833.2407|71.8642   |0        |180.712 |neutral |
|39 |Thousand Needles, The Great Lift          |-4656     |-1765     |-41       |1        |272.025 |neutral |
|70 |Silithus, Valor's Rest                    |-6432.2563|-278.2921 |3.794105  |1        |12.1501 |neutral |
|89 |Mulgore, Bloodhoof Village                |-2175.1897|-342.02692|-5.512325 |1        |106.65  |alliance|
[...]

PR #1633, opened 2022-09-21
Correct graveyard faction for Dolanaar

FlagFlayer:
An intended change from https://github.com/vmangos/core/pull/1608 that accidentally wasn't included.

## 🍰 Pullrequest

As per above description.

### Proof

- None

### Issues

- Currently, the horde has no graveyard to be sent to when dying in Teldrassil. This is because the Dolanaar graveyard, intended to be a neutral graveyard, is currently alliance only.

### How2Test

- Make a horde character
- Teleport to Teldrassil
- Die

### Todo / Checklist

- [X] None


Daribon:
Everything is good. This PR is ready to be merged.

PR #1634, opened 2022-09-21
Add cmake variable to change the configuration directory

vetu104:
## 🍰 Pullrequest

Define a cache variable in cmake for setting the path for realmd.conf and mangosd.conf. If installing system wide on an unix system, e.g. via a package manager, the default location /usr/etc is nonstandard.
### Proof

- None
### Issues

- None
### How2Test

1. Build with `-DCONF_DIR` unset and cache cleared, conf files will be in `$CMAKE_INSTALL_PREFIX/etc`.
2. Build with `-DCONF_DIR=/etc/vmangos`, conf files will be in system-wide /etc.

### Todo / Checklist

- [X] None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,7 +148,7 @@ if(WIN32)
   set(LIBS_DIR ${CMAKE_INSTALL_PREFIX})
 else()
   set(BIN_DIR ${CMAKE_INSTALL_PREFIX}/bin)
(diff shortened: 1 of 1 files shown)


vetu104:
For reference, this is my arch linux PKGBUILD I use for installation and updating: http://0x0.st/oWTL.txt

PR #1635, opened 2022-09-22
Correct ghost zones to Patch 1.6  sub-areas, remove invalid graveyards

FlagFlayer:
## 🍰 Pullrequest
Currently, when you die outside of a starting area, your ghost can still be sent to the graveyard in the starting area should it be the closest graveyard to the player character on death. This pull request corrects the ghost zones of graveyards in starting areas to use the ghost zones of the starting areas, instead of using the ghost zone for the entire zone.

### Proof
"Starting area graveyards are still reserved exclusively for characters of similar faction that die **within their confines**. For example, the spirit of a Horde character dying right outside of [Deathknell](https://wowpedia.fandom.com/wiki/Deathknell) will not use Deathknell graveyard, but will instead go to the next closest Tirisfal Glades graveyard." (Patch 1.6 change notes: https://web.archive.org/web/20060104092012/http://www.worldofwarcraft.com/patchnotes/patch-05-08-02.html)

TBC Classic behaviour recorded by @heyitsbench : https://www.youtube.com/watch?v=wyaBOODuwqs&t=53s

### Issues
- Fixes https://github.com/vmangos/core/issues/1417

### How2Test
- Make a horde/alliance character
- Die in a location such that the closest graveyard is the graveyard in the starting area, but the character is not in the starting area zone.

### Todo / Checklist

- [X] None

Issue #1639, opened 2022-09-26
🐝 [Bug] Kodo Hide Bag quest [769] is missing skill requirement, and is able to be taken multiple times


FlagFlayer:
## 🐝 Bug report - Quest

Current behaviour: The quest is able to be taken at a leatherworking skill 1. The quest will be available to be retaken should you abandon leatherworking after completing it then train leatherworking again (though it should be noted that there is no way to complete the quest twice, as the game knows you already completed the quest and doesn't allow you to turn it in). 

### Expected behavior

- The quest should require a leatherworking skill of 10 (see comments at https://classic.wowhead.com/quest=769/kodo-hide-bag )
- The quest should not be available to be taken again when abandoning and retraining leatherworking

### Steps to reproduce

1. Train leatherworking when above the quest level req. of 4
2. (The quest will now incorrectly be available to be taken)
3. Complete the quest
4. Abandon leatherworking, and then immediately train it back
5. (The quest will now be available to be taken again, though when you complete the objectives you won't be able to turn it in, as there will be no question mark above the quest giver Veren Tallstrider's head, and you will not be able to talk to him)

### Version & Environment

Client Version: 
  - 1.12.1.5875

Tested on Darrowshire and Everlook Closed PTR (I am unsure of the specific hash each of them use)

FlagFlayer:
Part of a larger issue with profession quests. Will open a more relevant issue on this later.

PR #1637, opened 2022-09-25
Fix patch for graveyards.

Daribon:
## 🍰 Pullrequest

This PR changes build_min to patch_min and adds patch_max to game_graveyard_zone.
It will read patch set in mangosd.conf, so you will not be sent to a graveyard that is not for the set patch.
If you build vmangos for a patch below 1.12.1, it will still work just as before and use correct graveyards.

Includes SQL fixes for all graveyards that was changed from between 1.2 to 1.12.
### Proof

- Patch notes.

### Issues

- Fixes https://github.com/vmangos/core/issues/1590
- Fixes https://github.com/vmangos/core/issues/1492

### How2Test

1. Set patch 3 in mangosd.conf
2. .tele ravenhill
3. .die
4. You will now spawn at Darkshire graveyard.
5. Set patch 10 in mangosd.conf
6. .tele ravenhill
7. .die
8. You will spawn at Raven Hill graveyard.

### Todo / Checklist

- [X] Make Goldshire GY neutral in pre 1.6
- [X] Set patch in creature table for all spirit healers that weren't spawned in prior patch.
- [X] Find a way to use Orgrimmar Durotar GY for Hall of Legends when Razor Hill GY is set to neutral.
- [X] Fix graveyard for dying in Uldaman on old patch.

PR #1640, opened 2022-09-27
Correct Starting Area Graveyards 

FlagFlayer:
## 🍰 Pullrequest
Correct all starting area subzones to use the relevant graveyard, matching their behaviour on Classic, correcting bugs created by https://github.com/vmangos/core/pull/1635.
### Proof

"Starting area graveyards are still reserved exclusively for characters of similar faction that die within their confines. For example, the spirit of a Horde character dying right outside of Deathknell will not use Deathknell graveyard, but will instead go to the next closest Tirisfal Glades graveyard." (Patch 1.6 change notes: https://web.archive.org/web/20060104092012/http://www.worldofwarcraft.com/patchnotes/patch-05-08-02.html)

### Issues

- Multiple bug reports relating to starting area graveyards

### How2Test

- This PR has been thoroughly tested on Everlook, no issues were found.

### Todo / Checklist

- [X] None


FlagFlayer:
Added fix for https://github.com/vmangos/core/issues/1639

Probably a priority to fix the zoning of a starting zone graveyard though, so I will leave this PR alone for now.

FlagFlayer:
y'know maybe I should just not mess with this any more

NickTyrer:
Why close? Let the reviewers decide if it's needed or not.

ratkosrb:
The conditions makes absolutely no sense to me. It's checking if that same quest has been done. If it has been done, then you wont' be able to take it anyway, so wtf does the condition do.

NickTyrer:
Maybe it stops https://github.com/vmangos/core/issues/1639 from happening?

FlagFlayer:
> Maybe it stops #1639 from happening?

Yes that was the intention

FlagFlayer:
This not being pushed yet gave me time to notice that the shaman class quest areas have their own zone ID. This should definitely be the final fix for this.

FlagFlayer:
I got that wrong, literally every minimap zone in the starting area uses an ID whos parent is the zone ID, rather than the starting area. Going to add all the missing entries now.

FlagFlayer:
If https://github.com/vmangos/core/pull/1637 gets merged before this PR, do I need to change `build_min` to `patch_min`?

FlagFlayer:
@Daribon thought I'd mention you, since it's your pull request I'm asking about

Daribon:
No need, once my PR gets merged we can just edit your PR on merge.

FlagFlayer:
I have now noticed that this fix also prevents graveyard hopping from Burning Blade Coven to Razor Hill, although that can easily be fixed by adding:
``` 
-- Add Razor Hill as a graveyard for Burning Blade Coven to retain the ability to graveyard hop from Valley of Trials to Razor Hill
INSERT INTO `game_graveyard_zone` (`id`, `ghost_zone`, `faction`, `build_min`) VALUES (32, 365, 0, 0);
```

Should this be changed? Technically speaking the ability to graveyard hop from Valley of Trials to Razor Hill is a bug, but one that has existed for so long that I don't think it should be changed. I'm uncertain of the vanilla or classic behaviour regarding this graveyard hop, too, which may be needed in order to reach a decision on this.

Gamemechanicwow:
I think the Kodo Hide Bag quest should be fixed in the code and not in the database.

Don't know what causes the bug, but I guess it could also happen for other profession quests. And having one fix seems like a prettier solution.

FlagFlayer:
I suppose so, though I could just make a PR to do this same process for every profession quest, but that would be quite an ugly solution. I'll remove the parts of this PR that relate to the Kodo Hide Bag, since the topic is not in any way related to the graveyards anyways.

FlagFlayer:
The matter of that tiny spot in Burning Blade Coven sending you to Razor Hill GY still needs to be checked. I have no way to access Classic as of the moment to check the behaviour, though my own sense of game design certainly tells me to add that code to allow the graveyard hop as it has been thoroughly embraced by the community in leveling guides used in private servers.

Daribon:
> The matter of that tiny spot in Burning Blade Coven sending you to Razor Hill GY still needs to be checked. I have no way to access Classic as of the moment to check the behaviour, though my own sense of game design certainly tells me to add that code to allow the graveyard hop as it has been thoroughly embraced by the community in leveling guides used in private servers.

I can confirm that the spot is blizzlike at least when compared to classic. I tested it on classic and got ported to Razor Hill.
WowClassicT_VV3KYCqPzp
WowClassicT_G6S53GnCK2

If you want to replicate bug for vmangos I'm not sure how you would go by doing that since the vmangos graveyard system is based around ZoneID and AreaID.

FlagFlayer:
Replicating the bug should be as easy as adding Razor Hill as a possible graveyard for Burning Blade Coven, no? If there are multiple graveyards that use the same AreaID, is there no check to see which is closer?

EDIT: Also, thank you for testing this, I appreciate it a lot!

Daribon:
Ah right thats true, that will probably work.

FlagFlayer:
I'm quite certain that the only reason that the Razor Hill graveyard hop exists is private servers and more notably Classic incorrectly using starting zone graveyards as zone-wide, which is not vanilla-like (as is described in game patch notes: "Starting area graveyards are still reserved exclusively for characters of similar faction that die within their confines. For example, the spirit of a Horde character dying right outside of Deathknell will not use Deathknell graveyard, but will instead go to the next closest Tirisfal Glades graveyard."). Considering the point of the project is to recreate things as they were in vanilla, it's pretty antithetical to implement this bug. @ratkosrb what are your thoughts on this?

AnonXS:
As I've recently corrected some more cmangos graveyard data, i can say that there is alot of wrongly inserted data in most databases, because the schema of the table was so badly disigned core wise that db devs inserted wrong data over time, so reviewing alot (most of this should be beneficial, speaking to you scholmance port to developmend land that is never hit)
```
DELETE FROM `game_graveyard_zone` WHERE `id` = 429 AND `ghost_loc` = 2057; -- Scholomance linked to Tirisfal Glades, Faol’s Rest!? Also has id 629 "TEST for GM Client Only – Do Not Bug" for 2057, leave that for now)

```

Daribon:
> As I've recently corrected some more cmangos graveyard data, i can say that there is alot of wrongly inserted data in most databases, because the schema of the table was so badly disigned core wise that db devs inserted wrong data over time, so reviewing alot (most of this should be beneficial, speaking to you scholmance port to developmend land that is never hit)
> 
> ```
> DELETE FROM `game_graveyard_zone` WHERE `id` = 429 AND `ghost_loc` = 2057; -- Scholomance linked to Tirisfal Glades, Faol’s Rest!? Also has id 629 "TEST for GM Client Only – Do Not Bug" for 2057, leave that for now)
> ```

Yes, I noticed the funny scholomance graveyards a while ago and fixed them in this PR:
https://github.com/vmangos/core/pull/1637/files

AnonXS:
yea just saying, it would be beneficial to all of us to just display all existing data in cleartext and remove/add the broken stuff once and for all instead of disussing case by case by case and not getting along well, because the gist is
"that existing data has to be correct as is", when its not.
But the fact is that alot of wrong data was added along the way.
i believe there is alot more wrong with it and it should be professionally revamped "instead of working through a cable tree like a it tech worker that does not know what he's doing".
because especially with graveyards, you fix one and you might break two others related to it without even knowing that the "broken part you removed was holding it all together somehow due to relations, priority"

FlagFlayer:
I'm not at all certain what you're trying to get at. I'm quite confident that existing data will be very, if not completely accurate once these PRs by my and Daribon are merged. Do you have any proof that there's something wrong with the current data that isn't being addressed? The only major bug in the data really was starting zone graveyards being messed up in this PR: https://github.com/vmangos/core/pull/1635 , which was due to my lack of awareness of the very simple mechanics by which graveyards work in vmangos.

AnonXS:
It's just me, i stumpled on a commit changing some of this tables data and for me it opened a can of worms where i thought it might be worth "dewiring" the whole thing because otherwise i'd just get a headach as all is connected.
i believe you'll be fine merging all these commits, but for me trying to get this backported to cmangos its best to have a "complete closed all in one pull request instead of merging single area/zone fixes".

FlagFlayer:
At this point I believe this situation has been completely "dewired" as far as I or Daribon are concerned. Importing graveyard changes to cmangos is a simple searching of previous PRs and working chronologically, which really isn't so bad when there's only a few very short PRs relating to graveyards, but I digress. This is not the place for this conversation. If you don't have something to contribute to this PR, don't comment here.

ratkosrb:
> I'm quite certain that the only reason that the Razor Hill graveyard hop exists is private servers and more notably Classic incorrectly using starting zone graveyards as zone-wide, which is not vanilla-like (as is described in game patch notes: "Starting area graveyards are still reserved exclusively for characters of similar faction that die within their confines. For example, the spirit of a Horde character dying right outside of Deathknell will not use Deathknell graveyard, but will instead go to the next closest Tirisfal Glades graveyard."). Considering the point of the project is to recreate things as they were in vanilla, it's pretty antithetical to implement this bug. @ratkosrb what are your thoughts on this?

To me it makes no sense they would change something like this specially for Classic, especially given that it's not consistent with other zones as you say. Whats the motive? This is likely a bug in the original data that has always existed. The patch notes are sadly not very reliable. Many times you see the same fix being mentioned in multiple patches, as it either didn't actually go live in the first patch in whose changelog they mentioned it, or the first fix didn't work, or they re-broke it in a subsequent patch. There are cases like "invisibility being removed on start of cast" being mentioned in patch notes, yet there exist pvp videos from after the patch in which its not removed until cast finishes. So given how it works in Classic, it makes the most sense they just messed up Razor Hill back in the day. The most trustworthy way to find out how it worked is finding a video from 2004-2006 in which the player dies in that spot. If that's not possible, then we could also perform some more tests and check how it works in Classic WotLK. If its different than how it works in Classic Era, then it's clearly an original bug in the Vanilla data that Classic was based on, that was fixed in one of the expansions. If it works the same in WotLK too, then any old video up to the end of original WotLK in 2010 could serve as proof that this was actually an intentional change for the Classic re-release, or a difference in functionality in the new server code compared to the original regarding graveyard selection.

FlagFlayer:
>This is likely a bug in the original data that has always existed. The patch notes are sadly not very reliable. Many times you see the same fix being mentioned in multiple patches, as it either didn't actually go live in the first patch in whose changelog they mentioned it, or the first fix didn't work, or they re-broke it in a subsequent patch.

Narrowing this down, the graveyard hop was probably introduced in patch 1.6: "Feralas, **Durotar**, Mulgore, Darnassus, Tirisfal Glades (excluding the Bulwark graveyard), Elwynn Forest & the Eastern Plaguelands have now all been converted over to the new graveyard system. The system will place your character's spirit at the closest allowed graveyard in the zone." The new proximity-based graveyard system, according to the patch notes, was introduced in patch 1.3 for Ashenvale, Felwood, and Azshara, but considering patch 1.6 reworked the graveyards for Durotar specifically, it's likely something like this would be introduced there.

Starting area graveyards were supposed to be excluded by this change, as I stated in the quote before, but, in fact, I would not be surprised if moving the Valley of Trials graveyard to the new proximity-based graveyard system was done intentionally rather than erroneously. 

The Valley of Trials is unique among starting zones as the only starting zone with multiple exits: One exit going eastward towards Sen'jin Village, and another one-way exit going out via Hidden Path to the southern reaches of the Southfury River. Considering this layout, it is likely that changing the Valley of Trials to the proximity-based system will almost always shorten your distance to get to your corpse, thereby making that change have game-sense logic to it. An unintended consequence of doing such a thing would be the tiny spot in Burning Blade Coven that accidentally sends you to Razor Hill.

If this behaviour is identical in WotLK Classic, I would not be at all surprised if the origin of this behaviour, and the graveyard hop, was a change, accidental or not, in patch 1.6, and remained in the game ever since. If this is the case, I can easily change the Valley of Trials graveyard to use the proximity-based ZoneID system to correctly emulate blizz-like behaviour (It will need to be rewritten with the patch minimum and maximums introduced by Daribon to include pre-patch 1.6 behaviour in the database).

FlagFlayer:
Ready for merge with the exception that this must only be merged after https://github.com/vmangos/core/pull/1637 has already been merged.

Daribon:
Fixed one incorrect areaid in the PR and made the comments easier to read.

It is ready for merge.

PR #1661, opened 2022-10-22
Change AV BG announcer to correct NPC.

### LINKED ISSUE #1370: 🐛 [Bug] Wrong NPC yelling in Alterac Valley
## 🐛 Bug report

Currently Stormpike Herald is yelling in Alterac Valley.
### Expected behavior

Stormpike Herald should not yell at all. It should instead be the NPC Herald https://classic.wowhead.com/npc=14848/herald that yells.
Currently Stormpike Herald has a second custom spawn outside AV which does the yelling, it should be removed.

Herald has tons and tons of spawns in the sniff DB for Alterac Valley, check sniff for entry 14848 DB to see what I mean.
From what I can gather from the sniff DB, it appears that the horde Herald should be on top of Iceblood Garrison and the alliance one should be on top of Stonehearth Outpost.

The only time Stormpike Herald ever yelled, was on the AV PTR, before it was even released:
image
This is the only image I've seen of Stormpike Herald yelling.

Images showing Herald yelling instead:
Pre 1.8:
alteractal41
2006/07/25
WoWScrnShot_072506_203122

2005/07/22:
WoWScrnShot_072205_231946

I have not found a single image, out of 1588 images I have from release Alterac Valley, that shows Stormpike Herald yelling. Neither have I found any video showing it. They all show Herald yelling.
### Steps to reproduce

1. Join Alterac Valley
2. Cap anything
3. Watch the incorrect NPC yell.
4. Profit

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  https://github.com/vmangos/core/commit/1c89a85fe3f91d160c9a195e3a2f1fd62fc418ea

OS Client: Win 11
OS Server: Win 11

### Crashlog

- None

Daribon:
## 🍰 Pullrequest

Changes AV BG announcer from Stormpike Herald to Herald.
Stormpike Herald ever only yelled the BG announcements in 1.5.0 PTR. From release 1.5.0 and forward it has been Herald that does all the announcements.

Screenshot taken after PR applied:
WoW_856N4Reapj

### Proof

- Check the issue this PR fixes.

### Issues

- Fixes https://github.com/vmangos/core/issues/1370

### How2Test

- .level 59
- .debug bg
- .go alterac
- Cap any graveyard, bunker or tower.

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20221022011155_world.sql
+++ b/sql/migrations/20221022011155_world.sql
@@ -1,0 +1,22 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20221022011155');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20221022011155');
+-- Add your query below.
+
+
+-- Change AV BG announcer from Stormpike Herald to Herald
+UPDATE `creature` SET `id`='14848' WHERE  `guid`=150155;
+DELETE FROM `creature_battleground` WHERE  `guid`=150154 AND `event1`=60;
+
+
+-- End of migration.
+END IF;
+END??
+delimiter ; 
+CALL add_migration();
+DROP PROCEDURE IF EXISTS add_migration;


AnonXS:
Herald is a trigger npc ontop of the keep for alliance. changing ids is not a good fix

Daribon:
@AnonXS 

Herald is an NPC with several spawns in Alterac Valley. Both Horde and Alliance has separate spawns using the same entry.
Alterac Valley used to have separate Heralds for Horde and Alliance on the PTR, before Alterac Valley was released. When the 1.5.0 finally landed, just one NPC was used for all yells.

They most likely just replaced the Alliance and Horde Herald entries with the neutral Herald on release.
It could explain why there are two spawns of Herald on Alliance side and two spawns of Herald on Horde side.

1.5.0 AV PTR as Horde:

1.5.0 AV PTR as Alliance, in korean but you can see the [Orcish] brackets on the last yell:

서리늑대전량의외침: translates to Frostwolf Massive Shout

---------------------

Here are the following Herald spawn locations, the gnome marks the spawn spot:
WoWScrnShot_121422_211808
WoWScrnShot_121422_211901
WoWScrnShot_121422_212136
WoW_tr7vrRZgqI

Prior to the PR, the Alliance Herald did both the yelling for Horde and Alliance. This was incorrect, since it shouldn't even be yelling at all.

I did not mirror blizzards behavior with 4 spawns as it is illogical. I do not find it reasonable to emulate the Blizzard way since the end result is the same.

In comparison, Blizzard uses trigger NPC rats in Molten Core for dousing which is not something VMaNGOS does, or any emulator. It is not necessary since same thing can be done without the trigger NPCs.

PR #1684, opened 2022-11-01
Add custom invisible walls to WSG from Classic.

Daribon:
## 🍰 Pullrequest

This adds invisible walls at terrain exploit spots in WSG. From Classic but the invisible walls never existed in vanilla
The object exists in 1.7.0 and forward, if you are using 1.12.1 you can change patch_min.

As this is a patch that adds non vanilla stuff, it is therefor optional and placed in the custom SQL folder.

Some examples:
WoW_8YijLKGduO
nomacs_rgispkK6hy
nomacs_aVz5hsWDAf

### Proof

- None

### Issues

- None

### How2Test

- Go to Alliance Graveyard, attempt to jump in the corner between the tree and alliance base building.
- It will be completely impossible due to invisible wall.

### Todo / Checklist

- [X] None


ratkosrb:
I suggest you add column names to the inserts, or sql will break if table format is updated to add a new column.

Daribon:
Right, forgot about that.
Will fix it.

Daribon:
Corrected SQL with column names.

PR #1687, opened 2022-11-05
Remove pvp quests which should not be available in classic

schell244:
## 🍰 Pullrequest
Noticed that those quests should not be available in classic, since with German locales Shattrath is mentioned in quest description.
https://www.wowhead.com/tbc/quest=8388/for-great-honor
https://www.wowhead.com/tbc/quest=8367/for-great-honor
https://www.wowhead.com/tbc/quest=8371/concerted-efforts
https://www.wowhead.com/tbc/quest=8385/concerted-efforts

[Update]
Those quests do exist in wowhead classic, so they might belong to classic !?
https://www.wowhead.com/classic/quest=8388/for-great-honor
https://www.wowhead.com/classic/quest=8367/for-great-honor
https://www.wowhead.com/classic/quest=8371/concerted-efforts
https://www.wowhead.com/classic/quest=8385/concerted-efforts 

### Proof
Quest tag in client seems to be off for some of those quests:
quest
questa

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [ ] needs prove

### DIFF
--- a/sql/migrations/20221105134018_characters.sql
+++ b/sql/migrations/20221105134018_characters.sql
@@ -1,0 +1,18 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20221105134018');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20221105134018');
+-- Add your query below.
+
+DELETE FROM `character_queststatus` WHERE `quest` IN(8388, 8367, 8371, 8385);
+
+-- End of migration.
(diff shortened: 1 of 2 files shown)


Gamemechanicwow:
I think the quests belong in vanilla, but the quest npc's was moved to Shattrath.

NPC is in Thunderbluff
https://www.wowhead.com/classic/quest=8388/for-great-honor

NPC is in Shattrath
https://www.wowhead.com/tbc/quest=8388/for-great-honor

Notice that there are also comments from 1.8 on wowhead. I suspect that the german locales originate from TBC.

schell244:
Thanks @Gamemechanicwow. I just noticed that those quests can be found at wowhead classic as well and added the related links. German locales are probably just off as you said. Weird that the quest tags are 'wrong' for both alliance quests and one horde quest. Maybe they also just forgot to update those in the client...

PR #1695, opened 2022-11-11
Add warden checks for autoloot hacks

Gamemechanicwow:
## 🍰 Pullrequest

Adds checks for autoloot hack for 1.12.1, 1.12.2 and 1.12.3

### Proof

- None

### Issues

- None

### How2Test

- Check if warden triggers on an unmodified client
- Check if warden triggers on a client with changed values at the memory address

### Todo / Checklist

- [X] None


ratkosrb:
I don't think this is really a hack, as you can achieve the same effect by holding shift.

PR #1696, opened 2022-11-12
Fixes for Warden on Mac.

ratkosrb:
- Scans for Mac weren't even being loaded, so the string hash request was never sent.
- The game crashed if you connected with the PPC client, because it was being sent a x86 module.
- The structure of WARDEN_SMSG_CHEAT_CHECKS_REQUEST was wrong for Mac.
- Added back the Warden.DebugLog option so I can actually see what's happening in the console.

Thanks to @Daribon and @GrenderG for helping to test.


GrenderG:
Don't forget to include the header to fix the Linux build while you are at it!

PR #1733, opened 2022-12-10
illumination follow up

Gamemechanicwow:
## 🍰 Pullrequest

Holy shock has a dummy effect, that casts different spells based on the target. Those spells don't have a mana cost, so illumination will not refund any.

This pullrequest makes sure we get the correct spellid.

Credit to Cmangos

### Proof

- None

### Issues

- [Holy Shock Illumination always refunds 0 mana](https://github.com/Wall-core/Everlook-Bugtracker/issues/769#top)

### How2Test

- .m spellcrit 100
- use holy shock

### Todo / Checklist

- [X] None


ratkosrb:
Changing spell data is a hack fix. It's better you handle this in the code. Pass the holy shock spell id as triggeredBy and use that in the proc handler when its set.

ratkosrb:
Or actually i think u don't have access to that in there, you only get the SpellEntry. But you do get if the spell is triggered, so check the client casted spell in that case i guess.

ratkosrb:
check the cmangos solution
https://github.com/cmangos/mangos-classic/blob/master/src/game/Spells/UnitAuraProcHandler.cpp#L1422

Gamemechanicwow:
> check the cmangos solution https://github.com/cmangos/mangos-classic/blob/master/src/game/Spells/UnitAuraProcHandler.cpp#L1422

Yup, that is definitely a better approach.

PR #1741, opened 2022-12-20
Add some missing german locales

schell244:
## 🍰 Pullrequest
- add missing german locales `RequestItemsText_loc3` and `OfferRewardText_loc3` for some repeatable quests. Texts where taken form wowhead (https://www.wowhead.com/classic/de)
- fixed wrong required item count in some quest texts
- added some missing gameobject translations


0blu:
Use `SET NAMES 'utf8';` and ensure the file is also UTF8, otherwise the migration can fail on some db versions.
After that we will be able to merge it.

PR #1752, opened 2023-01-04
A Game of Towers fixes

Oroxzy:
Fixes and Additions

- [x] Curing Shrine respawn fixed
- [x] Adding missing GameObjects
- [x] Adding missing sounds
- [x] Adding missing worldstates
- [x] Rework capture mechanics
- [x] Adding spell scripts
- [x] test

Actually the code should be rewritten and implemented here later:
https://github.com/vmangos/core/blob/9749c97b5a2babebac5b7c1ec2b64eb042aa5311/src/game/Objects/GameObject.cpp#L1917


ratkosrb:
When i tested this on classic, i noticed it despawns and immediatelly respawns. So isn't the fix as easy as setting respawn time to 1 in db ?

Oroxzy:
its spawn is hardcoded, maybe i find another way... 
i was just on classic to sniff the towers again and found other missing things

src/game/OutdoorPvP/OutdoorPvPEP.cpp
```diff
@@ -50,88 +50,142 @@ void OPvPCapturePointEP_EWT::ChangeState()
 
     uint32 artkit = 21;
     uint32 animation = 2;
+    uint32 sound = NULL;
```
ratkosrb: please just use 0 instead of NULL

Issue #539, opened 2020-05-11
Deduction calculation from Rep and Honor bug


NickTyrer:
**Current behavior**: 
The deductions calculation is slightly out see
![Untitled Diagram](https://user-images.githubusercontent.com/22723839/81585358-f35efa00-93ab-11ea-8c00-e081fb6e415b.png)
green is classic

**Client version**: 
1.12.1
**Commit hash**: 

**Operating system**: 
Win 10

ghost:
Can this be closed now @NickTyrer ?

PR #1756, opened 2023-01-07
Changes to deduction calculation from Rep and Honor

Gamemechanicwow:
## 🍰 Pullrequest

Changes the price after deduction to be rounded to the nearest integer. 
Changes price deduction for using a taxi to 5% at rank2 and an additional 5% at rank4.

### Proof

Screenshots to compare prices:
image
Math tells us that prices are rounded to the nearest int.

Decompiled TaxiNodeCost from 1.12.1 client:
image

This is also seen in classic
image
https://youtu.be/vnWPqK1VBXs?t=1118
Stormwind -> Ironforge : 50c
Ironforge -> Menethil Harbor : 330c
Menethil Harbor to Southshore: 330c
Total: 710c * 0,85 = 603,5c

10% reputation +5% discount for Rank2 (he is rank3 according to footage)

### Issues

- (https://github.com/vmangos/core/issues/539)

### How2Test

- Set ranking points/reputation value 
- Buy stuff / Train stuff / Fly somewhere
- Compare prices to known values and use a calculator - or compare taxiflight prices with the value shown on screen, as this is calculated by the client.

### Notes

- There is some minor precision loss with float calculations. Maybe we should add flt_epsilon to mod. Blizzard seems to do it :>
-

### DIFF
--- a/src/game/Handlers/ItemHandler.cpp
+++ b/src/game/Handlers/ItemHandler.cpp
@@ -841,7 +841,7 @@ void WorldSession::SendListInventory(ObjectGuid vendorguid, uint8 menu_type)
                 ++count;
 
                 // reputation discount
-                uint32 price = uint32(floor(pProto->BuyPrice * discountMod));
+                uint32 price = uint32(pProto->BuyPrice * discountMod + 0.5f);
 
                 data << uint32(count);
                 data << uint32(crItem->item);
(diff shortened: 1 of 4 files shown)


NickTyrer:
I did some comparisons here https://github.com/vmangos/core/issues/539, maybe related.

Issue #1759, opened 2023-01-09
🐛 [Bug]Crash


yangyitzzg:
## 🐛 Bug report
Code Version:[20230107](https://github.com/vmangos/core/commit/6c2cda47e1ccaadfe1eb3ef6269fcde039a21e0a)

### Expected behavior
The general way is that when some map creatures are attracted to some specific locations, mangosd will be crash.

### Version & Environment
Client Version:1.12.3.6141
Server: Windows Server 2012

### Crashlog:
[Crashs.zip](https://github.com/vmangos/core/files/10370287/Crashs.zip)

ratkosrb:
It is the same crash you posted in https://github.com/vmangos/core/issues/1755

yangyitzzg:
Yes, but I have updated the following files to the latest:

contrib/mmap/src/MapBuilder.cpp
contrib/vmap_extractor/vmapextract/vmapexport.cpp
contrib/vmap_extractor/vmapextract/wmo.cpp
dep/recastnavigation/Detour/Source/DetourNavMeshQuery.cpp
src/game/Anticheat/MovementAnticheat/MovementAnticheat.cpp
src/game/Anticheat/MovementAnticheat/MovementAnticheat.h
src/game/Handlers/MovementHandler.cpp
src/game/Maps/Map.cpp
src/game/Maps/Map.h
src/game/Maps/PathFinder.cpp
src/game/Maps/PathFinder.h
src/game/Maps/ZoneScript.cpp
src/game/Maps/ZoneScript.h
src/game/Movement/MotionMaster.cpp
src/game/Movement/MovementPacketSender.cpp
src/game/Movement/TargetedMovementGenerator.cpp
src/game/Movement/WaypointMovementGenerator.cpp

coolzoom:
you are on your own if the source is different than the official branch.
people are not able to repeat the issue if the sources are different.

yangyitzzg:
> you are on your own if the source is different than the official branch. people are not able to repeat the issue if the sources are different.

The code comes from the official branch and is currently in the learning state. Some minor modifications have been made, such as the code for generating dmp files. Other codes are consistent with the official ones. No changes have been made to the map code.

Gamemechanicwow:
The big issue here is that there is not enough information to reproduce it - or anyone else reporting similar crashes.

If it still happens with a clean vmangos build - I would advise you to extract and rebuild the maps.

yangyitzzg:
[...]

PR #1762, opened 2023-01-12
Fix Crash

yangyitzzg:
Fix Crash caused by null pointer

[Fix:](https://github.com/vmangos/core/issues/1759#issue-1524886340)

### DIFF
--- a/src/game/Objects/Object.cpp
+++ b/src/game/Objects/Object.cpp
@@ -1872,26 +1872,29 @@ bool WorldObject::GetRandomPoint(float x, float y, float z, float distance, floa
         rand_x = x;
         rand_y = y;
         rand_z = z;
-        if (map->GetWalkRandomPosition(GetTransport(), rand_x, rand_y, rand_z, distance, moveAllowed))
+        if (GenericTransport* transport = GetTransport())
         {
-            // Giant type creatures walk underwater
-            if ((isType(TYPEMASK_UNIT) && !ToUnit()->CanSwim()) ||
-                (IsCreature() && ToCreature()->GetCreatureInfo()->type == CREATURE_TYPE_GIANT))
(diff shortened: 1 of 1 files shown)


Gamemechanicwow:
GetTransport() returning null pointer is fine and won't cause a crash. It's resolved here:

```
bool Map::GetWalkRandomPosition(GenericTransport* transport, float &x, float &y, float &z, float maxRadius, uint32 moveAllowedFlags) const
{
   ...
    dtNavMeshQuery const* m_navMeshQuery = transport ? mmap->GetModelNavMeshQuery(transport->GetDisplayId()) : mmap->GetNavMeshQuery(GetId());
  ...
```

yangyitzzg:
dtStatus dtNavMeshQuery::findRandomPointAroundCircle(dtPolyRef startRef, const float* centerPos, const float maxRadius,
													 const dtQueryFilter* filter, float (*frand)(),
													 dtPolyRef* randomRef, float* randomPt) const
.....
	const dtPoly* startPoly = 0;
	m_nav->getTileAndPolyByRefUnsafe(startRef, &startTile, &startPoly);
	if (!filter->passFilter(startRef, startTile, startPoly))
		return DT_FAILURE | DT_INVALID_PARAM;

.......

startPoly is nullptr,You are right. I saw it. I also saw that the Sunwell project did not modify DetourNavMeshQuery.cpp, so I did not modify it. At present, there is no way to deal with this problem. After DEBUG the whole process, I found that there seemed to be no other place except GetTransport().

yangyitzzg:
After my testing for a period of time, this code can really solve the problem of crash caused by random pathfinding, but it also produces new problems. All creatures without path point configuration will no longer move, so please close this submission for me. This is a bug fix. thank you.

PR #1768, opened 2023-01-14
Sunken Temple: Rework Atal'ai Statue Event

schell244:
## 🍰 Pullrequest
- Prevent quest turn in of [Secret of the Circle](https://www.wowhead.com/classic/quest=3447/secret-of-the-circle) while boss Atal'alarion is alive.
- fix object state of Idol of Hakkar after server crash/restart
- fix state of Atal'ai Statue objects after crash/restart
- fix Atal'alarion visibility when alive after server crash/restart
- fix Atal'ai Statue traps getting triggered when statue in correct order was used
- some code cleanup

### Proof
https://www.wowhead.com/classic/quest=3447/secret-of-the-circle#comments
_Killing the guardian will allow you to click on the statue in the center of the pool to complete the quest._

### Issues
- None

### How2Test
- quest add 3447
- check circle traps
- activate circle (S, N, SW, SE, NW, NE)
- while Atal'alarion is alive, it should not be possible to interact with the Idol for quest turn in
- stop server and check dungeon state after restart

### Todo / Checklist
- [X] tested


ratkosrb:
Won't this break on server restart mid run? Object will have the flag applied on spawn, but never removed, because the npc will spawn already dead.

schell244:
Thanks for checking. I think the whole script needs some polishing.
Currently after restart both boss (-corpse) and idol object are no longer visibile - means boss loot is lost also.
So leaving the flag with current state does not break anything.

Some more todo's for the script:
The circle stones always deal damage when activated, which as far as I know incorrect and should only happen when activating order was wrong. 
Atal'alarion might be missing one abillity [Sweeping Slam](https://www.wowhead.com/tbc/spell=12887/sweeping-slam), which is only listed in tbc wowhead, but was mentioned in some classic dungeon guide.

schell244:
Idol of Hakkar state and visibility after crash/restart is fixed now, including the state of circle stone objects.

schell244:
Beside the `GameObjectAI->OnUse()` issue (where the related code is outside of the dungeon script), all other points are fixed now.

src/scripts/eastern_kingdoms/swamp_of_sorrows/sunken_temple/instance_sunken_temple.cpp
```diff
@@ -618,6 +610,30 @@ InstanceData* GetInstance_instance_sunken_temple(Map* pMap)
...
+
+struct go_atalai_lightAI : public GameObjectAI
+{
+    go_atalai_lightAI(GameObject* pGo) : GameObjectAI(pGo) {}
+
+    bool OnUse(Unit* pUser) override
+    {
+        if (ScriptedInstance* pInstance = (ScriptedInstance*)me->GetInstanceData())
+        {
+            pInstance->SetData64(me->GetEntry(), me->GetGUID());
```
ratkosrb: Isn't it possible to do this in db using `gameobject_scripts` table? You can assign script per guid of the spawn there.

schell244: I have tried it with `Command: Set Data 64` and both Modes `Safe Own Guid` and `Save Raw Value` but it did not work.
This was my attempt with  `Save Raw Value`:
```
UPDATE `gameobject_template` SET `script_name`='' WHERE `entry` IN (148830, 148831, 148832, 148833, 148834, 148835);

DELETE FROM `gameobject_scripts` WHERE `id` IN (27898,27899,28111,28112,28113,28114);
INSERT INTO `gameobject_scripts` (`id`, `delay`, `priority`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) values
(27898, 0, 0, 38, 148830, 27898, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle'),
(27899, 0, 0, 38, 148831, 27899, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle'),
(28111, 0, 0, 38, 148832, 28111, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle'),
(28112, 0, 0, 38, 148833, 28112, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle'),
(28113, 0, 0, 38, 148834, 28113, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle'),
(28114, 0, 0, 38, 148835, 28114, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Atal\'ai Statue - process secret circle');
```
Could this problem be due to the db script not using the override method which handels the script and instead just using the base instance implementation? I have attached the debugger to the override method and nothing happend when I activated a stone in game.

schell244: Converted to db. This did also fix the issue with the trap.

PR #1774, opened 2023-01-23
Dreadsteed Candle Fix

ghost:
## 🍰 Pullrequest
Followup to previous decline pull request. Thanks Zeed for the help!


ratkosrb:
The changes to the script are not necessary, as the damaging aura no longer afflicts players after the faction change of the trap gameobject. I did the quest twice, on a clean core and database, to confirm. It only affects the demons.

Please see this video:
https://streamja.com/ladQB

PR #1778, opened 2023-01-24
Feign death exploit fix

Gamemechanicwow:
## 🍰 Pullrequest

Fixes old exploit where a hunter can cast feign death and /camp. This enables the hunter to play normally while the logout timer is ticking.

### Proof

I have no evidence of how this should be handled. I guess one of 3 options:

- Player should not be able to log out while feigning death. (Which this pullrequest fixes)
- Player should be able to logout while in feign death, but not be able to cancel FD during that duration.  
- This is already working as intended :>

### Issues

- (https://github.com/LightsHope/issues/issues/163)

### How2Test

- Feign death
- /camp or logout from the main menu
- With this PR an error message should appear

### DIFF
--- a/src/game/Handlers/MiscHandler.cpp
+++ b/src/game/Handlers/MiscHandler.cpp
@@ -374,7 +374,7 @@ void WorldSession::HandleLogoutCancelOpcode(WorldPacket& /*recv_data*/)
     SendPacket(&data);
 
     // not remove flags if can't free move - its not set in Logout request code.
-    if (GetPlayer()->CanFreeMove() || GetPlayer()->IsFeigningDeath())
+    if (GetPlayer()->CanFreeMove())
     {
         //!we can move again
         GetPlayer()->SetRooted(false);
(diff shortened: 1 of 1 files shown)


Daribon:
On wotlk classic you can logout while in feign death, but you will be locked in place.
WowClassicT_U7ByAxh2zo
If you remove it by click off the buff, your character will just stand up and remain locked in place until logout.
nomacs_2PL7g03D2u

I haven't tested this in classic so I don't know if the behavior is the same.

GrenderG:
So it's currently working as intended? Interesting. In 0.5.3 the client behaves in the same way.

Gamemechanicwow:
> On wotlk classic you can logout while in feign death, but you will be locked in place. 

Interesting. After giving it some thought. I do think this feels more correct than preventing the player from logging out.

> So it's currently working as intended? Interesting. In 0.5.3 the client behaves in the same way.

Yup, except players can go pewpew 😅 

image

I will follow up on this pr later.

Gamemechanicwow:
Changed it to classic wotlk behavior.

Feign Death -> logout = unable to move while logging out - movement restored when logged in and still in FD
Feign Death -> logout -> cancel debuff = unable to move while logging out

Tested it will not interfere with other stun auras.

PR #1789, opened 2023-01-29
Typo fix in Gold logs

ghost:
1g = 10000c


ratkosrb:
Changed it to use the enum defines for Gold, Silver, Copper.

PR #1815, opened 2023-02-24
Add SQLite databases to DB dump action

### LINKED ISSUE #1812: 🚀 [Feature] Add SQLite database to DB dump
## 🚀 Feature
Add an SQLite database in addition to the raw sql files in order to make access to database contents easier for more people, and provide better grounds for long term archival.

### Benefits, Purpose

As I enumerated in the [cmangos issue](https://github.com/cmangos/issues/issues/3196):

* Loading an SQLite database is significantly easier than standing up a MySQL database and loading from that which means that the database contents are available to more developers, and significantly less effort.
* The SQLite library has bindings in most languages and is easily embedded into statically compiled files. This means that even the lowest spec computing unit can probably load an SQLite DB.
* SQLite files are [recommended by the Library of Congress](https://www.sqlite.org/locrsf.html) for archival purposes, and [SQLite itself is supported through to 2050](https://www.sqlite.org/lts.html) and quite likely even further. This means that even if MySQL bitrots and is unable to load the SQL files, the SQLite database is very highly likely to still be accessible and be supported.

### Implementation

I'm thinking something similar to [what I did for cmangos](https://github.com/cmangos/wotlk-db/pull/523/files):

1. Download `awk` and `sqlite`.
2. Clone [mysql2sqlite](https://github.com/dumblob/mysql2sqlite)
3. Run `mysql2sqlite` with `awk` and pipe into `sqlite`
4. Zip up and release.

Total size for all databases would be:

* 38 MB for `.zip`
* 23 MB for `.7z`

`.zip` is currently used so I'm assuming that's what would be preferred.

Let me know if this isn't wanted, otherwise I will come up with a PR soon-ish.

gtker:
## 🍰 Pullrequest
Adds SQLite databases to DB releases in addition to the raw SQL files, and fixes #1812.

### Proof
https://github.com/gtker/vmangos-core/actions/runs/4264014193/jobs/7421506198 shows a successful run on my fork which uploads to https://github.com/gtker/vmangos-core/releases/tag/db_latest. https://github.com/gtker/vmangos-core/releases/download/db_latest/db-sqlite-fc33961.zip contains the zipped sqlite files.
The current branch doesn't pass on my own fork because the job requires the permissions from https://github.com/gtker/vmangos-core/commit/ecaf9a1981a57193eba8330dbdfce2c5e7272fa5.

### DIFF
--- a/.github/workflows/db_dump.yml
+++ b/.github/workflows/db_dump.yml
@@ -76,6 +76,21 @@ jobs:
         # cp ${{github.workspace}}/sql/migrations/logs_db_updates.sql ${{github.workspace}}/dbexport/db_dump/update_check_only_do_not_import/logs_db_updates.sql
         # cp ${{github.workspace}}/sql/migrations/characters_db_updates.sql ${{github.workspace}}/dbexport/db_dump/update_check_only_do_not_import/characters_db_updates.sql
 
+    - name: Install SQLite dependencies
+      run: |
+        sudo apt install sqlite3 mawk -y
+        git clone https://github.com/dumblob/mysql2sqlite
+        chmod +x ${{github.workspace}}/mysql2sqlite/mysql2sqlite
+
+    - name: Create SQLite databases
+      run: |
+        mkdir ${{github.workspace}}/dbexport/sqlite_dump || true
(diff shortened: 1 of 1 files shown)


gtker:
Should I do anything else to get this merged?

PR #1843, opened 2023-03-22
Crashfix - Memory locks

TeamEverlook:
Changes:
- No longer NOTIFYs upon login, instead just adds message to chat
- Chat message on login now shows how many warnings a player has and its last warning


src/game/Language.h
```diff
@@ -299,6 +299,7 @@ enum MangosStrings
 
     LANG_COMMAND_MODIFY_REP             = 305,
     //LANG_COMMAND_MODIFY_ARENA           = 306,
+    LANG_ACCOUNT_WARNED_LOGIN = 306,
```
gibson: It would be nice if you would format the code accordingly to the surrounding code

TeamEverlook: It is on my end. Feel free to edit if you'd lilke.
image

PR #1850, opened 2023-03-24
Add missing ability to Pyrewood Elder

### LINKED ISSUE #1707: 🐛 [Bug]Pyrewood-Ältester Can't use spells
## 🐛 Bug report

Pyrewood-Ältester Can't use spells
In the game, it is found through multiple tests，Monster Pyrewood-Ältester（entry 1895）Can't use spells Geringes Heilen（spell ID：2053）  
WoWScrnShot_112322_125938_proc

### Steps to reproduce

1.
2.
3.
4. ...
5. Profit

### Version & Environment

Client Version: 
  - 1.12.1.5875

Commit Hash:  

OS Client: Win 10
OS Server: Win [10]
### Crashlog

- None

schell244:
## 🍰 Pullrequest
Add [Lesser Heal](https://www.wowhead.com/classic/spell=2053/lesser-heal) ability to Pyrewood Elder

### Proof
Spellid: confirmed with SniffReplay core
Timer and Missing HP: CMangos

### Issues
- fixes https://github.com/vmangos/core/issues/1707

### How2Test
- .tele pyrewood 
- check if Pyrewood Elder casts heal on self and allies

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20230324094209_world.sql
+++ b/sql/migrations/20230324094209_world.sql
@@ -1,0 +1,28 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20230324094209');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20230324094209');
+-- Add your query below.
+
+-- Events list for Pyrewood Elder
+DELETE FROM `creature_ai_events` WHERE `creature_id`=1895;
+INSERT INTO `creature_ai_events` (`id`, `creature_id`, `condition_id`, `event_type`, `event_inverse_phase_mask`, `event_chance`, `event_flags`, `event_param1`, `event_param2`, `event_param3`, `event_param4`, `action1_script`, `action2_script`, `action3_script`, `comment`) VALUES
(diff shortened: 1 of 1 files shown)


ratkosrb:
What is the reason for casting the spell through EventAI instead of adding it to the creature spell list?

schell244:
Since the npc already had EventAI assigned (for flee action) and also cmangos had the spell as entry at their EventAI, I did it as well. There are also plenty other creatures that have spell's added through EventAI. When to use EventAI and when creature spell list? (When a creature already has EventAI assigned, is it no just less complicated to put all in one place?)

ratkosrb:
Cmangos did not have creature spell lists until recently, and did not do mass conversion from EventAI to spell lists after they added them, so that's why it's done that way in cmangos. Every spell that does not have complex conditions for casting and is cast repeatably during combat should be in spell lists. EventAI should only have spells that are cast on Aggro, on Death, on Emote, etc, like that where it's on some special event. All normal combat casts belong in the creature spell list.

When you put a spell cast in EventAI, the event cooldown will still reset to full even if the cast fails. Like if you have a 60 second delay between casts, and as the creature is casting, you go behind Line of Sight, the creature will not try to cast again until 60 more seconds. That is not how repeatable combat casts should work. The delay should work as a cooldown, which is what creature spell lists do. If you put it in the spell list, the creature will try casting the spell again as soon as there's a valid target, because the spell will only go on cooldown if the cast succeeds.

schell244:
Thanks for clarification. I will have a look on how spell list works and convert it.

PR #1865, opened 2023-04-12
Crashfix - Memory Lock

ghost:
Added additional locks to i_unitsRelocated

Crash here:
image


ratkosrb:
This should be fixed properly now.

PR #1872, opened 2023-04-29
Fix for threat not clearing upon death & Broadcast Text typo

ghost:
Removed else case that was prohibiting threat references upon death. Added here: https://github.com/vmangos/core/commit/41aa1220328650b103f112caf44684f71d2c5f26#diff-75b8a0a7d9e5a165e1d5218269eb821d74eb7ffec6cd06e8fa8ad82deceb9fa4


ratkosrb:
Good catch, but if you fix it this way, then these functions are getting called twice on death:

pVictim->RemoveCharmAuras(AURA_REMOVE_BY_DEATH);
pVictim->CombatStop();

The first time will be inside the call to SetDeathState in the if above. So only this part should be outside of the else statement.

pVictim->GetHostileRefManager().deleteReferences();

ghost:
@ratkosrb - changed based on your comments. Also added a new commit to this PR fixing a likely oversight on bg broadcast text,

ratkosrb:
The whole reason for BattleGroundBroadcastBuilder to exist is that it sends a broadcast_text, it is not an oversight, but the intended purpose of the function. Why would you think it's an oversight?

ratkosrb:
And I did not mean delete the calls to RemoveCharmAuras and CombatStop, but keep them inside the else block.

PR #1877, opened 2023-05-06
prevent totems dodge/parry

Gamemechanicwow:
## 🍰 Pullrequest

Prevent totems from parrying and dodging incoming attacks.

- GetUnitParryChance() was unexpectedly returning 5 for totems. This is changed to 0.
- RollMeleeOutcomeAgainst() no longer uses skilldifference for calculating dodge/parry for totems.

### Proof

- None

### Issues

- None

### How2Test

- Log onto 2 accounts. One of them has to be a shaman - the other has to be lower level than the shaman
- Shaman targets player on different account and types .m mainspeed 10
- Shaman places totem, targets it and types .m hp 99999
- Player on 2nd account autoattacks totem
- Notice if dodges/parries occurs

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -2145,6 +2145,7 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(Unit const* pVictim, WeaponAttackT
     }
 
     bool from_behind = !pVictim->HasInArc(this);
+    bool victimIsTotem = ((Creature const*)pVictim)->IsTotem();
 
     if (from_behind)
         DEBUG_FILTER_LOG(LOG_FILTER_COMBAT, "RollMeleeOutcomeAgainst: attack came from behind.");
@@ -2152,7 +2153,7 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(Unit const* pVictim, WeaponAttackT
     // Dodge chance
 
     // only players can't dodge if attacker is behind
-    if (!pVictim->IsPlayer() || !from_behind)
+    if ((!pVictim->IsPlayer() || !from_behind) && !victimIsTotem)
     {
         dodge_chance -= dodgeSkillBonus;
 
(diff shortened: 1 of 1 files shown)


src/game/Objects/Unit.cpp
```diff
@@ -2145,14 +2145,15 @@ MeleeHitOutcome Unit::RollMeleeOutcomeAgainst(Unit const* pVictim, WeaponAttackT
     }
 
     bool from_behind = !pVictim->HasInArc(this);
+    bool victimIsTotem = ((Creature const*)pVictim)->IsTotem();
```
0blu, line 2148: Is that a valid cast?

ratkosrb:
Isn't this handled by CREATURE_STATIC_FLAG_NO_DEFENSE ?

balakethelock:
CREATURE_STATIC_FLAG_NO_DEFENSE is currently implemented as making the unit have 0 defense skill, still able to dodge or parry albeit at a very low chance if the attacker has 300 weapon skill. This may or may not be accurate to vanilla.

Gamemechanicwow:
> Isn't this handled by CREATURE_STATIC_FLAG_NO_DEFENSE ?

Not entirely.

https://github.com/vmangos/core/blob/development/src/game/Objects/Unit.cpp#L2254
dodge_chance  = 0 and dodgeSkillBonus can be negative, thereby resulting in a positive dodgechance

https://github.com/vmangos/core/blob/development/src/game/Objects/Unit.cpp#L2270
parry is not affected by this.

its a different issue imo. closing this one.

0blu:
@Gamemechanicwow Why was this issue closed? If the current solution doesn't address all cases and your PR offers a valid fix, we should consider merging it.

ratkosrb:
I think the static flag should be made to prevent dodge/parry, it's likely what it's supposed to do.

Gamemechanicwow:
> @Gamemechanicwow Why was this issue closed? If the current solution doesn't address all cases and your PR offers a valid fix, we should consider merging it.

i think it's preferred to figure out first if a unit with CREATURE_STATIC_FLAG_NO_DEFENSE should be able to gain additional avoidance based on leveldifference. If it shouldn't it would be a much cleaner fix imo and would span wider than just totems.

PR #1887, opened 2023-05-15
movemapgen - minor fix

Gamemechanicwow:
## 🍰 Pullrequest

Vector3 up describes the direction up - and is used for checking whether a terrain vertice is inside a model. So unpassable terrain is filtered and NPCs can't move inside objects.

This small change can be summarized as:

- Don't do multiplications with 0
- Don't scale the direction, since it's used in a G3D::Ray where a direction always should have a length of 1. This also allows movemapgen to be run in debug - as it will no longer trigger an assertation.
- Don't adjust orientation as up is always up, no matter how the model is orientated. This removes some artifacts generated on boundaries between terrain and tilted models.

image

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/contrib/mmap/src/TerrainBuilder.cpp
+++ b/contrib/mmap/src/TerrainBuilder.cpp
@@ -751,9 +751,7 @@ namespace MMAP
             /// Check every map vertice
             // x, y * -1
             Vector3 up(0, 0, 1);
-            up.x *= -1.0f;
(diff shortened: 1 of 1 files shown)


celguar:
Any real world coordinates where it makes things better?

PR #1901, opened 2023-05-22
Fix hitnormal calculation part 2

Gamemechanicwow:
## 🍰 Pullrequest

Fixed another instance of hitnormal calculation - similar to https://github.com/vmangos/core/pull/1896
+ minor code cleanup (added some const, removed else after returns, commented unused args etc)

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Maps/Map.cpp
+++ b/src/game/Maps/Map.cpp
@@ -203,6 +203,8 @@ public:
         map->LoadGrid(c, true);
         return false;
     }
+
+private:
     Map* map;
 };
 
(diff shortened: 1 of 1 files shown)


src/game/Maps/Map.cpp
```diff
@@ -338,12 +340,10 @@ Map::EnsureGridCreated(GridPair const& p)
 void
 Map::EnsureGridLoadedAtEnter(Cell const& cell, Player* player)
 {
-    NGridType* grid;
+    NGridType* grid = getNGrid(cell.GridX(), cell.GridY());
 
     if (EnsureGridLoaded(cell))
     {
-        grid = getNGrid(cell.GridX(), cell.GridY());
```
ratkosrb, line 346: Won't calling getNGrid before EnsureGridLoaded mean it will always return nullptr if not currently loaded?

Gamemechanicwow: ye, I can see the problem now. I have reverted that part. Let's never talk about it again :>

PR #1920, opened 2023-06-03
Fix challengers from quest The Affray not automatically getting aggro

schell244:
## 🍰 Pullrequest
https://www.wowhead.com/classic/quest=1719/the-affray

### Proof
- None

### Issues
- None

### How2Test
- start the quest with a level 60 character or move out of spawn rage with level 30 character
- challenger will spawn but do not attack the player


schell244:
Closing this for now, due to missing proofs.

PR #1921, opened 2023-06-03
Add missing conditions to spawn npc for quest Verog the Dervish

schell244:
## 🍰 Pullrequest
- Verog the Dervish no longer spawns when players without the quest kill relatet npcs at Stagnant Oasis
- Optional: increase spawn chance from 2% to 10% (when player has quest). No proof for that. But it seems the spawn rate was very low to prevent too many spawns of Verog the Dervish from players without the quest.

### Proof
- None (Hateforge)

### Issues
- None

### How2Test
- accept quest https://www.wowhead.com/classic/quest=851/verog-the-dervish
- kill npcs which can trigger spwan (Kolkar Marauder, Kolkar Pack Runner, Kolkar Bloodcharger)
- afterwards try again killing same npcs without quest and notice that Verog the Dervish longer gets spawned

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20230603085543_world.sql
+++ b/sql/migrations/20230603085543_world.sql
@@ -1,0 +1,42 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20230603085543');
+IF v=0 THEN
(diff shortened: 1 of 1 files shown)


heyitsbench:
> Verog the Dervish no longer spawns when players without the quest kill relatet npcs at Stagnant Oasis

Are you sure this is correct? My source is unfortunately Wrath Classic so not incredibly reputable, but I was able to summon him without the quest in my log.

cpevors:
On Classic Era servers, I was able to summon him without the quest in my log too.

PR #1947, opened 2023-06-15
Clean, refactor, optimize, improve, and secure Common.cpp

tsvenbla:
## 🍰 Pullrequest
I've changed the code to modernize it and improve its safety and efficiency. Here are the key differences between the original code and my updated version:

1. I replaced the C-style arrays and raw pointers with modern C++ constructs like `std::map` and `std::optional`. This makes the code safer and easier to understand.

2. I switched from using C-style strings (`char const*`) to `std::string_view` for representing locale names. This provides more flexibility and efficiency in handling strings.

3. Instead of using global arrays `localeNames` and `fullLocaleNameList` to store locale names and their corresponding `LocaleConstant` values, I used a `std::map`. This provides efficient lookup and better encapsulation.

4. The original `GetLocaleByName` function returned a `LocaleConstant` and defaulted to `LOCALE_enUS` if the name was not found. In my updated version, the function returns a `std::optional<Locale>`, which can represent an absence of a value (using `std::nullopt`) when the name is not found. This provides a clearer indication of an error condition.

5. Similarly, in the `GetDbcLocaleFromDbLocale` function, I return a `std::optional<Locale>` instead of a `LocaleConstant`. This allows us to represent an invalid `localeIndex` with `std::nullopt`, indicating an error condition.

6. Instead of using a switch statement in `GetDbcLocaleFromDbLocale` to map `DBLocaleConstant` values to `LocaleConstant` values, I directly cast the `localeIndex` to `Locale`. This simplifies the code and reduces the chance of errors.

Overall, while the original code and my updated version achieve similar goals (mapping locale names to locale constants and vice versa), this update does so in a safer, more efficient, and more idiomatic way in modern C++.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/shared/Common.cpp
+++ b/src/shared/Common.cpp
@@ -20,66 +20,46 @@
  */
 
 #include "Common.h"
+#include <map>
 
-char const* localeNames[MAX_LOCALE] = {
-  "enUS",                                                   // also enGB
-  "koKR",
-  "frFR",
-  "deDE",
-  "zhCN",
-  "zhTW",
-  "esES",
-  "esMX",
-  "ruRU",
+enum class Locale {
+    enUS,
+    koKR,
+    frFR,
+    deDE,
+    zhCN,
+    zhTW,
+    esES,
+    esMX,
+    ruRU,
+    MAX_LOCALE
 };
 
-// used for search by name or iterate all names
-LocaleNameStr const fullLocaleNameList[] =
-{
-    { "enUS", LOCALE_enUS },
-    { "enGB", LOCALE_enUS },
-    { "koKR", LOCALE_koKR },
-    { "frFR", LOCALE_frFR },
-    { "deDE", LOCALE_deDE },
-    { "zhCN", LOCALE_zhCN },
-    { "zhTW", LOCALE_zhTW },
-    { "esES", LOCALE_esES },
-    { "esMX", LOCALE_esMX },
-    { "ruRU", LOCALE_ruRU },
-    { NULL,   LOCALE_enUS }
+constexpr Locale MAX_LOCALE = Locale::MAX_LOCALE;
+
+const std::map<std::string_view, Locale> localeNames = {
+    {"enUS", Locale::enUS},
+    {"koKR", Locale::koKR},
+    {"frFR", Locale::frFR},
+    {"deDE", Locale::deDE},
+    {"zhCN", Locale::zhCN},
+    {"zhTW", Locale::zhTW},
+    {"esES", Locale::esES},
+    {"esMX", Locale::esMX},
+    {"ruRU", Locale::ruRU}
 };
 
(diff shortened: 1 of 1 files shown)


Gamemechanicwow:
hey. It looks clean. Good stuff. 👍 
But unless I missed a memo we are staying at c++14 and std:optional is c++17 afaik.

tsvenbla:
Is there a reason we are not aiming for later C++ versions other than legacy code and backward compatibility? Have we tried using a later version before?

ratkosrb:
This project is committed to never raising compiler requirements. Code you submit should be able to compile with Visual Studio 2015.

tsvenbla:
> This project is committed to never raising compiler requirements. Code you submit should be able to compile with Visual Studio 2015.

Why is the project committed to never raising compiler requirements? While Visual Studio 2015 does support C++14, staying on it indefinitely might not be the best choice for several compelling reasons. One major issue with staying on VS 2015 is missing out on the numerous bug fixes and performance enhancements that newer versions bring. And while it may be true that we may not have encountered major issues yet, that doesn't guarantee we won't in the future. Plus, performance improvements in newer versions can still make our work more efficient. We should also consider the security aspect. Newer versions come with improved security features and patches to potential vulnerabilities, thus ensuring our code's safety.

There might be a short learning curve, but newer versions offer features that make development easier, like improved integration with version control systems, better refactoring tools, more efficient code editors, etc. 

Even if we're using C++14 now, it's quite possible we'll want to leverage features from newer standards like C++17 or C++20 in the future. Moreover, the compilers in newer versions of Visual Studio have better optimizations and produce more efficient code, regardless of the C++ standard we use.

If we are using third-party libraries, they are more likely to drop support for older compilers over time. By upgrading, we ensure compatibility with these crucial resources. Staying up-to-date prevents us from getting into a situation where we are forced to do a big jump between versions if VS 2015 becomes completely unsupported.

Remember, upgrading is about enhancing our capabilities, improving our security, ensuring our tools stay compatible, and saving ourselves potential future headaches. The slight inconvenience of learning some new features is far outweighed by these benefits.

Or am I completely in the wrong?

ratkosrb:
Using newer versions of Visual Studio is allowed and works without issue. What you may not do is write code that cannot compile in VS 2015. If your goal is to gain practice and experience on the newest c++ standards, then there are other WoW emulator projects for you such as TrinityCore, which requires the latest version of Visual Studio 2022 and cannot compile on anything older, due to their extensive use of brand new language features as soon as they are introduced. Our project is staying on the c++14 standard forever, and that is non negotiable. Blizzard was able to release their vanilla servers in 2004, written in c++98, surely we can recreate them using c++14 which is a decade newer. We have no need for anything more.

ghost:
Using C++17, this PR throws a ton of errors.

PR #1996, opened 2023-06-28
Fix XML Whitespace & Comments

Wall-core:
## 🍰 Pullrequest

Fixes XML Whitespace issues with some comments, and standardizes the format of all comments along with adjusting indentation and typos
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


Gamemechanicwow:
I believe many of the /// comments are meant to be used by doxygen or similar documentation generator software.

Wall-core:
> I believe many of the /// comments are meant to be used by doxygen or similar documentation generator software.

orly. Had no clue. There is a few issues however with XML Whitespace in Visual Studio due to comments which contain the following:
///< this comment is closed by>
when the comment is preceded and closed by <>. Not sure what the proper way to handle it would be if this is the case, however. Are you familiar with all of this?

PR #2086, opened 2023-07-15
Hammer of Wrath does not use weapon skill

balakethelock:
## 🍰 Pullrequest

Recent commit to make special abilities use weapon skill competely broke Hammer of Wrath. Made it use ranged weapon skill (defaulted to 0 on paladin who don't have ranged weapon skill to begin with, the bug made the ability miss more than 50% of the time when fighting same level enemies)

Hammer of wrath seems to be the sole exception that **shouldn't** use weapon skill ever.

### Proof

[Videos from classic ptr](https://discord.com/channels/484741332850573317/484741332850573323/1129568118440734871) thanks to Brotalnia

### Issues

- None

### How2Test

- 

https://github.com/vmangos/core/assets/111737968/feccca46-3a6d-45bd-8d09-4ac245d48d2a

### Todo / Checklist

- Stop Hammer of Wrath from leveling up the equipped weapon skill maybe..?

### DIFF
--- a/src/game/Objects/SpellCaster.cpp
+++ b/src/game/Objects/SpellCaster.cpp
@@ -382,7 +382,7 @@ SpellMissInfo SpellCaster::MeleeSpellHitResult(Unit* pVictim, SpellEntry const* 
         return SPELL_MISS_NONE;
 
     // bonus from skills is 0.04% per skill Diff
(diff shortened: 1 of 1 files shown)


ratkosrb:
I addressed it another way.
https://github.com/vmangos/core/commit/df946fb683eeba332fd2cd0c66477be1b9bdeff7

PR #2119, opened 2023-07-26
Fix threat values of warrior abilities

balakethelock:
## 🍰 Pullrequest

Small changes to threat of warrior abilities to match them with values from classic testing.

### Proof

- https://github.com/magey/classic-warrior/wiki/Threat-Mechanics#threat-values
- https://docs.google.com/spreadsheets/d/1VLmhdNX_hZcud8Q0n2I1fPv7y8g8BHYhH_cmIkmd_-s/edit#gid=1992655717
- https://docs.google.com/spreadsheets/d/1z_I0oU5IOmQA5RQX26xOvj3ycLRMTVoZapFLY7OG-8c/edit#gid=1414903181

From these I discovered that most warrior abilities have static threat which is just spell level * (static multiplier)
- The classic testers found that revenge R1 (lvl 14) did 63 static threat, R5 (lvl 54) did 243 static threat, and R6 (lvl 60) did 270
63/14 = 243/54 = 270/60 = 4.5 from that I filled in the remaining ranks
- The classic testers found that sunder armor R1 (lvl 10) did 45 static threat, R5 (lvl 58) did 261 static threat.
45/10 = 261/58 = 4.5 same as Revenge, from that I filled in the remaining ranks
- I repeated this process for Shield Bash (static multiplier = 3 )
- For Hamstring (static multiplier = 2.5)
- For Battle Shout (static multiplier = 1)
- For Demoralizing Shout (static multiplier = 0.8), 

Note: static multiplier is just a hidden constant to calculate how much static threat the ability does based on its spell level. It is different from the multiplier in spell_threat table which is for how much threat the ability will do based on its damage.

### Issues

- None

### How2Test

- Create a warrior character, level up to 60
- `.add 6619` use it then `.learn all_trainer`
- teleport to a level 59-60 mob .go creature 41697
- test abilities and use `.list threat` to see how much threat they're generating
I tested all the abilities individually.
Made sure static threat is applied BEFORE the multiplier for Revenge and Shield Bash and Hamstring.
Made sure thunder clap has the bonus threat on all ranks.

Use bloodthirst or pummel on the mob, find that it lands normally because it's using default (level * 5) skill, not your current weapon skill as it should.
### Todo / Checklist

- [X] Heroic Strike and Cleave data is missing for lower ranks. Doesn't appear to use the `"static threat = spell level * static multiplier"` rule of other abilities
- [X] Other classes, especially bear demoralizing roar.

### DIFF
--- a/sql/migrations/20230726131045_world.sql
+++ b/sql/migrations/20230726131045_world.sql
@@ -1,0 +1,71 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20230726131045');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20230726131045');
+-- Add your query below.
+
+UPDATE `spell_threat` SET `Threat` = 126 WHERE `entry` = 21992; -- thunderfury single target
+UPDATE `spell_threat` SET `Threat` = 126 WHERE `entry` = 27648; -- thunderfury chain
+
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`, `build_min`, `build_max`) VALUES (676, 99, 1, 0, 0, 5875); -- Disarm
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`, `build_min`, `build_max`) VALUES (7384, 0, 0.75, 0, 0, 5875); -- Overpower
+
+UPDATE `spell_threat` SET `Threat` = 178 WHERE `entry` = 23922; -- Shield slam R1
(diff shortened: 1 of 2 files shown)


Kittnz:
can we back this info with sniffs?

balakethelock:
I don't know if threat is a packet that can be sniffed. If it is then yeah maybe?

Daribon:
> I don't know if threat is a packet that can be sniffed. If it is then yeah maybe?

Isn't threat sent in SMSG_THREAT_UPDATE on classic patch 1.13.5 and beyond?
EmEditor_3LcTg51akT

But going by sniff is basically same thing as just doing the follow on classic era: 
`/run local _,_,_,_,t=UnitDetailedThreatSituation("player","target") if t then DEFAULT_CHAT_FRAME:AddMessage(t) else DEFAULT_CHAT_FRAME:AddMessage("No threat") end`
or just
`/dump UnitDetailedThreatSituation("player", "target")`

Blizz sends the threat data in UnitDetailedThreatSituation in classic era:
https://wowpedia.fandom.com/wiki/API_UnitDetailedThreatSituation

PR #2162, opened 2023-08-14
Fix ravager axe not properly proccing deep wounds, weapon enchants et…

### LINKED ISSUE #1786: Ravager Axe🐛 [Bug]
## 🐛 Bug report
Right now the item Ravager is super weak. It has 3 bugs. 
1. When the enemy you were targetting when the Channel starts dies, the channeling stops.
2. It does not proc of itself.
3. It does not proc any enchant on the weapon or trinket.
[edit] 4. Crits from Ravager when Channeling Whirlwind don't trigger Deep Wounds.

### Expected behavior
Here you can see ravager proccing Fiery weapon. https://www.youtube.com/watch?v=PbtqPFx3qoM
Here you can see ravager proccing off itself and not being cancelled when targets die. https://www.youtube.com/watch?v=7BARDYrbFQM

### Steps to reproduce

1. A dev better skilled than me will know better how to reproduce the bugs described here.

### Version & Environment

Client Version: 
  - 1.4.2.4375

Commit Hash:  

OS Client: 
OS Server: 

### Crashlog

- None

scottnice:
## 🍰 Pullrequest

Fix ravager axe not properly proccing deep wounds, weapon enchants etc on bladestorm hits, fixed ravager axe issue where channeling would stop when target died, extra attacks generated while channeling will now be applied after channelling is complete, this matches behaviour of classic. Fix issue with ravager bladestorm where its whirlwind hits could not reproc blade storm.

### Proof
Windfury procs during bladestorm are stored and released after channelling is complete: 
https://www.wowhead.com/classic/item=7717/ravager#comments:id=3160508
https://www.youtube.com/watch?v=XEZpNdxlV1A

### Issues
- fixes #1786 

### How2Test
Create a character with ravager on a shaman and a warrior. Test with windfury and sweeping strikes. 
Ravager whirlwind procs also do not consume sweeping strikes matching the functionality in the linked youtube video from Classic SoM

### Todo / Checklist

- [x] Fix bladestorm and sweeping strikes
- [x] Fix White space issues


Wall-core:
the spell_template data is taken directly from the client dbc and is blizzlike. If absolutely necessary, this would be handled under spell_effect_mod or spell_mod but would be better resolved in the core as it's best to avoid adding mods.

scottnice:
Thanks Wall, I'll update it to use the spell_mod table for now but I'll also look to see if there's a way to handle this in the core.

scottnice:
@Wall-core @ratkosrb it's possible the fields changed between versions. Because according to this commit https://github.com/vmangos/core/commit/3f91713bed5b8f91046ebfbc9a903cee6faa87ca at one point the bladestorm ability had the SPELL_ATTR_EX2_TRIGGERED_CAN_TRIGGER_PROC flag set, and this behavior has now moved over to the SPELL_ATTR_EX3_NOT_A_PROC flag which is not set on these spells. If the same migration in the code happened in the database one would expect the SPELL_ATTR_EX3_NOT_A_PROC flag to be set on the Bladestorm and its Whirlwind proc.

ratkosrb:
Both the attributes assigned to spells you see in spell_template, and the naming of attributes in the source code comes from Blizzard. None of these should be changed.

scottnice:
> Both the attributes assigned to spells you see in spell_template, and the naming of attributes in the source code comes from Blizzard. None of these should be changed.

That makes sense. Was just wondering why SPELL_ATTR_EX2_TRIGGERED_CAN_TRIGGER_PROC changed to SPELL_ATTR_EX2_ACTIVE_THREAT without any database values migrating.

scottnice:
@Wall-core @ratkosrb Updated to use spell_mod table and made some small code changes to support proper functioning of Ravager without impacting other abilities.

src/game/Objects/Unit.cpp
```diff
@@ -276,7 +276,7 @@ void Unit::Update(uint32 update_diff, uint32 p_time)
 
     // extra attack
     Unit* victim = GetVictim();
-    if (IsInCombat() && m_doExtraAttacks && GetExtraAttacks() && victim && (CanAutoAttackTarget(victim) == ATTACK_RESULT_OK))
+    if (IsInCombat() && m_doExtraAttacks && GetExtraAttacks() && victim && (CanAutoAttackTarget(victim) == ATTACK_RESULT_OK) && !IsNonMeleeSpellCasted(false, false, true))
```
scottnice, line 279: This allows extra attacks to be applied after spell casting and channeling are complete, which is the behaviour in classic.

scottnice:
@ratkosrb some of this code can be replaced with overrides via the spell mod table. Let me know which path you prefer and I can update the pull request accordingly.

mserajnik:
Did this get fixed outside of this PR?

scottnice:
> Did this get fixed outside of this PR?

Not that I'm aware of

PR #2117, opened 2023-07-24
Spell groups and aura exclusivity

balakethelock:
## 🍰 Pullrequest

- Cleaned up spell groups table. 80% of the groups were duplicates, non-functional, or redundant from IsNoStackSpellDueToSpell.
- Rewrote wotlk code to something that makes much more sense for vanilla. Removed some exclusivity rules that had no logical reason to be there in the first place.

### Proof

- Spreadsheet documenting the new db tables in easy to read format [here](https://docs.google.com/spreadsheets/d/1Yfx8n_GjYcdyMcj1qWO5Xbc9iLibdKemDxPqtmCvHjg/edit?usp=sharing)

### Issues

- fixes #1884 

### How2Test

- None

### Todo / Checklist

- [X] None

PR #2167, opened 2023-08-18
Fix Elemental sharpening stones applied to an offhand weapon

### LINKED ISSUE #2165: 🐛 [Bug] Elemental Sharpening stone on offhand
## 🐛 Bug report

If you apply an elemental sharpening stone to your MH weapon, the crit is applied to both oh and mh attacks. But an elemental sharpening stone on OH doesn't add crit to any attacks.

**Small investigation:**

https://github.com/vmangos/core/blob/development/src/game/StatSystem.cpp#L475
We set the ranged and mh crit to appropriate playerfields - but offhand doesn't exist, so we don't store the value.

https://github.com/vmangos/core/blob/development/src/game/Objects/Unit.cpp#L2712
we read crit chance for oh and mh from the mainhand crit chance we set in playerfields 

Result MH crit used for both attacktypes - OH crit ignored.

### Expected behavior

I am not sure if elemental sharpening stone crit chance should apply to both weapons or only the weapon that its applied on.
But it should have an effect applying an elemental sharpening stone to the oh.

**Research on blizzlike behavior is needed before a fix can be applied imo**

### Steps to reproduce

1. Change crit percentage to better observe behavior:
UPDATE `spell_template` SET `effectBasePoints1`=99 WHERE  `entry`=22755 AND `build`=5302;
(don't forget to set it back to 1 when you are done testing :> )
2. Apply elemental sharpening stone to either MH and OH
3. Change attackspeed to faster gather data. 
.m attackspeed 100 - or .m offhand 100
4. hit stuff - observe crits or lack of crits

### Version & Environment

Client Version: 
  - 1.12.1.5875
 

Commit Hash:  

OS Client: 
OS Server: 

### Crashlog

- None

Gamemechanicwow:
## 🍰 Pullrequest

As described in https://github.com/vmangos/core/issues/2165 - applying an elemental sharpening stone to an offhand weapon have zero effect. 

This PR makes it consistent that Elemental sharpening stone whether applied to MH and OH always adds +2% to all melee attacks.

This PR also removes OFFHAND_CRIT_PERCENTAGE from BaseModGroup -as it no longer is needed.

### Proof

Crit damage in spellbook on classic is increased by 2% for each stone.

image

From FC discord:
image
image
--
image

1.10 patch notes for reference:
image

This is the closest to any evidence I could find. There are currently no raid consumes boxes on era ptr, so i have not personally done any data gathering.

### Issues

- Fixes https://github.com/vmangos/core/issues/2165 

### How2Test

- None

### Todo / Checklist

- [X] None


balakethelock:
This bug might be related to the bigger bug with auras stacking rules.
If you look [here](https://github.com/vmangos/core/blob/cb50113a514ae9decce0cfda2e62431aa61efbcc/src/game/Spells/SpellAuras.cpp#L8305C17-L8305C17) you'll see that whoever wrote this code years ago (in french) tried to make Elemental Sharpening Stone non-exclusive (i.e. stacks with everything else) but this person was very confused cause he put the item ID of sharpening stone where he should have put the spell ID of the aura it gives (22755)

You'll find that's not the only error in that code.
I went into some detail about what's broken in [here](https://github.com/slowtorta/turtlewow-bug-tracker/issues/3994#:~:text=Too%20big%20to%20list%2C%20but%20there%20are%20many%20bug%20reports%20all%20directly%20caused%20by%20this)


And proposed a fix here https://github.com/vmangos/core/pull/2117

wwElement:
Using brotalnias v22 pack from June 23rd of this year, confirming code not doing anything at all when you apply offhand sharpening stone:
image
Data using dpsmate

Important part - over the course of 30k auto attacks and 8k hamstrings, difference between no sharpening stone and offhand sharpening stone **was 0,3 % and 0,01%** respectively. With main hand sharpened, over the same amount of auto attacks and hamstrings, the difference in crit **was 1,85% and 1,33%**, showing a clear change.

I modified attack speeds to only get main hand swings when main hand was enchanted and the same with offhand attacks.

image
sheet for easier viewing

Issue #2166, opened 2023-08-16
🐛 [Bug][Errors][Console] Loading Aggro Spells Definitions.


gajet5:
## 🐛 Bug report
Loading Aggro Spells Definitions...
Spell 6196 listed in `spell_threat` has effects with different targets, threat may be assigned incorrectly
Spell 126 listed in `spell_threat` has effects with different targets, threat may be assigned incorrectly
Spell 633 must be listed in `spell_threat` as first rank for listed custom ranks of spell but not found!
Spell 8092 must be listed in `spell_threat` as first rank for listed custom ranks of spell but not found!
Spell 16811 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16812 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16813 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 17329 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16812 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16813 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 17329 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16813 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 17329 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 17329 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16810 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16811 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16812 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 16813 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 17329 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant
Spell 25918 listed in `spell_threat` as custom rank has same data as Rank 1, so redundant

### Expected behavior
No errors

### Steps to reproduce

1. Update server

### Version & Environment
Client Version: 1.12.1.5875

Commit Hash:  No

OS Client: Win 11
OS Server: Ubuntu 22.02

### Crashlog
- None

ajbathe:
-- Question:

USE mangos;
[...]

PR #2188, opened 2023-09-03
Correct threat for more spells.

### LINKED ISSUE #2163: 🐞 [Bug] Touch of Weakness incorrect threat.
## 🐞 Bug report - Spell

Touch of Weakness currently does no threat on when used. The threat caused by its proc damage is incorrect.

### Expected behavior

Touch of Weakness should cause threat when you use the ability.
Touch of Weakness should do bonus threat on proc.

Here is the correct threat for casting the spell:
r6: 60
r5: 50
r4: 40
r3: 30
r2: 20
r1: 10

Here is a fix for the cast threat:
```-- Touch of Weakness R1
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (2652, 10, 0, 0);
-- Touch of Weakness R2
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19261, 20, 0, 0);
-- Touch of Weakness R3
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19262, 30, 0, 0);
-- Touch of Weakness R4
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19264, 40, 0, 0);
-- Touch of Weakness R5
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19265, 50, 0, 0);
-- Touch of Weakness R6
INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19266, 60, 0, 0);
```

The problem is that the proc threat is incorrect and I can't find a way to fix it through the spell_threat table.

Every Touch of Weakness uses the same spell with spellid: 28598.
However, every rank should cause different threat when procced.
The threat proc should be threat damage + 2x the bonus threat of the cast.

For example a rank 6 proc if it dealt 111 damage would cause 111 threat + 120 bonus threat, resulting in 231 threat.
WowClassicT_X1yErVbUk8
WowClassicT_KtkjfzH4Vf

Same thing goes with for example r1 dealing 56 damage. It would cause 56 damage threat + 20 bonus threat, resulting in 76 threat.
WowClassicT_R9bqJgAse3
WowClassicT_vjVBIELeck

### Steps to reproduce

1. Create a priest
2. .learn 19266
3. Walk near enemy mob
4. Cast Touch of Weakness
5. .list threat

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  https://github.com/vmangos/core/commit/b6eef3471c09a487941e6dc664accab8422ca844

OS Client: Windows 11 Pro N 10.0.22621.1413
OS Server: Windows 11 Pro N 10.0.22621.1413

### Crashlog

- None

Daribon:
## 🍰 Pullrequest

This PR corrects threat to spells that I previously missed.
### Proof

- I've confirmed the threat values on PTR, by requesting threat from server with the following macro: ```/run local _,_,_,_,t=UnitDetailedThreatSituation("player","target") if t then DEFAULT_CHAT_FRAME:AddMessage(t) else DEFAULT_CHAT_FRAME:AddMessage("No threat") end```

Currently this warning shows up while server is loading:
WindowsTerminal_knRbyQHQV4
However all the spells in that list do correct threat in game.

### Issues

- fixes #2163
- "fixes" #2166

### How2Test

- pull any enemy mob and .cast any of the spells in the SQL.

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20230903173317_world.sql
+++ b/sql/migrations/20230903173317_world.sql
@@ -71,9 +71,9 @@ DELETE FROM `spell_threat` WHERE  `entry`=19742;
 -- Feral Charge
 INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (16979, 52, 0, 0);
 
--- Bash R3
+-- Bash R1
 INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (5211, 42, 0, 0);
--- Bash R3
+-- Bash R2
 INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (6798, 90, 0, 0);
 -- Bash R3
 INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (8983, 138, 0, 0);
--- a/sql/migrations/20230903173317_world.sql
+++ b/sql/migrations/20230903173317_world.sql
@@ -1,0 +1,134 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20230903173317');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20230903173317');
+-- Add your query below.
+
+
+-- Touch of Weakness R1
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (2652, 10, 0, 0);
+-- Touch of Weakness R2
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19261, 20, 0, 0);
+-- Touch of Weakness R3
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19262, 30, 0, 0);
+-- Touch of Weakness R4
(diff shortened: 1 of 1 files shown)


sql/migrations/20230903173317_world.sql
```diff
@@ -0,0 +1,134 @@
...
+-- Touch of Weakness proc R4
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19252, 80, 1, 0);
+-- Touch of Weakness proc R5
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19253, 100, 1, 0);
+-- Touch of Weakness proc R6
+INSERT INTO `spell_threat` (`entry`, `Threat`, `multiplier`, `ap_bonus`) VALUES (19254, 120, 1, 0);
+
+-- Remove Blessing of Wisdom
+-- It had 0 threat value and should not exist in spell_threat
+DELETE FROM `spell_threat` WHERE  `entry`=19742;
```
ratkosrb, line 69:
```
World of Warcraft Client Patch 1.9.0 (2006-01-03)
- Blessing of Wisdom - Mana regeneration over time will no longer
  generate threat.
  ```

Have you tested if it still generates threat pre 1.9 with this deletion? We should not impose the 1.12 state on previous patches.

Daribon: All I did was remove the 0 threat from entry 19742 which shouldn't have been there in the first place since it defaults to 0 threat if spell doesn't exist in spell_threat. With this PR and without this PR Blessing of Wisdom generates 0 threat when cast.

The patch note you mention is not currently implemented in vmangos, mp5 from Blessing of Wisdom is currently not able to cause threat in 1.8.4 or below. I tested this with no mana with Blessing of Wisdom applied to 100% mana and no threat was generated, even without this PR.

PR #2243, opened 2023-10-21
Add Corrupted Blood config

Wall-core:
## 🍰 Pullrequest

Adds config option to enable the original Corrupted Blood event. Tested in 1.12.1, should work in older versions, not sure if 1.14 has spellid #24330 visuals but it should.
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Threat/ThreatManager.cpp
+++ b/src/game/Threat/ThreatManager.cpp
@@ -55,9 +55,9 @@ float ThreatCalcHelper::CalcThreat(Unit* pHatedUnit, float threat, bool crit, Sp
 //================= HostileReference ==========================
 //============================================================
 
-HostileReference::HostileReference(Unit* pUnit, ThreatManager *pThreatManager, float pThreat)
+HostileReference::HostileReference(Unit* pUnit, ThreatManager *pThreatManager, float threat)
 {
-    iThreat = pThreat;
+    iThreat = threat;
     iTempThreatModifyer = 0.0f;
     link(pUnit, pThreatManager);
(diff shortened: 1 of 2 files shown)


ratkosrb:
I'm pretty sure that was changed in a minor patch and that's why it's not a thing on vmangos. It probably was only a thing on 1.7.0 not on 1.7.1.

codestothestars:
Maybe a stupid question, but is there a difference in the spell data between 1.7.0 and 1.7.1?

Gamemechanicwow:
> Maybe a stupid question, but is there a difference in the spell data between 1.7.0 and 1.7.1?

spell.dbc entries for 24328 and 24330 are the same in 1.7.0.4671 and 1.7.1.4695

Wall-core:
I spent most of the morning reading over what info I could in the client and DB, from what it seems it could have been a hotfix serverside by blizzard. Spell data supports this notion that it was an issue with maxaffectedtargets used. This config reenables the original functionality of the spell (capped at 100 targets per tick for speed sake), and has checks to use original spell data in 1.7 client builds.

wdstudios2:
> I'm pretty sure that was changed in a minor patch and that's why it's not a thing on vmangos. It probably was only a thing on 1.7.0 not on 1.7.1.

No, the plague occurred throughout both 1.7.0 and 1.7.1

I'm not sure what the hotfixes did, or were supposed to do, but the behavior of Corrupted Blood **_inside the raid instance_** did not change until 1.8

PR #2273, opened 2023-11-10
Make RP decay adjustable.

mserajnik:
## 🍰 Pullrequest
This makes it so the rank points decay can be adjusted to a different percentage than the default 20%.

### Proof
- None

### Issues
- None

### How2Test
- Earn contribution points
- Trigger rank points distribution through honor maintenance
- Observe rank points getting calculated based on configured decay

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/World.h
+++ b/src/game/World.h
@@ -449,6 +449,7 @@ enum eConfigFloatValues
     CONFIG_FLOAT_RATE_XP_PERSONAL_MAX,
     CONFIG_FLOAT_AC_MOVEMENT_CHEAT_TELEPORT_DISTANCE,
     CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE,
+    CONFIG_FLOAT_RP_DECAY,
     CONFIG_FLOAT_VALUE_COUNT
 };
 
@@ -1047,7 +1048,7 @@ class World
         std::unique_ptr<MovementBroadcaster> m_broadcaster;
 
         std::unique_ptr<ThreadPool> m_updateThreads;
-        
+
         static uint32 m_currentMSTime;
         static TimePoint m_currentTime;
         static uint32 m_currentDiff;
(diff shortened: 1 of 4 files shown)


ratkosrb:
Don't you need to declare CONFIG_UINT32_RP_DECAY_PERCENTAGE in World.h first? I see no change in this file. Also why not use a float option instead of uint32.

mserajnik:
> Don't you need to declare CONFIG_UINT32_RP_DECAY_PERCENTAGE in World.h first

Of course, sorry. I didn't want to commit a bunch of whitespace changes too (since my editor configuration normally trims trailing whitespace automatically), so I removed those manually before pushing and then somehow missed to copy over the change in `World.h`. I've now added it.

> Also why not use a float option instead of uint32.

I considered it easier from the user's point of view to think `20 percent` instead of `0.2`. But I just saw you already have a float configuration like that for `Death.Ghost.RunSpeed.World`, for example, so I think it makes sense to adjust mine to use a float too. __Edit__: I've adjusted it.

src/game/World.h
```diff
@@ -449,7 +449,8 @@ enum eConfigFloatValues
     CONFIG_FLOAT_RATE_XP_PERSONAL_MAX,
     CONFIG_FLOAT_AC_MOVEMENT_CHEAT_TELEPORT_DISTANCE,
     CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE,
-    CONFIG_FLOAT_VALUE_COUNT
+    CONFIG_FLOAT_VALUE_COUNT,
+    CONFIG_FLOAT_RP_DECAY
```
ghost: This needs to be above "CONFIG_FLOAT_VALUE_COUNT"

mserajnik: Whoops, thanks! Fixed in https://github.com/vmangos/core/commit/bd5841a4b84941f6148a66ac4568f59d9cfd0289.

ratkosrb: How do so many people make this mistake, I don't know. It's not the first time it has happened.

mserajnik: Speaking only for myself, I wasn't familiar with this technique of getting the number of elements in an enum, though I won't use that as excuse since the naming `CONFIG_FLOAT_VALUE_COUNT` is pretty clear. 😄

PR #2295, opened 2023-11-19
Fix whitespace in VMAP extractors

Wall-core:
Fixes whitespace in filenames, which client normally parses and replaces with underscores (razorfen leanto03.m2)


mserajnik:
With this fix I still see the following warning during the extraction (running on Ubuntu 22.04 LTS):
```
Warning: Some models could not be extracted, see below
Could not find file of model World\Generic\Quilboar\Passive Doodads\Leantos\Razorfen_Leanto03.m2
```
I thought this was supposed to fix the extraction of that model; or am I mistaken and this fix is for something else?

PR #2301, opened 2023-11-21
Battlebots AutoJoin customizable.

gajet5:
Sometimes you need to clean the BG, but do not disable the auto join function. For example, when atari lasts for several days.

_My mistake, I didn't need to delete the project in my account._


oowow-core:
is it possible to control by batttle ground template? thanks!
eg
.battlebot autojoin on warsong
.battlebot autojoin on alterac

PR #2311, opened 2023-11-24
Fix a few mage talents that were changed in 1.11 for old versions.

TuriansNotBad:
## 🍰 Pullrequest

- Mage talent Shatter was changed to affect all spells, previously limited to Frost spells.
- Improved Frost Ward was changed to Frost Warding. Old effect - 50% of the damage absorbed by your Frost Ward is added to your mana.
- Improved Fire Ward was changed. Old effect - Causes your Fire Ward to reflect 20%/35% of the Fire damage absorbed back to the caster.

Tested 1.10 and 1.12

### Proof

- None

### Issues

- None

### How2Test

On 1.10 client:
- Check that Shatter only affects Frost spells.
- Check that Improved Fire Ward does % absorbed damage to attacker. 20% at rank 1, 35% at rank 2.
- Check that Improved Frost Ward converts 50% of absorbed damage to your mana.
- Newer clients behave as before.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -1876,6 +1876,27 @@ void Unit::CalculateDamageAbsorbAndResist(SpellCaster* pCaster, SpellSchoolMask 
             // Need remove it later
             if (mod->m_amount <= 0)
                 existExpired = true;
+
+#if SUPPORTED_CLIENT_BUILD <= CLIENT_BUILD_1_10_2
+            // Patch 1.11.0 changed Mage talent Improved Frost Ward to Frost Warding.
+            // Improved Frost Ward - 50% of the damage absorbed by your Frost Ward is added to your mana.
+            const SpellEntry* absorbProto = (*i)->GetSpellProto();
+            if (absorbProto->IsFitToFamily<SPELLFAMILY_MAGE, CF_MAGE_FROST_WARD>() && HasAura(11189))
+                CastCustomSpell(this, 27679, currentAbsorb / 2, {}, {}, true);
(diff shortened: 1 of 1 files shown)


ratkosrb:
I have an alternative idea for Improved Frost and Fire Wards. We can utilize proc system instead of having checks for these spells in CalculateDamageAbsorbAndResist. Can you try it out?

Add rows in spell_proc_event for the talent spells, set chance to 100% there, proc flags to 131072 (take harmful spell), proc ex to 1027 (hit + crit + absorb), and then add proc handlers for the dummy auras from the talents in Unit::HandleDummyAuraProc inside UnitAuraProcHandler.cpp. You can get the absorbed amount by doing (originalAmount - amount).

TuriansNotBad:
@ratkosrb Can't see a way to tell which spell has absorbed damage. procSpell will have the damage spell like Frostbolt, triggeredByAura will be the talent. Need to be able to tell Frost/Fire Ward from Greater Frost/Fire Protection Potion and other absorbs.

Can't set the proc to happen on the Frost/Fire Ward spells either cos the aura is instantly removed in CalculateDamageAbsorbAndResist if broken.

ratkosrb:
Damn, guess we cant do it with procs then.

balakethelock:
Is there proof that Frost Ward only restored mana from spells absorbed by it? It's not impossible that Blizzard implemented it differently than the tooltip implies, and it restored mana from all absorbs as long as it was up on the player.

Same question for Fire Ward reflecting all absorbed damage, not just the ones absorbed by it.

TuriansNotBad:
I have no proof either way.

These talents were not implemented at all and all I had for their function was the tooltip description.

PR #2322, opened 2023-11-27
very minor refactor of Creature::HasSpell

Gamemechanicwow:
## 🍰 Pullrequest

Latest GCC and Clang already optimizes this with -O3 - MSVC doesn't (with -O3)

MSVC Example 
image

It also simplifies the code and makes it more readable.

minor note for optimization nerds.. The following snippet is often more faster on gcc/clang - but comes with its own set of drawbacks..
   return (spellId == m_spells[0] ||spellId == m_spells[1] || spellId == m_spells[2] ||spellId == m_spells[3]  );

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Objects/Creature.cpp
+++ b/src/game/Objects/Creature.cpp
@@ -3053,8 +3053,7 @@ bool Creature::IsInEvadeMode() const
 
 bool Creature::HasSpell(uint32 spellId) const
 {
-    uint8 i;
-    for (i = 0; i < CREATURE_MAX_SPELLS; ++i)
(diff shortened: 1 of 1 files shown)


src/game/Objects/Creature.cpp
```diff
@@ -3056,8 +3056,8 @@ bool Creature::HasSpell(uint32 spellId) const
     uint8 i;
```
ratkosrb: The declaration of 'i' can go inside the for after that change.

PR #2426, opened 2024-01-10
Partybot in combat target selection AI

### LINKED ISSUE #2406: 🐛 [Bug]Partybots don't properly focus marked targets
## 🐛 Bug report
This code states they should focus marked targets in combat:
https://github.com/vmangos/core/blob/development/src/game/PlayerBots/PartyBotAI.cpp#L373
What they do is will focus the target of the party leader or whatever is hitting them or a target at random completely(?). Most of the time it seems to be the party leader's target upon first engage. They never switch targets until their current target is dead, but the code makes it look like they should change target and go to the marked target then stick on it the entire time in combat, but the don't.

### Expected behavior
Mark a target, party bots focus it no matter what. Is not the the case.

### Steps to reproduce

1. Spawn a dps partybot
2. mark a target
3. engage something not the marked target
4. watch as the bots not swap to the marked target until their first target is dead
5. Profit

### Version & Environment

Client Version:  1.12.1

Commit Hash:  newest

OS Client:  win10
OS Server:  win10

### Crashlog

- None

ShiyoKozuki:
## 🍰 Pullrequest
Original Behavior:
Partybots do not swap to the party leaders target or swap to a marked target after the focusmark in game command is executed while in combat. 

Corrected Behavior:
Partybots properly target a marked target(if focusmark command is executed) or the party leaders target while in combat. Priority is marked(if focusmark command executed) - > Party leaders target(if no mark or command isn't executed).

### Proof

- None

### Issues
- fixes #2406

### How2Test

Summon party bots, type .partybot focusmark skull. engage an enemy.  Mark a new enemy with skull, bots will swap to it.
Summon party bots, engage an enemy, swap target(do not mark), partybots will swap to your target

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/PlayerBots/PartyBotAI.cpp
+++ b/src/game/PlayerBots/PartyBotAI.cpp
@@ -960,6 +960,27 @@ void PartyBotAI::UpdateInCombatAI()
             return;
     }
 
+    Unit* pVictim = me->GetVictim();
+
+    // Swap to marked target or party leader's target
+    if (GetRole() != ROLE_HEALER)
+    {
+        if (Player* pLeader = GetPartyLeader())
+        {
+            Unit* newVictim = SelectAttackTarget(pLeader);
+
+            if (newVictim && (newVictim != pVictim))
+            {
+                if (pVictim)
+                    me->AttackStop();
+                else
+                    AttackStart(newVictim);
+
+                return;
+            }
+        }
+    }
+
     switch (me->GetClass())
     {
         case CLASS_PALADIN:


gajet5:
Will this work in a raid if the bots were not called by the leader? For example, the hiller leader. which does not target the goals.

ShiyoKozuki:
> Will this work in a raid if the bots were not called by the leader? For example, the hiller leader. which does not target the goals.

AFAIK it assigns the summoner as pt leader, not the real party leader in game
This is how it was aleady set up for pt leader assignment and was not changed

gajet5:
Okay, I got it. Thanks, I will test it.

gajet5:
I have checked this functionality and it works well. They really started hitting targets. However, there is a downside to this update. Without this update, the role of the tank was trying to run around the NPCs and tried to provoke them. Now he hits one target and the NPCs kill the healer.

GoonBoiBarry:
> I have checked this functionality and it works well. They really started hitting targets. However, there is a downside to this update. Without this update, the role of the tank was trying to run around the NPCs and tried to provoke them. Now he hits one target and the NPCs kill the healer.

You can assign targets to specific bots by targeting them individually and typing the command. If you do this to everyone but the tank, the tank will work the way it worked before. This way you can have tanks dedicated to tanking 1 target and general tanks etc.

gajet5:
I understand it. But it won't work on trash. And I'm talking about him first of all. Bosses don't cause problems. I need more tests, I'll try to exclude tanks from this logic. 

In any case, the DPS roles have become more responsive.

gajet5:
```c++
    // Swap DPS to marked target or party leader's target
    if (m_role == ROLE_MELEE_DPS || m_role == ROLE_RANGE_DPS)
    {
        Unit* pVictim = me->GetVictim();

        if (Player* pLeader = GetPartyLeader())
        {
            Unit* newVictim = SelectAttackTarget(pLeader);

            if (newVictim && (newVictim != pVictim))
            {
                if (pVictim)
                    me->AttackStop();
                else
                    AttackStart(newVictim);

                return;
            }
        }
    }
``` 

I think it's better to make this priority only for DPS. Otherwise, the logic above, the tank protects the group members, does not work.

I have tested two different scenarios.
The first is when the tank hits the target. The second is when the tank ignores this logic. 
In the second case, the tank protected the group members better.

GoonBoiBarry:
> ```c++
>     // Swap DPS to marked target or party leader's target
>     if (m_role == ROLE_MELEE_DPS || m_role == ROLE_RANGE_DPS)
>     {
>         Unit* pVictim = me->GetVictim();
> 
>         if (Player* pLeader = GetPartyLeader())
>         {
>             Unit* newVictim = SelectAttackTarget(pLeader);
> 
>             if (newVictim && (newVictim != pVictim))
>             {
>                 if (pVictim)
>                     me->AttackStop();
>                 else
>                     AttackStart(newVictim);
> 
>                 return;
>             }
>         }
>     }
> ```
> 
> I think it's better to make this priority only for DPS. Otherwise, the logic above, the tank protects the group members, does not work.
> 
> I have tested two different scenarios. The first is when the tank hits the target. The second is when the tank ignores this logic. In the second case, the tank protected the group members better.

Okay, but what if you're in a raid with multiple tanks and need one tank to stay on the boss no matter what? Would it not make more sense to leave the option to select which tanks behave this way via targeting with the command? I don't understand, have you tried just not setting a focus mark for the tank?

gajet5:
ok, I understood, yes, in this scenario, the tank will break off the label, but then this is a bug within the logic of the tank, why does it ignore the label if it is set for him.

GoonBoiBarry:
Well isn't that what your proposed change is doing in effect? Making tanks ignore focusmarks?

gajet5:
Yes, this is true, but only partially. The logic that was proposed completely disables the usual behavior of the tank, which is spelled out above. If the player selects one enemy. I'm not talking about the label, just the right mouse click.

I don't have enough experience to finish writing the logic for the tank. But after line 934 there should be a check for the presence of a label.
https://github.com/vmangos/core/blob/99c26c662be516976ad72f605deb8944f845ffa3/src/game/PlayerBots/PartyBotAI.cpp#L934

Then if there is an assigned tag for this AI, it should only attack it. Otherwise, try to protect the group.

GoonBoiBarry:
> Yes, this is true, but only partially. The logic that was proposed completely disables the usual behavior of the tank, which is spelled out above. If the player selects one enemy. I'm not talking about the label, just the right mouse click.
> 
> I don't have enough experience to finish writing the logic for the tank. But after line 934 there should be a check for the presence of a label.
> 
> https://github.com/vmangos/core/blob/99c26c662be516976ad72f605deb8944f845ffa3/src/game/PlayerBots/PartyBotAI.cpp#L934
> 
> Then if there is an assigned tag for this AI, it should only attack it. Otherwise, try to protect the group.

Ah I see, thanks for explaining it to me. I agree with you then.

gajet5:
You are welcome. I tried to write a solution. It's not the prettiest, but it works.
https://github.com/vmangos/core/compare/development...gajet5:core:fix-attack-marked-and-target

First, we are changing the attack priority only for DPS.
Secondly, we add a check for the tank, if there is a mark, then we attack only it.

GoonBoiBarry:
Oh cool, nice job! I'll give it a test sometime when I come back to vmangos

PR #2427, opened 2024-01-11
The announcement of the addition of a bg bot is configured

gajet5:
## 🍰 Pullrequest
Flood protect
This option is more like a debag.

I'm sorry about the last pull request. I'm just learning how to do everything well, but alas, it doesn't work out right away :(


mserajnik:
Personally, I would rather have it off by default, but that's for ratkosrb to decide.

gajet5:
Yes, I would like that too, but I did it the way it is now "by default".

Issue #1856, opened 2023-03-30
🚀 [Feature] Equipment don't break at partybot.


gajet5:
## 🚀 Feature
Make a setting, thanks to which, bots will not break equipment at death.

### Benefits, Purpose
This will improve the experience of using bots when going through dungeons.

PR #2433, opened 2024-01-13
PB repair items command

gajet5:
https://github.com/vmangos/core/issues/1856


balakethelock:
https://github.com/vmangos/core/blob/4264cded0dc90b1fed5370ffcf30c88451072cf7/src/game/Commands/CharacterCommands.cpp#L5670
Does this command not work on partybots? If not it may be cleaner to fix this one so it works on them, rather than implementing a new command.

ratkosrb:
This is redundant, there already is a command to repair items. And you should not have commands be available to players by default.

PR #2441, opened 2024-01-20
Anubisath Defender EventAI adjustements

balakethelock:
## 🍰 Pullrequest

- Removes the aggro and social pull Events.
- Adjusted spell timers based on sniff data.
- Simplified the script to only use 4 phases
- Implemented dropping Obsidian Shard on death
- Adjusted the range of Shadow Storm spells to 25 yards. The tooltip says 25 yards, and I haven't seen any evidence from classic discords, raid guides, wiki, videos to suggest that it should be less than that.

### Proof

- Proof that aggro & social pull events are not correct to classic:
https://www.youtube.com/watch?v=eWPnt_Lersc&t=1018s
https://www.youtube.com/watch?v=RF1hE_YFx4g&t=901s

### Issues

- None

### How2Test

- none

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20240120155120_world.sql
+++ b/sql/migrations/20240120155120_world.sql
@@ -1,0 +1,88 @@
+DROP PROCEDURE IF EXISTS add_migration;
+delimiter ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20240120155120');
+IF v=0 THEN
+INSERT INTO `migrations` VALUES ('20240120155120');
+-- Add your query below.
+
+-- Removing unused script actions.
+DELETE FROM `creature_ai_scripts` WHERE `id` IN (1527701, 1527702, 1527703, 1527704, 1527705, 1527706, 1527707, 1527708, 1527709, 1527710, 1527711, 1527712, 1527713, 1527714, 1527715, 1527716, 1527717, 1527718, 1527719, 1527720, 1527721, 1527722, 1527723, 1527724, 1527725, 1527726, 1527727, 1527728, 1527729, 1527730, 1527731, 1527732, 1527733, 1527734, 1527735, 1527736, 1527737, 1527738, 1527739, 1527740, 1527741, 1527742, 1527743, 1527744, 1527745, 1527746, 1527747, 1527748, 1527749, 1527750, 1527751, 1527752, 1527753, 1527754, 1527755, 1527756, 1527757, 1527758, 1527759, 1527760, 1527761, 1527762, 1527763, 1527764, 1527765, 1527786, 1527787, 1527767, 1527768, 1527769, 1527770, 1527771, 1527772);
+
+-- Scripts list for Anubisath Defender
+DELETE FROM `creature_ai_scripts` WHERE `id`=1527701;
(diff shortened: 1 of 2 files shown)


sql/migrations/20240120155120_world.sql
```diff
@@ -0,0 +1,88 @@
...
+(1527701, 0, 0, 46, 1, 4, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Anubisath Defender - Random Phase between 1 and 4'),
+(1527701, 0, 0, 49, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Anubisath Defender - Combat Pulse');
+
+DELETE FROM `creature_ai_scripts` WHERE `id`=1527702;
+INSERT INTO `creature_ai_scripts` (`id`, `delay`, `priority`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) VALUES
+(1527702, 0, 0, 15, 27630, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Anubisath Defender - Drop Obsidian');
+
+DELETE FROM `creature_ai_scripts` WHERE `id`=1527703;
+INSERT INTO `creature_ai_scripts` (`id`, `delay`, `priority`, `command`, `datalong`, `datalong2`, `datalong3`, `datalong4`, `target_param1`, `target_param2`, `target_type`, `data_flags`, `dataint`, `dataint2`, `dataint3`, `dataint4`, `x`, `y`, `z`, `o`, `condition_id`, `comments`) VALUES
+(1527703, 0, 0, 15, 26555, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 'Anubisath Defender - Cast Thunderclap (Phase 1+2)');
```
ratkosrb: I think you made a mistake here. This is the spell id of Shadow Storm yet comment says Thunderclap.

ratkosrb:
By the way, looking at the list of spell ids that the creature has been seen casting, not all seem to be used in your script. I've highlighted the ones missing from the script here.

Untitled

ratkosrb:
Spell 25699 seems to be triggered by 25698 which is cast in the script, so ignore that one.

balakethelock:
> By the way, looking at the list of spell ids that the creature has been seen casting, not all seem to be used in your script. I've highlighted the ones missing from the script here.

These spells are dummy effects, I was faced with the choice of either implementing them as blizzard probably did which would require hardcoding their effects in spelleffects.cpp, or just go with EventAI

I decided to apply [Vmangos's project guidelines](https://github.com/vmangos/core#:~:text=off%20by%20default.-,DB%20Scripting,-%3A%20Content%20should%20be) and just went with the simpler method

ratkosrb:
If the dummy spells are just for selecting spell to cast, they can still be handled in EventAI, by using EVENT_T_HIT_BY_SPELL (when we are hit by a spell) and EVENT_T_HIT_BY_AURA (when aura is applied to us) events. Or we can just have the script cast them at the appropriate time even if they are not handled, just to make it so the order of spell cast packets in vmangos is the same as on blizzard servers on the packet level. I think it would be nice to utilize them. If you don't want to do it, I can merge it as is, but it would be cooler if you did.

PR #2456, opened 2024-01-26
Do not add a mark if it already exists.

gajet5:
Now: the list may consist of the same mark.
Improvement: Checking the list. If there is a mark, it is not added again.


ratkosrb:
Please move the opening brackets on the next line. That is the code style we use in vmangos.

gajet5:
Done, a habit with js. (:

PR #2503, opened 2024-02-10
Mage double-check the aura.

gajet5:


ratkosrb:
This is a bad PR. You use pArcaneIntellect without checking it for null. You search for a buff target but don't use it, instead casting on self.

gajet5:
I saw an error, I probably should have just pointed it out for correction without closing the pr.
But, it's done, what's done.

PR #2505, opened 2024-02-10
Equipment don't break at partybot.

gajet5:
https://github.com/vmangos/core/issues/1856


ratkosrb:
This has no place being there. That function should check if bot should auto revive, not have other totally unrelated side effects.

Issue #2400, opened 2024-01-01
🐛 [Bug] Partybots will not reapply Arcane Brilliance to other players unless they have expired on themselves


GoonBoiBarry:
## 🐛 Bug report
If a player in a party dies and a partybot knows Arcane Brilliance, they will not recast it on them when they are revived because they only check if their own aura has expired to rebuff.

### Expected behavior
Bots should still check other party member's auras for arcane brilliance and reapply it if it is not active and they know the spell.

### Steps to reproduce

1. Summon a partybot that knows arcane brilliance.
2. Wait for it to cast buff, then remove it from yourself.
3. Watch them not reapply it.
4. ...
5. Profit

### Version & Environment

  - 1.12.1.5875

Commit Hash:  

OS Client: Win 10
OS Server: Win 10

### Crashlog

- None

ShiyoKozuki:
My party bots have the opposite issue, they use group wide buffs to buff a single person/bot who died and it makes them really mana inefficient.

GoonBoiBarry:
Which buffs are they using? I have only witnessed arcane brilliance work how I described and assumed it worked that way with others

ShiyoKozuki:
Prayer of fortitude
Prayer of spirit

GoonBoiBarry:
Yep, it appears this issue is limited to arcane brilliance.

PR #2510, opened 2024-02-13
Fix mage buff brilliance when necessary

gajet5:
Fixed a bug in the past pr. Of course, I meant checking the effect on the group and updating it if someone is missing.


mserajnik:
Since I implemented the fix in the same way in https://github.com/vmangos/core/pull/2590 (because I missed this PR), I can confirm this is working. This fixes #2400.

mserajnik:
Did you close this on purpose or by accident? Because this is a legitimate fix, I tested it.

gajet5:
> Did you close this on purpose or by accident? Because this is a legitimate fix, I tested it.

I moved from github to another control platform. All my pullrequests have been deleted :(. 
I should have left the repository on github, I didn't think, I'm sorry.

PR #2511, opened 2024-02-13
Wait for database availability in workflows.

mserajnik:
## 🍰 Pullrequest
This ensures that the MySQL database is available during the GitHub workflows before trying to access it. This (hopefully) prevents failing workflows like [this one](https://github.com/vmangos/core/actions/runs/7877314892/job/21493225639).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/.github/workflows/db_check.yml
+++ b/.github/workflows/db_check.yml
@@ -25,11 +25,12 @@ jobs:
         sudo docker pull mysql
         sudo docker run --name mysqldb -p 3306:3306 -e MYSQL_ROOT_PASSWORD=root -d mysql:5.6 --max_allowed_packet=128M
         sudo docker start mysqldb
+        until nc -z localhost 3306 ; do sleep 5 ; echo "Waiting for database availability..." ; done
 
-    
+
     - name: Checkout update sql
       uses: actions/checkout@v2
-      
+
     - name: Checkout base sql
       run: |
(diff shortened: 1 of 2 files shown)


mserajnik:
I've just had the issue occur in a workflow run in my fork so it seems like this isn't a fix after all (at least not a fully correct one).
I can only assume that checking the port from the host via nc might not be reliable as Docker maybe reports the port as open before it's mapped to the container's internal port; the database might still not be up and ready at that point.

I'll look into alternative methods; I think inspecting the container health might be a viable (and more appropriate) solution.
If I find something that seems to work, I'll make a new PR.

ArielOakMT:
Wouldn't the simplest way of checking for database availability be to connect to it?
`docker exec mysqldb sh -c 'exec mysql -u root -proot -e "exit"'`

imranh2:
If we are embracing github actions fully then the usual approach is via container healthchecks while also using service/auxiliary containers, see https://docs.github.com/en/actions/using-containerized-services/about-service-containers#creating-service-containers

Here's a postgress example we can steal and tweak https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers#running-jobs-in-containers

mysql healthcheck example: https://github.com/docker-library/mysql/issues/930#issuecomment-1463541157

PR #2524, opened 2024-02-26
CI: Update actions.

mserajnik:
## 🍰 Pullrequest
This updates the used actions to their latest versions. ~~This is especially important for the artifact actions, as older versions will [no longer work after June 30th 2024](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/).~~ __EDIT:__ I might have misread and the _removal_ only affects GitHub Connect for now; regardless, they are getting deprecated and will likely also be removed here (non-enterprise) at some point, so it's good to get the update out of the way.

In addition, I've replaced the `set-output` according to [this article](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) as that will also be removed at some point in the future.

The PR also gets rid of some of the Node.js deprecations notices (marvinpinto/action-automatic-releases will have to be updated to get rid of the rest of them, or we switch to a different action for that step).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/.github/workflows/db_check.yml
+++ b/.github/workflows/db_check.yml
@@ -29,7 +29,7 @@ jobs:
 
 
     - name: Checkout update sql
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
 
     - name: Checkout base sql
       run: |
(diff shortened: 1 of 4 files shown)


mserajnik:
Note: The Ubuntu builds here will fail until #2523 is merged.

PR #2527, opened 2024-02-26
[linux] use gnu readline for linux console cli

imranh2:
## 🍰 Pullrequest

Use GNU Readline to do CLI input on Linux. Lets you have a command history and edit commands inline. This also opens up tab completion that can be implemented at some point.

Windows users still get the old way of doing input via fgets reading stdin...

I moved some of the world stopping stuff out of the CLI files too and into the main function that creates and destroys it all.

### Proof

- None

### Issues

- None

### How2Test

- Compile and build on linux (you'll need the deb/ubuntu package `libreadline-dev`)
- Make sure `Console.Enable` is enabled in mangosd.conf and try some commands! Use the arrow keys on your keyboard :)

### Todo / Checklist

- [x] Test on windows (i'll need someone else to do the windows building testing please)

### DIFF
--- a/.github/workflows/vmangos.yml
+++ b/.github/workflows/vmangos.yml
@@ -57,7 +57,7 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
       run: |
         sudo apt-get -qq update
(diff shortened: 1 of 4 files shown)


imranh2:
Tested on windows, works as expected :)

PR #2538, opened 2024-03-05
Fix missing locale columns.

ellingtonsantos:
## 🍰 Pullrequest
Create missing locale columns.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


ratkosrb:
These tables have localized text from the dbc, and there is no dbc for those locales, so it's useless to add columns for them. Both esMX and ruRU do not actually exist in vanilla.

ellingtonsantos:
I did it to normalize all locale tables.

PR #2543, opened 2024-03-09
minor code cleanup in world.cpp

Gamemechanicwow:
### Proof

- None

### Issues

Minor code cleanup for world.cpp

- return type in World::GetQueuedSessionPos changed from int32 to uint32 - as the value is used as a uint32 both inside and outside the function
- deletes holder in World::BanAccount to prevent possible memoryleak
- no "else" after return
- use bool for boolean settings
- unify parameter name for UpdateRealmCharCount in .cpp and .h

### How2Test

- None

### Todo / Checklist

- [X] None


src/game/World.cpp
```diff
@@ -2561,10 +2561,13 @@ BanReturn World::BanAccount(BanMode mode, std::string nameOrIP, uint32 duration_
             break;
         }
         default:
+            delete holder;
+
             return BAN_SYNTAX_ERROR;
     }
 
     db->DelayQueryHolderUnsafe(&banHandler, &BanAccountHandler::HandleAccountSelectResult, holder);
+    delete holder;
```
ratkosrb: The function called right before this says "Delay" in the name. Are you sure it's not storing and using the holder later on?

ratkosrb:
You added a gigantic temporary file in your last commit.

PR #2544, opened 2024-03-09
Change format of guid string to append the full GUID in hex format

balakethelock:
## 🍰 Pullrequest

Title, commands & logs to show GUID show entry & counter but don't show the real GUID that is most useful.

### Proof

- None

### Issues

- None

### How2Test

- (I didn't test on objects that aren't units)

### Todo / Checklist

- [X] None
image

### DIFF
--- a/src/game/ObjectGuid.cpp
+++ b/src/game/ObjectGuid.cpp
@@ -25,6 +25,7 @@
 #include "ObjectMgr.h"
 
 #include <sstream>
+#include <iomanip>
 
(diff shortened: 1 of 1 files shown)


ratkosrb:
What is the full guid useful for exactly? It's the low guid portion that's useful, so you can see what guid the target has in the database.

balakethelock:
> What is the full guid useful for exactly? It's the low guid portion that's useful, so you can see what guid the target has in the database.

client reverse engineering shenanigans :')
Maybe its better if I add it as an independant chat command

PR #2552, opened 2024-03-16
Fix build Debian

ReyDonovan:
## 🍰 Pullrequest

fix https://github.com/vmangos/core/commit/38b42e1ad5bd023e936510e0b2819eb57b961955
tested Debian 10


ratkosrb:
I reverted that commit since it didn't fix the issue, so there's no need anymore.

PR #2565, opened 2024-03-29
Removed strict range check in dungeons

balakethelock:
## 🍰 Pullrequest

This range check is not only excessively strict and breaks with extrapolation, it's also redundant as GetRandomAttackPoint makes that check already before extrapolation happens.

- extrapolation happens [here](https://github.com/balakethelock/core/blob/7ad28135e076052501bf40831108383a115089ae/src/game/Objects/Unit.cpp#L2969)
- called by [this](https://github.com/balakethelock/core/blob/7ad28135e076052501bf40831108383a115089ae/src/game/Objects/Unit.cpp#L10249)
- that is in turn called by [this](https://github.com/balakethelock/core/blob/7ad28135e076052501bf40831108383a115089ae/src/game/Movement/TargetedMovementGenerator.cpp#L90). if it returned false, that means the target was unreachable regardless of extrapolation so making the check again is unnecessary.

Another thing to do is to do yet another extrapolation calculation for the dungeon-only range check but I think that's ultimately just spending server resources for no benefit.

### Proof

https://www.youtube.com/watch?v=j19wdNo2gXY

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Movement/TargetedMovementGenerator.cpp
+++ b/src/game/Movement/TargetedMovementGenerator.cpp
@@ -161,15 +161,6 @@ void TargetedMovementGeneratorMedium<T, D>::_setTargetLocation(T &owner)
     if (!m_bReachable && !!(pathType & PATHFIND_INCOMPLETE) && owner.HasUnitState(UNIT_STAT_ALLOW_INCOMPLETE_PATH))
         m_bReachable = true;
 
-    // Enforce stricter checking inside dungeons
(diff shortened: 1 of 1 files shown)


ratkosrb:
@Gamemechanicwow opinion?

Gamemechanicwow:
> @Gamemechanicwow opinion?

Seems good. It solves the issue I described 👍

PR #2572, opened 2024-04-06
Linux daemon issue

### LINKED ISSUE #2561: 🐛 [Bug] Vmangos has been increasing physical memory without an upper limit
## 🐛 Bug report
Vmangos has been continuously increasing the use of physical memory without limits
Even if I set it to run in the foreground or background, it will continue to increase the use of physical memory until the process crashes

The server is ubuntu22.04

imranh2:
## 🍰 Pullrequest

In an attempt to fix the memory leak & process going nuts when launched in `daemon` mode with the changes I made in https://github.com/vmangos/core/pull/2527, simply don't launch the CLI when running in this mode. Rationale is that you can't even use the CLI easily if launched like this.

There's also a change around paths and what your relative path is in `daemon` mode, making it *more consistent* with non daemon mode.

Also a small memory leak fix for the linux mangosd CLI.

### Proof

- None

### Issues

- fixes #2561 

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/mangosd/CliRunnable.cpp
+++ b/src/mangosd/CliRunnable.cpp
@@ -145,5 +145,8 @@ void CliRunnable::operator()()
             }
             sWorld.QueueCliCommand(new CliCommandHolder(0, SEC_CONSOLE, nullptr, command.c_str(), &utf8print, &commandFinished));
         }
+        #ifndef WIN32
+        free(command_str);
+        #endif
     }
 }
--- a/src/mangosd/Main.cpp
+++ b/src/mangosd/Main.cpp
@@ -216,7 +216,7 @@ extern int main(int argc, char **argv)
 
     // and run the 'Master'
     // TODO: Why do we need this 'Master'? Can't all of this be in the Main as for Realmd?
-    return sMaster.Run();
+    return sMaster.Run(serviceDaemonMode);
 
     // at sMaster return function exist with codes
     // 0 - normal shutdown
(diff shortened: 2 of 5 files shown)


src/mangosd/CliRunnable.cpp
```diff
@@ -144,6 +144,9 @@ void CliRunnable::operator()()
                 continue;
             }
             sWorld.QueueCliCommand(new CliCommandHolder(0, SEC_CONSOLE, nullptr, command.c_str(), &utf8print, &commandFinished));
+        #ifndef WIN32
+        free(command_str);
```
ratkosrb, line 149: the indentation does not match the rest of the function

imranh2:
Rebased changes / fixed merge conflict.

src/mangosd/CliRunnable.cpp
imranh2: Indenting aligns? It's intentionally after the if, so it's free'd even if it's nullptr which is ok to do.

mserajnik: > It's intentionally after the if

Looks to be inside the `if (command_str != nullptr)` to me though.

imranh2: 🤦 I'm clealy an idiot, thanks for the spotting this both!

imranh2 (commented):
Rewrote commit to place the code in the right place

PR #2573, opened 2024-04-06
Fix #2561: close() instead of freopen() to stop readline leaks

### LINKED ISSUE #2561: 🐛 [Bug] Vmangos has been increasing physical memory without an upper limit

imranh2:
## 🍰 Pullrequest

Direct fix for #2561, use `close` instead of `freopen` to `/dev/null`

### Proof

- None

### Issues

- fixes #2561 

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/shared/PosixDaemon.cpp
+++ b/src/shared/PosixDaemon.cpp
@@ -55,8 +55,13 @@ void startDaemon(uint32_t timeout)
     signal(SIGTERM, daemonSignal);
     signal(SIGALRM, daemonSignal);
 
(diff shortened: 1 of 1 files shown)


imranh2:
Force push was to undo the `chdir` fix I accidentally included.

PR #2575, opened 2024-04-07
Limit maximum number of file descriptors of mangosd/realmd processes

orygg:
## 🍰 Pullrequest

Both mangosd and realmd use (in certain configurations) ACE_Dev_Poll_Reactor from the ACE library, which allocated memory for every potential open file descriptor. Depending on the operating systems setting for maximum file descriptors per process this can lead to gigabytes of storage allocated.

We restrict the number of maximum file descriptors of our processes to at most 1024 (ACE handles adjusting the limit for the process), as this is a common setting that should be sufficient.

### Proof

You can observe this behaviour by changing the file descriptor limit with `ulimit -n`.

### Issues

This is related to my issue described in the comments of https://github.com/vmangos/core/issues/2561, which eventually turned out to be unrelated to the originally reported problem.

### How2Test

Run `ulimit -n 1073741816` (number taken from the environment I encountered the problem in), start mangosd/realmd and observe gigabytes of memory being allocated (before this patch is applied).

### Todo / Checklist

- [X] None

### DIFF
--- a/src/framework/Network/MangosSocketMgrImpl.h
+++ b/src/framework/Network/MangosSocketMgrImpl.h
@@ -17,6 +17,7 @@
 
 #include <set>
 #include <atomic>
+#include <algorithm>
 
 #include "Log.h"
 #include "Common.h"
@@ -76,7 +77,7 @@ public:
 
 #if defined (ACE_HAS_EVENT_POLL) || defined (ACE_HAS_DEV_POLL)
 
-        imp = new ACE_Dev_Poll_Reactor();
+        imp = new ACE_Dev_Poll_Reactor(std::min(1024, ACE::max_handles()));
 
         imp->max_notify_iterations(128);
         imp->restart(1);
--- a/src/mangosd/Master.cpp
+++ b/src/mangosd/Master.cpp
@@ -53,6 +53,8 @@
 #include <ace/Dev_Poll_Reactor.h>
 #include <signal.h>
 
+#include <algorithm>
+
 #ifdef WIN32
 #include "ServiceWin32.h"
 extern int m_ServiceStatus;
@@ -100,7 +102,7 @@ void remoteAccess()
 {
     #if defined (ACE_HAS_EVENT_POLL) || defined (ACE_HAS_DEV_POLL)
 
-    ACE_Dev_Poll_Reactor imp;
+    ACE_Dev_Poll_Reactor imp (std::min(1024, ACE::max_handles()));
 
     imp.max_notify_iterations (128);
     imp.restart (1);
--- a/src/realmd/Main.cpp
+++ b/src/realmd/Main.cpp
@@ -37,6 +37,8 @@
 #include <openssl/opensslv.h>
 #include <openssl/crypto.h>
 
+#include <algorithm>
+
 #include <ace/Get_Opt.h>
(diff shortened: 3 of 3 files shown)


ratkosrb:
But wont this limit the amount of players that can connect?

ghost:
Project Hyjal is calling, they want their bug back

orygg:
I thought about this as well and think you are correct (every open connection has its own fd), but the default for the maximum seems to be 1024 on Linux in general (I looked at a few VMs with different distributions running). I also did not find anything in the code that suggests that mangosd itself changes the limit (looked for `setrlimit` from POSIX and `set_handle_limit` from ACE).

Maybe we could have a look how some 'production' instances are configured (via `ulimit -n`) and if necessary increase the number or even make it configurable.

orygg:
@Nolin7777 comment made me look at other mangos descendants

Cmangos implementation of the code change a lot. The most significant comment probably being the following 8 year old switch from ACE to boost:

https://github.com/cmangos/mangos-classic/commit/bd56d3e61476bddd414dc4f27eb31a472366a96b

It would probably be better to port these changes to vmangos.

shudza:
The limit of 1024 makes no sense, the problem is absurd amount set by some systems (eg 1073741816). It should be something like 32k or 64k, to account for number of online players some of the biggest servers have (afaik 15k) and possible queued players.

mserajnik:
> Run `ulimit -n 1073741816` (number taken from the environment I encountered the problem in)

Just to reference where that particular value is coming from, in case anyone is wondering: https://github.com/containerd/containerd/pull/7566#issuecomment-1285417325

To me it sounds like the actual culprit is bad design in ACE (which, to be fair, also originally comes from a time where such high values where probably unthinkable).

For Docker containers in particular, lowering the maximum number of open files per process is [easy to do](https://docs.docker.com/compose/compose-file/compose-file-v3/#ulimits) if this is the only actually documented case where it's an issue right now.

I agree with orygg though that switching from ACE to boost or finding another way that doesn't require fiddling around with `ulimit` sounds like the best long-term solution.

orygg:
I suggest we do not merge this (so I close the request). I looked into porting the changes from cmangos, but that’s not an afternoon project :)

ratkosrb:
We are not adding boost or any other huge library dependency. If someone wants to undertake the herculean task of removing ACE, it must be replaced with native code for both linux and windows behind #ifdef checks. To replace one huge framework with another is pointless.

imranh2:
Oringally I was going to suggest we tie it to https://github.com/vmangos/core/blob/development/src/mangosd/mangosd.conf.dist.in#L118

But as we can set 0 there some sort of default needs to be read in, which should be the fd limit from the system which is current behaviour anyway...

orygg:
> We are not adding boost or any other huge library dependency. If someone wants to undertake the herculean task of removing ACE, it must be replaced with native code for both linux and windows behind #ifdef checks. To replace one huge framework with another is pointless.

I would argue that switching to a for more broadly used framework is not pointless.

And I would argue that, given limited resources, it would make sense to try to share as much code with other (more active) projects (cmangos) in order to benefit from the work they are doing and maybe even contributing back.

Though, as I have no time to do that myself at the moment, I leave both sentences in the conjunctive (not going to argue).

PR #2464, opened 2024-01-27
Improvement of the PB addition handler.

gajet5:
Added the ability to specify the second argument of the function. Allows you to specify a class and a role at once.
https://github.com/vmangos/core/issues/2362

PR #2588, opened 2024-04-18
Add partybot pull command.

### LINKED ISSUE #1794: 🚀 [Feature] New command for .partybot
## 🚀 Feature
Hi!

Please consider adding commands for tanks. For example, it would look like this - .partybot pull. 

This command works only for tank - .partybot pull [target boss, mob]
This command only works on the tank. All other bots (healers and dps) do nothing until they receive the command .partybot attackstart.

### Benefits, Purpose

This command will allow the tanks to better aggro mobs or boss, because when you use the command .partybot attackstart attack begin all the bots and damage rdd ahead of the damage of the tank.

mserajnik:
## 🍰 Pullrequest
This adds a `.partybot pull` command which instructs all tanks in the group to attack a targeted mob and pauses all DPS in the group (for 10 seconds by default, but the duration can also be specified).

Usage (when targeting a mob to be pulled):

```sh
.partybot pull [duration] # duration in milliseconds that the DPS should be paused for; optional, default is 10000 milliseconds
```

The duration has to be specified in milliseconds so it's consistent with [the `.partybot pause` command](https://github.com/vmangos/core/blob/ae1a317374dbf020693ae396f135a42436ef8a3f/src/game/PlayerBots/PlayerBotMgr.cpp#L1551-L1629), but it's a quick change if seconds are preferable here (imo, seconds would be precise enough and simpler to use, but I didn't want to deviate from the established syntax).

### Proof

- None

### Issues

- Closes #1794

### How2Test

- `.partybot add tank`
- `.partybot add dps`
- `.partybot add healer`
- Target a mob
- `.partybot pull`
- Observe the tank attacking the target and the DPS getting paused for 10 seconds; the healer should not be paused and start healing the tank
- Target another mob after the first one dies
- `.partybot pull 5000`
- Observe the tank attacking the target and the DPS getting paused for 5 seconds (since we override the default 10 seconds); like before, the healer should not be paused

### Todo / Checklist

- [x] None

### DIFF
--- a/src/game/PlayerBots/PlayerBotMgr.cpp
+++ b/src/game/PlayerBots/PlayerBotMgr.cpp
@@ -340,7 +340,7 @@ void PlayerBotMgr::Update(uint32 diff)
                 uint32 const minLevel = bg->GetMinLevel() + 10 * bracketId;
                 ASSERT(minLevel <= PLAYER_MAX_LEVEL);
                 uint32 const maxLevel = std::min<uint32>(minLevel + 9, PLAYER_MAX_LEVEL);
-                
+
                 for (uint32 i = queuedAllianceCount[bracketId]; i < bg->GetMinPlayersPerTeam(); ++i)
                 {
                     uint32 const botLevel = urand(minLevel, maxLevel);
@@ -1076,7 +1076,7 @@ bool ChatHandler::HandlePartyBotAttackStartCommand(char* args)
         SetSentErrorMessage(true);
         return false;
     }
-    
+
     Group* pGroup = pPlayer->GetGroup();
     if (!pGroup)
     {
@@ -1099,10 +1099,10 @@ bool ChatHandler::HandlePartyBotAttackStartCommand(char* args)
                     if (pMember->IsValidAttackTarget(pTarget))
                         pAI->AttackStart(pTarget);
                 }
-            }            
+            }
         }
     }
-    
+
     PSendSysMessage("All party bots are now attacking %s.", pTarget->GetName());
     return true;
 }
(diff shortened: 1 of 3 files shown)


src/game/PlayerBots/PlayerBotMgr.cpp
```diff
@@ -1638,6 +1638,65 @@ bool ChatHandler::HandlePartyBotUnpauseCommand(char* args)
...
+    if (!pGroup)
+    {
+        SendSysMessage("You are not in a group.");
+        SetSentErrorMessage(true);
+        return false;
+    }
+
+    bool back = false;
+    uint32 duration = 0;
+    if (char* arg1 = ExtractArg(&args))
```
ratkosrb: Why not ExtractUInt32 ?

src/game/PlayerBots/PlayerBotMgr.cpp
```diff
@@ -1638,6 +1638,65 @@ bool ChatHandler::HandlePartyBotUnpauseCommand(char* args)
...
+
+    Group* pGroup = pPlayer->GetGroup();
+    if (!pGroup)
+    {
+        SendSysMessage("You are not in a group.");
+        SetSentErrorMessage(true);
+        return false;
+    }
+
+    bool back = false;
```
ratkosrb: Unused variable.

src/game/PlayerBots/PlayerBotMgr.cpp
mserajnik: Good suggestion, I didn't see this.

src/game/PlayerBots/PlayerBotMgr.cpp
mserajnik: Fixed, it's a remnant I forgot to remove from an attempt to implement a `back` parameter (see comments in #1794).

ShiyoKozuki:
Nice! Thanks for this.
Can you make the tank pull with ranged in a future commit?

mserajnik:
> Can you make the tank pull with ranged in a future commit?

Sounds interesting as an option; I'm not sure how hard it would be, but I might look into it at some point.

mickle75:
Can also make a class for (example) `.partybot add warrior` work with this?
Currently works with role only (example) `.partybot add tank`
Cheers.

mserajnik:
@mickle75 It works when you add a bot with a specific class too, you just have to make sure the bot has the tank role (because the command only works with that role, it wouldn’t make sense otherwise); target the added warrior (or any other bot you want to assign the tank role) and use `.partybot setrole tank`.

mickle75:
Thanks,  thought would ask maybe of more shortcut way.
Thank you.

mserajnik:
@mickle75 And see also https://github.com/vmangos/core/pull/2464, when that gets merged it will be possible to specify role and class in one command when adding a bot.

PR #2599, opened 2024-04-27
Partybot rework of the pause command, more flexible.

gajet5:
Added options for roles.
The approach to specifying time has been changed, not in milliseconds but in seconds. It's more convenient for a person.
The maximum pause time cannot be more than 5 hours.

Test:
.partybot pause 5 (default all)
.partybot pause 5 tank
.partybot pause 5 dps
.partybot pause 5 healer
.partybot pause 5 all

unpause to
.partybot unpause (default all)
.partybot unpause tank
.partybot unpause dps
.partybot unpause healer
.partybot unpause all

### DIFF
--- a/src/game/PlayerBots/PlayerBotMgr.cpp
+++ b/src/game/PlayerBots/PlayerBotMgr.cpp
@@ -1670,12 +1670,11 @@ bool ChatHandler::HandlePartyBotPullCommand(char* args)
     }
 
     uint32 duration;
-    char* arg = ExtractArg(&args);
 
-    if (isNumeric(arg))
-        duration = atoi(arg) * IN_MILLISECONDS; // in sec
-    else
-        duration = 10 * IN_MILLISECONDS;
+    if (!ExtractUInt32(&args, duration))
+        duration = 10;
+
+    duration = duration * IN_MILLISECONDS;
 
     for (GroupReference* itr = pGroup->GetFirstMember(); itr != nullptr; itr = itr->next())
     {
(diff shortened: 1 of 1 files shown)


mserajnik:
The syntax of the `.partybot pull` command that was recently added (https://github.com/vmangos/core/commit/b5d1c07474e5eec7ae1993f7c01e2eb44b318b58) should also be changed to use seconds instead of milliseconds so it's consistent with the `.partybot pause` syntax when this gets merged.

gajet5:
> The syntax of the `.partybot pull` command that was recently added ([b5d1c07](https://github.com/vmangos/core/commit/b5d1c07474e5eec7ae1993f7c01e2eb44b318b58)) should also be changed to use seconds instead of milliseconds so it's consistent with the `.partybot pause` syntax when this gets merged.

Yes, it will need to be done, the seconds are much more convenient and understandable.

src/game/PlayerBots/PlayerBotMgr.cpp
```diff
@@ -1658,7 +1670,11 @@ bool ChatHandler::HandlePartyBotPullCommand(char* args)
     }
 
     uint32 duration;
-    if (!ExtractUInt32(&args, duration))
+    char* arg = ExtractArg(&args);
+
+    if (isNumeric(arg))
+        duration = atoi(arg) * IN_MILLISECONDS; // in sec
+    else
         duration = 10 * IN_MILLISECONDS;
```
mserajnik: Why not just
```cpp
if (!ExtractUInt32(&args, duration))
    duration = 10

duration = duration * IN_MILLISECONDS;
```

gajet5: Copied and pasted, I didn't think :)

PR #2601, opened 2024-04-30
Fix nesting of free().

mserajnik:
## 🍰 Pullrequest
This fixes the GCC/Clang build error:
```
/home/runner/work/core/core/src/mangosd/CliRunnable.cpp: In member function ‘void CliRunnable::operator()()’:
/home/runner/work/core/core/src/mangosd/CliRunnable.cpp:128:14: error: ‘command_str’ was not declared in this scope
  128 |         free(command_str);
      |              ^~~~~~~~~~~
[ 99%] Building CXX object src/mangosd/CMakeFiles/mangosd.dir/MaNGOSsoap.cpp.o
make[2]: *** [src/mangosd/CMakeFiles/mangosd.dir/build.make:76: src/mangosd/CMakeFiles/mangosd.dir/CliRunnable.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:556: src/mangosd/CMakeFiles/mangosd.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

The new `if (s_canReadLine)` introduced in https://github.com/vmangos/core/commit/79c5f503df39cd92eccc399443f8502eb8f9c248 probably caused the auto-merge to fuck up the nesting.

The PR also makes it so `free()` is used on Windows as well; when #2572 was opened, readline support didn't exist on Windows, but now it does (so the `free()` needs to happen there as well).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/mangosd/CliRunnable.cpp
+++ b/src/mangosd/CliRunnable.cpp
@@ -123,9 +123,7 @@ void CliRunnable::operator()()
                 s_canReadLine = false;
                 sWorld.QueueCliCommand(new CliCommandHolder(0, SEC_CONSOLE, nullptr, command.c_str(), &utf8print, &commandFinished));
             }
+            free(command_str);
         }
-        #ifndef WIN32
-        free(command_str);
-        #endif
     }
 }


mserajnik:
But I just realized, since https://github.com/vmangos/core/commit/79c5f503df39cd92eccc399443f8502eb8f9c248 introduced readline support for Windows, the `free()` should probably happen on Windows as well?

imranh2:
Yep, we should do an unconditional free, just remove the preprocessor ifdefs

mserajnik:
> Yep, we should do an unconditional free, just remove the preprocessor ifdefs

Yeah, I didn't realize this at first because I was only focused on the wrong nesting. 😄 
I adjusted this PR now.

PR #2605, opened 2024-05-03
Prevent BattleBot addition messages from spamming all players in world.

GrenderG:
## 🍰 Pullrequest
At this moment the message of adding a BattleBot will spam all players in world; with this change only players who are queued or inside the relevant BG will get notified.

### DIFF
--- a/src/game/World.cpp
+++ b/src/game/World.cpp
@@ -2245,8 +2245,11 @@ void World::SendWorldText(int32 string_id, ...)
 }
 
 // Send a System Message to all players in the same battleground or queue (except self if mentioned)
-void World::SendWorldTextToBGAndQueue(int32 string_id, uint32 queuedPlayerLevel, BattleGroundQueueTypeId queueType, ...)
+void World::SendWorldTextToBGAndQueue(int32 string_id, uint32 queuedPlayerLevel, uint32 queueType, ...)
 {
+    BattleGroundTypeId bgTypeId = BattleGroundMgr::BgTemplateId(static_cast<BattleGroundQueueTypeId>(queueType));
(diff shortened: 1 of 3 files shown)


gajet5:
https://github.com/vmangos/core/pull/2450

GrenderG:
> #2450

While I agree with your approach, @ratkosrb doesn't want to completely disable it so here's the middle ground I found after speaking with him.

src/game/World.cpp
```diff
@@ -2244,6 +2244,45 @@ void World::SendWorldText(int32 string_id, ...)
...
+    MaNGOS::LocalizedPacketListDo<MaNGOS::WorldWorldTextBuilder> wt_do(wt_builder);
+    for (const auto& itr : m_sessions)
+    {
+        if (WorldSession* session = itr.second)
+        {
+            Player* player = session->GetPlayer();
+            if (player && player->IsInWorld())
+            {
+                // Always announce it to all GMs.
+                if (player->GetSession()->GetSecurity() > SEC_PLAYER)
```
ratkosrb: Actually you have the session above, dont need to call GetSession here, just use session->

src/game/World.h
```diff
@@ -43,6 +43,7 @@
 #include <memory>
 #include <unordered_map>
 #include <thread>
+#include <BattleGroundDefines.h>
```
ratkosrb: can you just use uint32 to avoid the include, sucks to have to recompile whole server from changing 1 file, we should minimize includes in .h files

src/game/World.cpp
GrenderG: Done.

src/game/World.h
GrenderG: Done.

PR #2640, opened 2024-06-06
[Not Vanilla][Classic] Modification - trade in raid loot for 2 hours.

gajet5:
## 🍰 Pullrequest
I was motivated to develop this modification by the time I spend to resolve problems with loot on my server.
This PR is not for the main development branch, but for those who want to repeat the positive experience of classic wow.

### Proof
- Players can independently transfer items received in the raid within two hours, without involving a game master.

### Issues
- I think the errors are present.

### How2Test
- Kill the boss after looting and try to trade it to another player.

### Todo / Checklist
- [X] Kill the boss and give the loot to the player who was in the raid. The result is positive.
- [X] Kill the boss and transfer loot to a player who was not in the raid. The result is negative.
- [X] 2 hours have passed, check the loot transfer. The result is negative.
- [X] Relog and loot transfer verification. The result is positive.
- [X] Relog 30-60 minutes and checking the loot transfer. The result is positive.
- [X] They killed the boss, put on clothes, try to pass them on. The result is negative.
- [X] The clothes were put on by relog, try to pass them on. The result is negative.
- [X] The bride was dressed, taken off, relog, try to pass it on. The result is negative.
- [X] The clothes were put away in the bank, try to redo them. The result is positive.
- [X] The clothes were put away in the bank, relog, try to transfer them. The result is positive.
- [X] Loot received OUTSIDE the raid is not transferred. The result is negative.
- [X] No Auction. The result is positive.
- [X] No mail. The result is positive.
- [X] If you put on a thing, it becomes personal.


Kittnz:
Why don't you make this toggleable this behind a config?

gajet5:
> Why don't you make this toggleable this behind a config?

Set the value to 0 in config to disable.

0blu:
I think having an official and opt-in implementation of this feature in the core is a neat feature.
I saw many servers implementing this on their own **or wanting this but not having the resources to implement it**.
Having it in the core makes it more accessible and maintainable.

I am not sure if regex is the right approach, I saw some implementations from some private cores without using regex.
I also think this feature should be disabled by default and inside the value description you can state a recommend value.

ratkosrb:
I am heavily against polluting the codebase with giant custom features. It will not be merged. But it can sit open here for visibility for people who are looking for it.

Issue #2647, opened 2024-06-07
💬 [Discussion] Sweeping strikes


Otari98:
Greetings, I've been playing warrior on many private servers and on classic. I've noticed that sweeping strikes work differently. On vmangos when you use whirlwind and SS it will consume 1 charge and generate 1 additional attack, but on classic the same move will result in 2 additional attacks. I've done tests on classic era servers:

https://www.youtube.com/watch?v=xEF8WA0Sc38
https://youtu.be/f_0aAQ7RFTA
https://youtu.be/lu75PgGddCs
https://youtu.be/Z8rNx0u5HHY
https://youtu.be/dvDXetdfCy8
https://youtu.be/y6ALTZoDCw4

Also on classic SS consumes charges even if there is no secondary target (on vmangos it doesn't) and doesn't proc from Rend (on vmangos it does and results in SS that deals 1 damage on secondary target which looks broken). In many tests i've done on classic server I've never seen SS getting dodged/parried/missed/blocked (it happens on vmangos with one of the dummy spells for SS, the one that is triggered when you use Execute and the secondary target is not below 20%).

Going back to Whirlwind.
Patch notes from 1.10: 
"Whirlwind- When this ability is used with Sweeping Strikes, it will burn only one charge of Sweeping Strikes and will generate only one additional attack." This video from vanilla confirms this (i assume patch 1.12 since the guy is wearing gear from naxx): 

https://youtu.be/oKQNJL5IL2s?t=99  1 (SS from WW)
https://youtu.be/DBqITL0wR_M?list=PLHYN6a2afxplMAmz4uNGaPHT6aP3K9NOQ&t=575 (1 SS from WW also)

Also there is this guide (i think from 2007?) confirms this behavior too:
https://hilbert.wordpress.com/sweeping-strikes-guide/ 

So how is it supposed to work? What is the refence in this situation? Thank you for your time =)

gl00m3r:
> So how is it supposed to work? What is the refence in this situation?

Well Classic WoW is something different than original Vanilla WoW was. The patch notes says themselves. The project based and focus on Vanilla data and sometimes use sniffed Classic WoW data which we can't get.

Nothing to discuss I guess.

Otari98:
> > So how is it supposed to work? What is the refence in this situation?
> 
> Well Classic WoW is something different than original Vanilla WoW was. The patch notes says themselves. The project based and focus on Vanilla data and sometimes use sniffed Classic WoW data which we can't get.
> 
[...]

PR #2650, opened 2024-06-09
Sweeping strikes like on classic

Otari98:
## 🍰 Pullrequest
Title ^

### Proof
Issues and videos from classic in this discussion (https://github.com/vmangos/core/issues/2647)

### How2Test
create warrior
.level 59
.cheat power on
.cheat cooldown on
.npc summon 17685  x4
.learn 12292
.learn 1680
.learn 11574
.learn 20569

### Todo / Checklist

-- Still not perfect. On classic fully dodged/parried whirlwind still procs SS. (doesnt make sense but thats how it is there)
-- changing this line in UnitAuraProcHandler

if (!target)
    return SPELL_AURA_PROC_FAILED;
    
to SPELL_AURA_PROC_OK enables stacks to burn while solo target. It might allow to get accidentaly pvp flagged, although i dont think it was preventing that from happening before anyway.


src/game/UnitAuraProcHandler.cpp
```diff
@@ -371,7 +371,11 @@ SpellProcEventTriggerCheck Unit::IsTriggeredAtSpellProcEvent(Unit* pVictim, Spel
                 return SPELL_PROC_TRIGGER_OK;
         }
 #endif
-
+        //  Never proc for Rend.	
+        if (spellProto->Id == 12292)
```
0blu, line 375: I think the indentation look wrong

PR #2651, opened 2024-06-09
More database connection changes

0blu:
## 🍰 Pullrequest

- Use `std::string` instand of `const char*` for `SqlConnection::Query`, although there are still many references to a raw char* left :/
- Rename config `MaxPingTime` to `Database.AliveCheckInternal` and make it an interval in seconds

### DIFF
--- a/src/mangosd/mangosd.conf.dist.in
+++ b/src/mangosd/mangosd.conf.dist.in
@@ -55,8 +55,9 @@ ConfVersion=2010100901
 #        Amount of async threads (with dedicated connection) which will be used for async SELECT, executes, and transactions.
 #        Default: 1 async worker
 #
-#    MaxPingTime
-#        Settings for maximum database-ping interval (minutes between pings)
+#    Database.AliveCheckInternal - in seconds
+#        The interval in seconds to check the reachability of the database
+#        Default: 600 (10 min)
 #
 #    WorldServerPort
 #        Port on which the server will listen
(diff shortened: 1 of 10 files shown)


Gamemechanicwow:
What is the point in using std::string - if we don't use std::string functionality? Looks like it's just gonna add overhead and bloated asm.

0blu:
I mean was copied here anyways previously 
https://github.com/vmangos/core/pull/2651/files#diff-679bb7c6e8ba802a2e1c878b99891cdc7a39c96d1900d1b3a9bf5f80c80fdf6dL203

and I wanted to get rid of this code

https://github.com/vmangos/core/pull/2651/files#diff-679bb7c6e8ba802a2e1c878b99891cdc7a39c96d1900d1b3a9bf5f80c80fdf6dL233-L241

But maybe I can also reduce this pr to only the `Database.AliveCheckInternal` change.

Gamemechanicwow:
> I mean was copied here anyways previously https://github.com/vmangos/core/pull/2651/files#diff-679bb7c6e8ba802a2e1c878b99891cdc7a39c96d1900d1b3a9bf5f80c80fdf6dL203

Aha, I missed that. But is that mangos_strdup() even needed? 🤔

0blu:
Yeah, otherwise you would try delete a `const char*` that is embedded in the binary `.rodata` section. (which will cause a segfault)

Gamemechanicwow:
> Yeah, otherwise you would try delete a `const char*` that is embedded in the binary `.rodata` section. (which will cause a segfault)

Aren't we already doing that fine? Its a `char const*` in the pair.
        typedef std::pair<char const*, QueryResult*> SqlResultPair;

0blu:
Yes, it worked fine but you had to manually use `mangos_strdup` **and** `delete` that way.

PR #2680, opened 2024-06-22
Native Branch | Use modern networking sockets for low CPU usage and high performant code

0blu:
## 🍰 Pullrequest
vMaNGOS may be transitioning to its own I/O implementation.
This pull request aims to the networking code of realmd with a more modern version, similar to those used in cMaNGOS and TrinityCore.
The networking backend is inspired by Boost::ASIO and Nginx.
The usage of it is inspired by cMaNGOS.

Using winsock2's async interface results in less CPU time and a more parallel and performant code.
Currently every event is handled in the main thread. That can be changed later.


0blu:
In future commits there will be a UNIX implementation with epoll and some namespace refactoring.
I could also use `io_urings` on UNIX, but there are still relatively modern and I don't want to implement two methods for UNIX.

ratkosrb:
Can you also implement the config option MaxSessionDuration? This is a time limit how long clients are allowed to remain connected to realmd. They should be disconnected after this time, to prevent idle connections.

https://github.com/vmangos/core/blob/7836306ddd6a376e13189c0516c72cf91cd9fd59/src/realmd/BufferedSocket.cpp#L52

PR #2685, opened 2024-06-23
Fix patch for several objects/NPCs.

Daribon:
## 🍰 Pullrequest

This PR fixes the patch for several objects/NPCs that should not be spawned in earlier versions of the game.
The PR also adds two objects that were previously missing from alliance/horde camp, used only pre patch 1.12.

### Proof

- The horde and alliance camps were added in 1.9.0 and then in patch 1.12.0 the terrain changed when the sandlol pvp event was added.

- Images from vanilla camps pre 1.12:
March 27, 2006:
WoWScrnShot_032706_000831
Unknown date pre 1.12:
BattleKnightskillAbomination

From vmangos after PR applied:
WoW_NxjF8de0cA
WoW_ExvXibchLA

Old images from from Cenarion Hold pre 1.9:
Wearing old t2 art, t2 was updated in 1.9 so this image is from 1.8 (Hallows End was added in 1.8 October 10 2005):

Wearing old t2 art, t2 was updated in 1.9 so this image is from 1.8 (Hallows End was added in 1.8 October 10 2005):

Unknown date, but from originally from same image folder as the rogue wearing old t2. (Hallows End was added in 1.8 October 10 2005):

Wearing old t2 art, t2 was updated in 1.9 so this image is from 1.8 (Hallows End was added in 1.8 October 10 2005):

Wearing old t2 art, t2 was updated in 1.9 so this image is from 1.8 (Hallows End was added in 1.8 October 10 2005):

Unknown date, but from originally from same image folder as the rogue wearing old t2.

Unknown date, but as the Cenarion Hold Infantry are mounted, I believe its from 1.8:
ThebattlefieldofCenarionHold
- September 30, 2005 (patch 1.8.2):
WoWScrnShot_093005_103242
Unknown date, but as the Cenarion Hold Infantry are mounted, I believe its from 1.8:
125 - QIDtM
This image is dated July 22, 2006 (patch 1.11.2) Why the guard by the ladder is not spawned I do not know. It's possible it was added in later patch than 1.9.:
WoWScrnShot_072206_192602
It's possible in this image that someone has bugged the NPC into a pillar, since there is a very faint C in the middle of the screen inside the pillar.

### Issues

- None

### How2Test

- Use 1.7.1 or set wowpatch to 5
- Try to .go creature or .go ob any of the guids

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20240623045909_world.sql
+++ b/sql/migrations/20240623045909_world.sql
@@ -1,0 +1,54 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20240623045909');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20240623045909');
+-- Add your query below.
+
+
+-- Bonfires at Blackwood GY shouldn't be spawned pre 1.11 naxx
+UPDATE `gameobject` SET `patch_min`=9 WHERE `guid`=45609;
+UPDATE `gameobject` SET `patch_min`=9 WHERE `guid`=45611;
+UPDATE `gameobject` SET `patch_min`=9 WHERE `guid`=45610;
+UPDATE `gameobject` SET `patch_min`=9 WHERE `guid`=45608;
+
+-- Objects in Horde camp in Silithus shouldn't be spawned pre 1.9
+UPDATE `gameobject` SET `patch_min`=7 WHERE `guid`=49377;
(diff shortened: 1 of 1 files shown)


Daribon:
Added two objects that were previously missing from both alliance and horde camp in Silithus, in pre patch 1.12.

Daribon:
Updated with removal of several Cenarion Hold guards pre 1.9.
I think I've missed some NPCs, since I don't have images in all angles of Cenarion Hold. Some more dated images would be nice since the guard by the ladder has me confused why it would not be spawned post 1.9.

Needs more research.

Daribon:
Updated according to 1.8.0 sniff. Ready for merge.

WoW_L7ocGZKlX0

PR #2690, opened 2024-06-23
handle negative time value in CharacterScreenIdleKick

shudza:
## 🍰 Pullrequest
Fixes idle time calculation in CharacterScreenIdleKick.

The values can be negative since m_idleTime is set after the sessionUpdateTime, because the logout packet is proccessed after the sessionUpdateTime is set.

### Proof
Proof that negative values are possible:
idletime

### Issues
If `CharacterScreenMaxIdleTime` is set, session will close upon logout and skip character screen.

### How2Test
1. set `CharacterScreenMaxIdleTime` to any value
2. Login
3. Logout
4. Disconnected from server instead of character screen

### DIFF
--- a/src/game/WorldSession.cpp
+++ b/src/game/WorldSession.cpp
@@ -1333,7 +1333,8 @@ bool WorldSession::CharacterScreenIdleKick(uint32 currTime)
     if (!maxIdle) // disabled
         return false;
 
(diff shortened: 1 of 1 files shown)


ratkosrb:
Can't we avoid temporary variables and casting to int64 by first checking that (currTime > m_idleTime) ?

shudza:
fixed, thanks

PR #2714, opened 2024-08-07
Add mangosd parameter "Rate.Reputation.Gain.Threshold"

ghost:
## 🍰 Pullrequest
This parameter can be used to apply the reputation gain rate (parameter "Rate.Reputation.Gain") only if a specific reputation value is reached. For example, if set to 9000 an increased reputation gain rate is only applied if the player's reputation is at least honored, so the introduction to a new faction (quests etc.) is not affected.

### Proof
- None

### Issues
- None

### How2Test
- Set "Rate.Reputation.Gain" to 5
- Use default (-42000): The increased reputation gain rate is always applied
- Use value 9000: The increased reputation gain rate is only applied after reaching honored

### Todo / Checklist
- [X] None


ghost:
Not needed.

PR #2740, opened 2024-08-29
Replace ACE with native IO implementation 

0blu:
## 🍰 Pullrequest
This pr will replace ACE with a self-made but highly performant native solution. 

The implementation was inspired by boost asio, meaning that everything is asynchronous and requires a callback.
Users have to guarantee the lifetime of the socket throughout the transaction (aka. use `std::shared_ptr<>`).

Since cMaNGOS and TrinityCore already switched to Boost and this native IO implementation is very similar to Boost, some similarities in the network code can be found and compared against.

Native system function are used. There are slightly different backend implementations on each OS:
- Windows `IOCP`
- Linux `epoll` (with the possibility to add `io_uring` support)
- macOS `kqueue`

### Want to test it? 🧪 
Just merge these two commits into your repo.
<https://github.com/0blu/vmangos-core/tree/native-squash-merge-2024-11-17>
As soon this `native` branched is squashed-merged with `development` you can easily rebase->drop->cherry-pick it with the new official commit.

### Todo / Checklist
- [x] Test on low pop server
- [x] Test on medium pop server (3k-6k)
- [ ] Update build guide


0blu (commented):
woops

mserajnik:
I'm maintaining a [Docker setup for VMaNGOS](https://github.com/mserajnik/vmangos-deploy) and am in the process of [making the necessary changes for this native IO implementation](https://github.com/mserajnik/vmangos-deploy/commit/433f4da5309da1890c06991506171ad7a6bbc69a).

With these changes, the Docker images [build without issues](https://github.com/mserajnik/vmangos-deploy/actions/runs/10929031447) (built off of https://github.com/0blu/vmangos-core/tree/native-squash-merge), but when starting realmd, I encounter an error:

```
ERROR: Could not parse externalAddress: test.example.com for realm "VMaNGOS" (id 1) will skip realm update.
ERROR: No valid realms specified.
```

I replaced the domain name with `test.example.com` here to not leak it, but the actual domain name is in the same scheme (with a subdomain) and has been used for running VMaNGOS without issues in the past years. mangosd starts just fine, by the way. I can also confirm that swiching back to Docker images built off of https://github.com/vmangos/core/commit/91ecfc7ce3cfa1b17b5ea2c5265f07de0d60ecb2 makes the error go away again.

~~I'll try debugging this more later today, I just figured I'd post it right away in case it's a bug; but could also well be an oversight on my end and I need to adjust more things.~~

__Edit:__ Looking at the native IO code, it looks like [`IO::Networking::IpAddress::TryParseFromString`](https://github.com/0blu/vmangos-core/blob/21b83d34156a8534a55f2ecdf9b183c47fdebe2c/src/shared/IO/Networking/IpAddress.cpp#L15-L66) (which is used for parsing the external address) can only parse IP addresses, not domain names; I am not sure if this change is intended, but currently, domain names set as external address work just fine in VMaNGOS and are useful for DDNS situations where the server doesn't have a fixed IP. Of course, maybe it is also unintentional that domain names currently do work, but it seems like a good feature to have.

---

~~In case it's useful for anyone else: you can try these Docker images by following the instructions in the [README](https://github.com/mserajnik/vmangos-deploy/blob/master/README.md), but use [`vmangos-server-testing:0blu-vmangos-core-native-squash-merge-5875`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing/275875758?tag=0blu-vmangos-core-native-squash-merge-5875) (or [`vmangos-server-testing:0blu-vmangos-core-native-squash-merge-5875-anticheat`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing/275876971?tag=0blu-vmangos-core-native-squash-merge-5875-anticheat)) and [`vmangos-database-testing:0blu-vmangos-core-native-squash-merge`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-database-testing/275879813?tag=0blu-vmangos-core-native-squash-merge) respectively instead of the regular/non-testing image names/tags.~~
With my issues resolved (see later messages), I've removed those testing images and instead I now provide images tagged [`vmangos-server-testing:native-io-5875`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing) (or [`vmangos-server-testing:native-io-5875-anticheat`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing)) and [`vmangos-database-testing:native-io`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-database-testing) that are build off of the [`native`](https://github.com/vmangos/core/tree/native) branch, if you want to try these changes by using my Docker setup. I will rebuilt the images when new commits are added.

Also note that in this case only server images for `1.12.1.5875` are available (since the workflow these images are built with only creates them for one specified version, for testing purposes); if anyone wants images for a different client version, let me know and I'll run a workflow for that (or you can also just fork my repository and run the workflow yourself).

0blu:
@mserajnik Many thanks for your super details report and bug hunting.
I readded the DNS resolve option. You example should now work.

mserajnik:
> @mserajnik Many thanks for your super details report and bug hunting. I readded the DNS resolve option. You example should now work.

Thank you, I built new images off of [this branch](https://github.com/0blu/vmangos-core/tree/realmlist-dns) and realmd now starts fine.

However, when trying to log in with an account from a machine in the same LAN as the machine the server is running on, I'm getting an error (and the client gets a "disconnected" message accordingly):

```
realmd-1    | Accepting connection from 
realmd-1    | [AuthChallenge] Account 'TEST' using IP '' successfully authenticated
mangosd-1   | Accepting connection from 192.168.50.150
mangosd-1   | WorldSocket::HandleAuthSession: Sent Auth Response (Account IP differs from realmd).
```

This first curious thing are the `Accepting connection from [blank]` and the `Account 'TEST' using IP '' successfully authenticated` log messages; I am not sure why it doesn't print any IP here, because mangosd manages to print the actual LAN IP of my machine (`192.168.50.150`) just fine in the second line.
~~__Edit:__ Thinking about it, I am also not quite certain how mangosd manages to see the actual LAN IP of my client because due to the Docker bridge network, afaik it should only see the IP of the bridge network interface for all requests coming in from outside; unless of course the client is explicitly telling the server its LAN IP address (I don't know enough about the client <> server architecture to tell if that could be the case or not).~~
__Edit 2:__ I talked to a friend about this and apparently, for connections from machines on the same LAN, Docker can choose to not perform full NAT in some cases to increase performance, which could be what is happening here (that allows mangosd to see the actual LAN IP).

But it looks like it tries to compare the client IPs that it sees in realmd and mangosd and this probably fails because realmd can't determine the correct (or any?) IP.

I am not sure if this could be somehow related to the Docker networking (my setup uses the default bridge network driver), but it hasn't been an issue in the past and it's also interesting that it appears like a realmd-only issue.

As an additional test, I also tried connecting via WAN (having my client hooked up to a different internet connection than the server), and here, I encounter a different error: the client prints a "Success!" message when logging in, but then it just stays there indefinitely (and doesn't seem to load the character list).

I can't actually see any mangosd output in this case, only realmd output: 

```
realmd-1    | Accepting connection from 
realmd-1    | [AuthChallenge] Account 'TEST' using IP '' successfully authenticated
```

Let me know if I can provide any more information that would help figuring this out.

---

~~For anyone wanting to test those new images, they are tagged [`vmangos-server-testing:0blu-vmangos-core-realmlist-dns-5875`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing/277796497?tag=0blu-vmangos-core-realmlist-dns-5875) (or [`vmangos-server-testing:0blu-vmangos-core-realmlist-dns-5875-anticheat`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing/277800493?tag=0blu-vmangos-core-realmlist-dns-5875-anticheat)) and [`vmangos-database-testing:0blu-vmangos-core-realmlist-dns`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-database-testing/277804922?tag=0blu-vmangos-core-realmlist-dns) respectively.~~
With my issues resolved (see later messages), I've removed those testing images and instead I now provide images tagged [`vmangos-server-testing:native-io-5875`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing) (or [`vmangos-server-testing:native-io-5875-anticheat`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-server-testing)) and [`vmangos-database-testing:native-io`](https://github.com/mserajnik/vmangos-deploy/pkgs/container/vmangos-database-testing) that are build off of the [`native`](https://github.com/vmangos/core/tree/native) branch, if you want to try these changes by using my Docker setup. I will rebuilt the images when new commits are added.

mserajnik:
@0blu

Just as an update:

With the latest commits from [here](https://github.com/0blu/vmangos-core/tree/native-realmd-no-init-ip) I can now successfully log in and play (ran around and killed/looted a few mobs) without any apparent issues when my client machine is on the same LAN as the Docker host machine:

```
realmd-1    | Accepting connection from 192.168.50.150
realmd-1    | [AuthChallenge] Account 'TEST' using IP '192.168.50.150' successfully authenticated
mangosd-1   | Accepting connection from 192.168.50.150
```

However, when I try to connect over WAN (with the client using a different internet connection than the server), I still get stuck with a "Success!" message in the client. realmd now logs correct the WAN IP too (I made the IP below up to not leak my real one), but no output from mangosd (like before):

```
realmd-1    | Accepting connection from 44.51.114.150
realmd-1    | [AuthChallenge] Account 'TEST' using IP '44.51.114.150' successfully authenticated
```

0blu:
@mserajnik do you have wireshark installed? Can you verify the world server ip that is sent to the client via `realmd`?
image

mserajnik:
@0blu 

Good suggestion, I checked and realmd is indeed sending the LAN IP of the world server to my client that is connecting over WAN:
image
So I assume that's where the error occurs.

In case this could matter in any way: I am on a macOS system here that's connected to the internet via a phone hotspot (to test connecting to the server over WAN). And the game client runs in a Windows 11 virtual machine through Parallels, using the default [shared networking mode](https://kb.parallels.com/4948#section1). The Wireshark sniff is from macOS. While I do normally not actually play like this, I have used this setup quite frequently to test things in VMaNGOS and haven't had network issues before.

mserajnik:
@0blu I think I've found the issue: I run a Pi-hole instance and it probably responds with the LAN IP when realmd does the DNS lookup for the domain I use for my home server. I'll check if that is actually the problem and get back to you. 😄

mserajnik:
> @0blu I think I've found the issue: I run a Pi-hole instance and it probably responds with the LAN IP when realmd does the DNS lookup for the domain I use for my home server. I'll check if that is actually the problem and get back to you. 😄

And indeed, I had a local DNS record set up in Pi-hole for my home server domain because I was debugging some network issue a while ago and apparently forgot to remove it. So whenever realmd made the DNS lookup it resolved to the LAN IP, which of course then got sent even over WAN clients.

I've removed the local DNS record and as expected, it works just fine now. I'm sorry for the confusion.

0blu:
Stuff that I've noticed (while merging it into another codebase) that need to be fixed:
- [x] Re-add `Log::OpenWorldLogFiles()` to `/src/mangosd/Main.cpp`
- [x] Log file for `LOG_NETWORK`  in `mangosd` and also config
- [ ] Fix windows service daemon load order <-> config read
- [x] `BarGoLink`, assert `int32::max`, not `uint32::max`
- [x] Errors.h, wrong comment at `PrintStacktraceAndThrow` we dont print to `std::cout`
- [x] Errors.h, wrong comment at `PrintStacktrace`, `sLog.Out()` (`Out` missing)
- [ ] `nonstd::expected` is not working on C++17 when using MSVC (`std::in_place` conflict)
- [x] missing `cstring` include for `std::memcmp` at `ProxyV2Reader.cpp`

mserajnik:
@0blu

One more (minor) thing I have noticed while testing this: for my Docker setup, I currently do simple [healthchecks using `nc`](https://github.com/mserajnik/vmangos-deploy/blob/433f4da5309da1890c06991506171ad7a6bbc69a/compose.yaml.example#L108-L113) for both mangosd and realmd. `nc -z` starts a TCP handshake and sends a `RST` after the `SYN-ACK` response; that's enough to determine if the port is in use (which is sufficient for the healthcheck).

However, I saw that with the native IO changes running the healthcheck now causes an error to get logged by mangosd:

```
mangosd-1   | Accepting connection from 127.0.0.1
mangosd-1   | WorldSocket::DoRecvIncomingData: IoError: SocketClosed
mangosd-1   | ERROR: Failed to send packet SocketClosed
```

Note that realmd doesn't print the same error (and note also that realmd print a time with that line while mangosd doesn't):

```
realmd-1    | 13:23:57 Accepting connection from 127.0.0.1
```

It's just a cosmetic issue, but causing an error log once per minute (whenever the healthcheck runs) isn't exactly ideal.

I suppose having a script that gracefully closes the connection instead of sending a `RST` would be the easiest solution here; or maybe something like a `ss -ltn | grep -c ':8085'` would be better to avoid opening a TCP connection for the healhthcheck altogether? Although that would only confirm that the port is bound and won't work correctly in the case that the server hangs...

mserajnik:
> However, I saw that with the native IO changes running the healthcheck now causes an error to get logged by mangosd

I tried solving this with a small [custom healthcheck program](https://github.com/mserajnik/vmangos-deploy/blob/19ca148a7618f31d7de0f19857d5c348dbd53ef2/docker/server/healthcheck.c) now that gracefully closes the connection instead of sending a `RST` like `nc -z` does. However, I still get the same error message when the healthcheck runs:

```
mangosd-1   | Accepting connection from 127.0.0.1
mangosd-1   | WorldSocket::DoRecvIncomingData: IoError: SocketClosed
mangosd-1   | ERROR: Failed to send packet SocketClosed
```

If this is intended, I guess I'll have to switch to using something like `ss -ltn | grep -c ':8085'` instead (as previously mentioned), but I'd prefer to continue being able to actually check if the server is responding (without logging an error).

mserajnik:
Also, unrelated to my healthcheck issue, it seems like mangosd is no longer shutting down correctly (realmd still does):

__Edit:__ Sorry, maybe that's because https://github.com/vmangos/core/commit/93bb983d546961e95d67982db174ed18595b672d didn't get merged into `native` (yet)?

```
realmd-1    | 13:38:13 Waiting for IO thread to finish
realmd-1    | 13:38:13 Halting process...
realmd-1 exited with code 0
mangosd-1 exited with code 137
```

Compare that to https://github.com/vmangos/core/commit/e56fce5e65e254a01af938747cebfc30d7d6ac1b:

```
mangosd-1   | Shutting down world...
realmd-1    | Halting process...
mangosd-1   | Stopping network threads...
mangosd-1   | Unloading all maps...
mangosd-1   | Cleaning character database...
mangosd-1   | Sending queued mail...
mangosd-1   | Closing database connections...
realmd-1 exited with code 0
mangosd-1   | Halting process...
mangosd-1   | [NETWORK] Stopping movement broadcaster...
mangosd-1 exited with code 0
```

0blu:
> Also, unrelated to my healthcheck issue, it seems like mangosd is no longer shutting down correctly (realmd still does):
> ```
> realmd-1    | 13:38:13 Waiting for IO thread to finish
> realmd-1    | 13:38:13 Halting process...
> realmd-1 exited with code 0
> mangosd-1 exited with code 137
> ```
Everything from `development` should be merged, even the commit that you mentioned.

How can I reproduce this?
I think I got the project running and I am now trying to update it.
If I use `ghcr.io/mserajnik/vmangos-server-testing:native-io-5875` as the base image I repletely get
(edit: forgot to refetch the new version of the database image)

It seems like when `mangosd` is running in the container the `SIGTERM` signal is ignored.
If I run `kill -SIGTERM $(pidof mangosd)` or `docker compose kill --signal=SIGTERM mangosd` nothing happens.

But if I run it on my local machine natively `kill -SIGTERM $(pidof mangosd)` works and shuts down gracefully.

mserajnik:
> (edit: forgot to refetch the new version of the database image)

Glad you got it running!

> It seems like when mangosd is running in the container the `SIGTERM` signal is ignored.

Yeah, it's strange to me because it works with images built off https://github.com/vmangos/core/commit/e56fce5e65e254a01af938747cebfc30d7d6ac1b and I never had any issues there in the past. I didn't change how mangosd is started in the container either for the native IO changes, it's just [`exec /opt/vmangos/bin/mangosd -c /opt/vmangos/config/mangosd.conf`](https://github.com/mserajnik/vmangos-deploy/blob/19ca148a7618f31d7de0f19857d5c348dbd53ef2/docker/server/docker-cmd-mangosd.sh#L21), which makes mangosd `PID 1` and the `SIGTERM` that's sent when you run `docker compose down` should be handled properly.

0blu:
Even if I log into the container with `docker compose exec mangosd bash` and run
```bash
vmangos@d4c55a0da091:/$ pidof mangosd
1
vmangos@d4c55a0da091:/$ kill -SIGTERM 1
vmangos@d4c55a0da091:/$ echo $?
0
```
nothing happens.
Are you compiling these images manually?

Can you try to add this and see if the signals are logged?
```diff
Subject: [PATCH] Make AsyncServer cpp file
diff --git a/src/mangosd/Master.cpp b/src/mangosd/Master.cpp
--- a/src/mangosd/Master.cpp	(revision 335d50816e6247e4d7d13a1c9f4b2e10f520d5c8)
+++ b/src/mangosd/Master.cpp	(date 1727962636192)
@@ -521,6 +521,8 @@
 
 void Master::_OnSignal(int s)
 {
+    printf("GOT SIGNAL %d\n", s);
+    fflush(stdout);
     switch (s)
     {
         case SIGINT:

```

mserajnik:
> Are you compiling these images manually?
> Can you try to add this and see if the signals are logged?

They are built through [GitHub Actions workflows](https://github.com/mserajnik/vmangos-deploy/tree/master/.github/workflows), but I can add this patch for those testing images.

I'll build new images and give them a test, I'll let you know.

mserajnik:
> I'll build new images and give them a test, I'll let you know.

Tried the new images and don't see the `GOT SIGNAL` log when running `docker compose down`, so it really seems like for some reason the `SIGTERM` never reaches this code now.

mserajnik:
@0blu

> Tried the new images and don't see the GOT SIGNAL log when running docker compose down, so it really seems like for some reason the SIGTERM never reaches this code now.

I did some more debugging and got thinking that maybe it has to do with the way I run mangosd in the container; as mentioned before, it's run via `exec /opt/vmangos/bin/mangosd -c /opt/vmangos/config/mangosd.conf` so it becomes `PID 1`. I also use the following two Docker (Compose) options:

```yaml
tty: true
stdin_open: true
```

Without `stdin_open: true`, the container boot loops, which I guess makes sense since I don't run mangosd as a daemon, it's effectively running in the foreground.
`tty: true` is there so [`docker attach`](https://github.com/mserajnik/vmangos-deploy?tab=readme-ov-file#creating-the-first-account) can be used to access the VMaNGOS console.

Are there changes in this PR that could cause the `SIGTERM` to be "ignored" when mangosd is run as described (in non-daemon mode)? Because that could also explain this:

> But if I run it on my local machine natively `kill -SIGTERM $(pidof mangosd)` works and shuts down gracefully.

At least if you ran it as a daemon instead during your testing.

0blu:
I have no idea what would have caused docker to swallow the signals. The ACE version is working, right?
The only thing that I've changes in this regard, was adding `::` to the signal functions.
https://github.com/vmangos/core/commit/28dd205796fb0d688386bf9a1b2af0d5a4ceb7ab#diff-d1e0ba997761b1785cc29db66b962b2fe0e111a81dd0d446ff7f6fe705f12104R482-R496

I have no idea how to debug this inside a docker container.
On my local machine the signals were forwarded to the application.
Mangos was not running as a daemon.

shudza:
@mserajnik  Are you sure this is not docker related? I may have had this happen to me before. It's best to try both branches on the same docker environment.

mserajnik:
> The ACE version is working, right?

Yeah, the ACE version is working just fine. Since there were also no actual changes to the Docker setup (besides removing the ACE dependency and updating the example config files) it led me to believe that it had to be a change introduced here.

> The only thing that I've changes in this regard, was adding `::` to the signal functions.

Yeah, I saw this, but that's just the explicit way to use the global `signal` functions, right? So behavior-wise that shouldn't have changed anything.

> On my local machine the signals were forwarded to the application.
> Mangos was not running as a daemon.

If that's the case and mangosd reacts to `SIGTERM` just fine when started foregrounded, then I guess it must be something else.

> I have no idea how to debug this inside a docker container.

I am not a C++ programmer, but I'll try looking into how to be able to track this down; worse cast, I guess with a ton of prints at all the places related to signal handling. 😄

Considering that realmd processes the signals just fine even with the native IO changes and is using the exact same Docker image I still believe that the issue must be somewhere inside the mangosd code (or related to the way it's launched inside the container).

Anyway, this Docker setup is a third party project and thus I don't expect you to invest time to fix this if it's working when using the "intended" ways to run VMaNGOS (not inside a container). So I'll attempt to fix this and let you know when I discover anything.

mserajnik:
@shudza 

> Are you sure this is not docker related

Quite sure; as mentioned in my previous comment, realmd behaves fine and is using the same Docker image. So does mangosd using Docker images built from the current, ACE-using codebase. And if you check [here](https://github.com/mserajnik/vmangos-deploy/commit/19ca148a7618f31d7de0f19857d5c348dbd53ef2) there were essentially no changes to the Docker setup, besides removing the ACE package and updating the example config.

I've been running this on several machines to test, even multiple times back-to-back to rule out that it's just flaky behavior in general ("flaky" as in: with it sometimes working and sometimes not, regardless of if it's the ACE-using codebase or the new native one).

mserajnik:
Just as an update, I added GDB to the images, and also an [additional print in `Master::_HookSignals()`](https://github.com/mserajnik/vmangos-deploy/commit/51e46397172c0e986bbee0c9e1758cd84e27fe0b#diff-2e614127584201be31334ba1e2faa7f59d4dc56ace7c7f5687d3e5fb302b0b5f), which can be seen when the server starts.

I've never used GDB before, but attaching to mangosd (`gdb -p 1`) and running `bt` results in the following output (this is from an `aarch64` image):
```gdb
(gdb) bt
#0  0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=128, cancel=true,
    abstime=0x0, op=265, expected=60, futex_word=0xffff4640e710)
    at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=128, abstime=0x0, clockid=0,
    expected=60, futex_word=0xffff4640e710) at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0xffff4640e710,
    expected=60, clockid=clockid@entry=0, abstime=abstime@entry=0x0,
    private=private@entry=128) at ./nptl/futex-internal.c:139
#3  0x0000ffffa44c7368 in __pthread_clockjoin_ex (threadid=281471860401728,
    thread_return=0x0, clockid=0, abstime=0x0, block=<optimized out>)
    at ./nptl/pthread_join_common.c:102
#4  0x0000ffffa46dfba0 in std::thread::join() ()
   from /lib/aarch64-linux-gnu/libstdc++.so.6
#5  0x0000aaaace34ed88 in Master::Run() ()
#6  0x0000aaaace3369bc in main ()
```

I also ran `info threads`:
```gdb
(gdb) info threads
  Id   Target Id                                        Frame
* 1    Thread 0xffffa5899020 (LWP 1) "Main"             0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=128, cancel=true, abstime=0x0, op=265, expected=60,
    futex_word=0xffff4640e710) at ./nptl/futex-internal.c:57
  2    Thread 0xffff4320e640 (LWP 65) "MapContinent[1]" 0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff485d3e1c) at ./nptl/futex-internal.c:57
  3    Thread 0xffff43c0e640 (LWP 64) "MapContinent[0]" 0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff485d3e1c) at ./nptl/futex-internal.c:57
  4    Thread 0xffff4500e640 (LWP 63) "WorldUpdate[0]"  0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff485d3f58) at ./nptl/futex-internal.c:57
  5    Thread 0xffff4460e640 (LWP 62) "SystemTimer"     0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0xffff4460dcf8, op=393, expected=0,
    futex_word=0xffff4b41dfa8) at ./nptl/futex-internal.c:57
  6    Thread 0xffff45a0e640 (LWP 61) "CLI"             0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=req@entry=0xffff45a0db08, rem=rem@entry=0x0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  7    Thread 0xffff4640e640 (LWP 60) "WorldRunnable"   0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff4640dcf8, rem=0xffff4640dcf8)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  8    Thread 0xffff46e0e640 (LWP 59) "AsyncPacket"     0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff46e0dd00, rem=0xffff46e0dd00)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  9    Thread 0xffff4780e640 (LWP 58) "CharDB"          0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff4780dd10, rem=0xffff4780dd10)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  10   Thread 0xffff4820e640 (LWP 57) "LfgUpdate"       0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff4820dcd0, rem=0xffff4820dcd0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  11   Thread 0xffff5aa0e640 (LWP 56) "MapVis[2]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefdedc) at ./nptl/futex-internal.c:57
  12   Thread 0xffff5b40e640 (LWP 55) "MapVis[1]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefdedc) at ./nptl/futex-internal.c:57
  13   Thread 0xffff5be0e640 (LWP 54) "MapVis[0]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefdedc) at ./nptl/futex-internal.c:57
  14   Thread 0xffff5c80e640 (LWP 53) "MapObj[2]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefe158) at ./nptl/futex-internal.c:57
  15   Thread 0xffff5d20e640 (LWP 52) "MapObj[1]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefe158) at ./nptl/futex-internal.c:57
  16   Thread 0xffff5dc0e640 (LWP 51) "MapObj[0]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff5fefe158) at ./nptl/futex-internal.c:57
  17   Thread 0xffff74c0e640 (LWP 50) "MapVis[2]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1c58) at ./nptl/futex-internal.c:57
  18   Thread 0xffff7560e640 (LWP 49) "MapVis[1]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1c58) at ./nptl/futex-internal.c:57
  19   Thread 0xffff7600e640 (LWP 48) "MapVis[0]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1c58) at ./nptl/futex-internal.c:57
  20   Thread 0xffff76a0e640 (LWP 47) "MapObj[2]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1ed8) at ./nptl/futex-internal.c:57
  21   Thread 0xffff7740e640 (LWP 46) "MapObj[1]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1ed8) at ./nptl/futex-internal.c:57
  22   Thread 0xffff77e0e640 (LWP 45) "MapObj[0]"       0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff7d6f1ed8) at ./nptl/futex-internal.c:57
  23   Thread 0xffff8960e640 (LWP 44) "mangosd"         0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff912e9758) at ./nptl/futex-internal.c:57
  24   Thread 0xffff8a00e640 (LWP 43) "MapManager[1]"   0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff912e99d8) at ./nptl/futex-internal.c:57
  25   Thread 0xffff8aa0e640 (LWP 42) "MapManager[0]"   0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffff912e99d8) at ./nptl/futex-internal.c:57
  26   Thread 0xffff9240e640 (LWP 41) "IO[0]"           0x0000ffffa452bd74 in __GI_epoll_pwait (epfd=30, events=0xffff9240cd68, maxevents=250, timeout=500, set=0x0)
    at ../sysdeps/unix/sysv/linux/epoll_pwait.c:40
  27   Thread 0xffff92e0e640 (LWP 40) "DB:logs"         0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff92e0dcf0, rem=0xffff92e0dcf0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  28   Thread 0xffff9380e640 (LWP 39) "SqlResult[5]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  29   Thread 0xffff9420e640 (LWP 38) "SqlResult[4]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  30   Thread 0xffff94c0e640 (LWP 37) "SqlResult[3]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  31   Thread 0xffff9560e640 (LWP 36) "SqlResult[2]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  32   Thread 0xffff9600e640 (LWP 35) "SqlResult[1]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  33   Thread 0xffff96a0e640 (LWP 34) "SqlResult[0]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca158) at ./nptl/futex-internal.c:57
  34   Thread 0xffff9740e640 (LWP 33) "DB:realmd"       0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff9740dcf0, rem=0xffff9740dcf0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  35   Thread 0xffff97e0e640 (LWP 32) "SqlResult[5]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  36   Thread 0xffff9880e640 (LWP 31) "SqlResult[4]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  37   Thread 0xffff9920e640 (LWP 30) "SqlResult[3]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  38   Thread 0xffff99c0e640 (LWP 29) "SqlResult[2]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  39   Thread 0xffff9a60e640 (LWP 28) "SqlResult[1]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  40   Thread 0xffff9b00e640 (LWP 27) "SqlResult[0]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43ca798) at ./nptl/futex-internal.c:57
  41   Thread 0xffff9ba0e640 (LWP 26) "DB:characters"   0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffff9ba0dcf0, rem=0xffff9ba0dcf0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  42   Thread 0xffff9c40e640 (LWP 25) "SqlResult[5]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  43   Thread 0xffff9ce0e640 (LWP 24) "SqlResult[4]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  44   Thread 0xffff9d80e640 (LWP 23) "SqlResult[3]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  45   Thread 0xffff9e20e640 (LWP 22) "SqlResult[2]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  46   Thread 0xffff9ec0e640 (LWP 21) "SqlResult[1]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  47   Thread 0xffff9f60e640 (LWP 20) "SqlResult[0]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cadd8) at ./nptl/futex-internal.c:57
  48   Thread 0xffffa000e640 (LWP 19) "DB:mangos"       0x0000ffffa44fbed0 in __GI___clock_nanosleep (clock_id=<optimized out>, clock_id@entry=0, flags=flags@entry=0,
    req=0xffffa000dcf0, rem=0xffffa000dcf0)
    at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
  49   Thread 0xffffa0a0e640 (LWP 18) "SqlResult[5]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
  50   Thread 0xffffa140e640 (LWP 17) "SqlResult[4]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
  51   Thread 0xffffa1e0e640 (LWP 16) "SqlResult[3]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
  52   Thread 0xffffa280e640 (LWP 15) "SqlResult[2]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
  53   Thread 0xffffa320e640 (LWP 14) "SqlResult[1]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
  54   Thread 0xffffa3c0e640 (LWP 13) "SqlResult[0]"    0x0000ffffa44c1ebc in __futex_abstimed_wait_common64 (private=0, cancel=true, abstime=0x0, op=393, expected=0,
    futex_word=0xffffa43cb418) at ./nptl/futex-internal.c:57
```

Feeding the GDB output into an LLM results in a [number of suggestions](https://gist.github.com/mserajnik/3e4f687a1f2d0fc1a65c2238754f834b).

I'll continue to look into this. I might also be able to convince a friend of mine with a lot more experience in C++ to help out.

shudza:
did you check what happens when you send signal with gdb? `(gdb) signal SIGINT`

mserajnik:
> did you check what happens when you send signal with gdb? `(gdb) signal SIGINT`

Thank you for the suggestion, I just did:

```gdb
(gdb) signal SIGINT
Continuing with signal SIGINT.
█
```

`(gdb) signal SIGINT` doesn't seem to finish/exit, it prints a new blank line with a cursor/block character and just sits there (I let it run for 5 minutes). Hitting Enter just prints new blank lines (as it would in a shell with a command that hasn't exited yet).
According to mangosd's stdout output it also seems to continue to run normally (I can see the logs about events and such).

When I then hit Ctrl+C:

```gdb
^C
Thread 1 "Main" received signal SIGINT, Interrupt.
```

At this point, mangosd continues to listen and respond to TCP packets (I confirmed that via a healthcheck that checks if something is listening on that port), but I no longer get anything logged to stdout.
__Edit:__ And I guess that's normal and I have to run `(gdb) continue` at this point.

In any case, mangosd doesn't shut down this way.

shudza:
ah my bad, you should try SIGTERM actually

mserajnik:
> ah my bad, you should try SIGTERM actually

Thanks; just tried, I can observe exactly the same behavior as described before for `(gdb) signal SIGINT`.

mserajnik:
@0blu 

Alright, I did a bunch of more testing in regards to the signal handling issue.
I set up a Ubuntu 24.04 LTS VM and tested it there, not using Docker.
I chose Ubuntu 24.04 LTS because that's also what I'm using for the Docker images, so I replicated the setup as closely as I could.

Running `kill { PID }` (which sends a `SIGTERM` and thus behaves the same way as Docker when it stops containers) to stop **realmd**, it shuts down just fine and I can see the appropriate log output relating to the shutdown (as expected):
realmd_kill

However, stopping **mangosd** in the same manner, it does stop, but the usual log output relating to the shutdown is missing:
mangosd_kill
The only output is the `Terminated`, which of course is from the Bash session; if I start via `exec ./mangosd` instead (and not `./mangosd`), the terminal window just closes (instead of printing `Terminated`) when the `kill` is issued, otherwise it's the same. There is also nothing in the `Server.log` relating to the shutdown, it seems like it just stops immediately without doing any of its shutdown tasks (there is no observable delay between running `kill` and the process stopping).

Just to be sure, I also tried with `kill -TERM` explicitly (just in case Ubuntu somehow makes `kill` not default to `SIGTERM`), but I can observe the same with that.

Finally, using `Ctrl-C` to stop the server directly also behaves the same (immediate shutdown, no log output).

This is the log output I am talking about which I can normally observe when mangosd is stopping:
```
Shutting down world...
Stopping network threads...
Unloading all maps...
Cleaning character database...
Sending queued mail...
Closing database connections...
Halting process...
[NETWORK] Stopping movement broadcaster...
```

Of course, I double-checked that the log level is set correctly to see these (but that should be the default configuration anyway).

So, while it does behave differently than in the container (since mangosd does stop), it still seems like something is wrong there if those log lines are missing and mangosd instantly terminates.

0blu:
image

mserajnik:
> image

Interesting; I am not quite sure what’s going on here. Some “Ubuntu 24.04 only” issue that manifests in different ways when in container context (signal apparently “ignored”) and during regular usage (signal not ignored, but shutdown tasks not run)? And only with this changed codebase and only with mangosd and not realmd? Super weird to me…

Anyway, back to more debugging for me tomorrow, I guess I’ll try on 22.04 LTS as well.

mserajnik:
@0blu 

I think I have finally figured out the issue after doing more gdb and strace debugging with the help of a friend. If `SOAP.enabled = 1` isn't set (SOAP being disabled is the default configuration), `Master::Run()` continues all the way down to [`_UnhookSignals()`](https://github.com/vmangos/core/blob/39a6cde2df92c7373c5ac097a88574642c9bc2a8/src/mangosd/Master.cpp#L315) during startup which unhooks the signal handlers.

---

Here is an `strace -e trace=signal ./mangosd -c /home/test/vmangos/etc/mangosd.conf` output when `SOAP.enabled = 1` is not set (there are a lot of `rt_sigprocmask` and some other, unrelated `rt_sigaction` calls during boot before this point here that I removed for readability):

```
rt_sigaction(SIGINT, {sa_handler=0xb0e3bcc2f064, sa_mask=[INT], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
rt_sigaction(SIGTERM, {sa_handler=0xb0e3bcc2f064, sa_mask=[TERM], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
rt_sigaction(SIGSEGV, {sa_handler=0xb0e3bcc2f064, sa_mask=[SEGV], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], [], 8)   = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], [], 8)   = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigprocmask(SIG_BLOCK, ~[], [], 8)   = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
rt_sigaction(SIGINT, {sa_handler=SIG_DFL, sa_mask=[INT], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[INT], sa_flags=SA_RESTART}, 8) = 0
rt_sigaction(SIGTERM, {sa_handler=SIG_DFL, sa_mask=[TERM], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[TERM], sa_flags=SA_RESTART}, 8) = 0
rt_sigaction(SIGSEGV, {sa_handler=SIG_DFL, sa_mask=[SEGV], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[SEGV], sa_flags=SA_RESTART}, 8) = 0
```

You can see the signal handlers first set (those are [these three ::signal() calls](https://github.com/vmangos/core/blob/39a6cde2df92c7373c5ac097a88574642c9bc2a8/src/mangosd/Master.cpp#L578-L580)):
```
rt_sigaction(SIGINT, {sa_handler=0xb0e3bcc2f064, sa_mask=[INT], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
rt_sigaction(SIGTERM, {sa_handler=0xb0e3bcc2f064, sa_mask=[TERM], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
rt_sigaction(SIGSEGV, {sa_handler=0xb0e3bcc2f064, sa_mask=[SEGV], sa_flags=SA_RESTART}, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=0}, 8) = 0
```

But then also immediately removed/set to the default handlers again (via [these calls](https://github.com/vmangos/core/blob/39a6cde2df92c7373c5ac097a88574642c9bc2a8/src/mangosd/Master.cpp#L596-L598)):
```strace
rt_sigaction(SIGINT, {sa_handler=SIG_DFL, sa_mask=[INT], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[INT], sa_flags=SA_RESTART}, 8) = 0
rt_sigaction(SIGTERM, {sa_handler=SIG_DFL, sa_mask=[TERM], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[TERM], sa_flags=SA_RESTART}, 8) = 0
rt_sigaction(SIGSEGV, {sa_handler=SIG_DFL, sa_mask=[SEGV], sa_flags=SA_RESTART}, {sa_handler=0xb0e3bcc2f064, sa_mask=[SEGV], sa_flags=SA_RESTART}, 8) = 0
```

---

__If I now set `SOAP.enabled = 1` the server shuts down fine and as expected in both the Docker container and also the Ubuntu 24.04 LTS VM I tested with:__

image
Screenshot 2024-10-16 at 15 00 28

---

I can see from the log output in your screenshot you also have `SOAP.enabled = 1` set, so if you disable it, you should be able to replicate the issue that mangosd instantly terminates and doesn't run shutdown tasks, because the signal never reaches the intended handler.

So I assume the `_UnhookSignals();` should be placed somewhere else to fix this issue.

---

__Edit:__

My friend suggests removing the `sWorldSocketMgr->Wait();` ~~likely~~ _maybe_ caused this:

image

---

__Edit 2:__

Setting `MaxCoreStuckTime = 1` (or higher, the default being `0`) probably has the same effect as `SOAP.enabled = 1` in this context because that would also introduce a blocking thread join before `_UnhookSignals()` is called (and thus prevent the immediate unhook during startup), but I didn't explicitly test this.

0blu:
@mserajnik Many thanks for your analysis. It seems it was really the error in this case.
It should be fixed in the latest version.

@mserajnik btw. you should extend the `stop_grace_period` in your docker compose script for `mangosd`. I think the default is `10 sec` which might be a bit too close for larger servers.

mserajnik:
> Many thanks for your analysis. It seems it was really the error in this case.
> It should be fixed in the latest version.

Looks good to me!

> you should extend the `stop_grace_period` in your docker compose script for `mangosd`. I think the default is `10 sec` which might be a bit too close for larger servers.

Thanks for the suggestion. I don't have experience with how long larger servers might take, what would you suggest is a good value? 30 seconds maybe?

0blu:
Its the grace period when the server is _"not"_ responding to `docker compose stop`.
(and with "not" I mean that docker compose cannot know if the process accepted the signal and is shutting down or not)
Docker compose will hard-kill the process after this time.
I would say `2m` (minutes) would be fine.

mserajnik:
> Its the grace period when the server is "not" responding to docker compose stop.

Yup, I know, just didn’t think about larger servers taking longer for the shutdown tasks, but it does of course make sense.

I’ll set it two 2 minutes then to be on the safe side and put a comment that the user might want to adjust it lower if it’s a low-population server (because if it takes more than a few seconds there the server probably hangs anyway, no need for the user to wait 2 minutes for the kill).

Thanks!

mserajnik:
> > However, I saw that with the native IO changes running the healthcheck now causes an error to get logged by mangosd
> 
> I tried solving this with a small [custom healthcheck program](https://github.com/mserajnik/vmangos-deploy/blob/19ca148a7618f31d7de0f19857d5c348dbd53ef2/docker/server/healthcheck.c) now that gracefully closes the connection instead of sending a `RST` like `nc -z` does. However, I still get the same error message when the healthcheck runs:
> 
> ```
> mangosd-1   | Accepting connection from 127.0.0.1
> mangosd-1   | WorldSocket::DoRecvIncomingData: IoError: SocketClosed
> mangosd-1   | ERROR: Failed to send packet SocketClosed
> ```
> 
> If this is intended, I guess I'll have to switch to using something like `ss -ltn | grep -c ':8085'` instead (as previously mentioned), but I'd prefer to continue being able to actually check if the server is responding (without logging an error).

@0blu I think this got a bit lost because of the signal issue; any suggestion here? After thinking about it more, I'm leaning towards just keeping the healthcheck as is after all (and not just checking if the port is bound, which a `netstat`/`ss` solution would be); of course, as mentioned before, ideally there would be a way to do it without causing an error log each time.

mserajnik:
@0blu 

Build fails under Ubuntu/GCC with https://github.com/vmangos/core/pull/2740/commits/8a78b312ae5ace1770023b4118d4c122f78539c5 when `ENABLE_MAILSENDER` is set:

```
/core/src/realmd/AuthSocket.cpp: In member function 'void AuthSocket::_HandleLogonProof__PostRecv(const std::shared_ptr<const sAuthLogonProof_C>&, const std::shared_ptr<const PINData>&)':
/core/src/realmd/AuthSocket.cpp:714:56: error: 'GetRemoteIpAddress' was not declared in this scope; did you mean 'GetRemoteIpString'?
  714 |                 mail->substitution("%originating_ip%", GetRemoteIpAddress());
      |                                                        ^~~~~~~~~~~~~~~~~~
      |                                                        GetRemoteIpString
```

~~Maybe it would be a good idea to enable all the optional features (like `ENABLE_MAILSENDER`) for the "test" workflows to catch stuff like this easier?~~ Saw that that's implemented now (👍) and that the workflow currently fails with the same error.

src/shared/IO/Networking/IpAddress.cpp
```diff
@@ -145,3 +145,24 @@ void IO::Networking::IpAddress::UpdateCachedString()
...
 }
+
+bool IO::Networking::operator==(IpAddress const& lhs, IpAddress const& rhs)
+{
+    if (lhs.GetType() != rhs.GetType())
+        return false;
+
+    switch (lhs.GetType())
+    {
+    case IpAddress::Type::IPv4:
```
ratkosrb: indentation

0blu:
Finally :P

PR #2742, opened 2024-08-29
Native move stacktrace to cpp file

0blu:
## 🍰 Pullrequest
- Replaced backward.hpp with cpptrace
- Moving non-performance critical part to a `.cpp` file, which helps against include pollution.


src/scripts/eastern_kingdoms/eastern_plaguelands/naxxramas/instance_naxxramas.cpp
```diff
@@ -751,7 +751,7 @@ bool instance_naxxramas::IsEncounterInProgress() const
 
 void instance_naxxramas::SetData(uint32 uiType, uint32 uiData)
 {
-    ASSERT(this)
+    ASSERT(this);
```
0blu, line 754: How can this ever be not true?

ratkosrb: If SetData is called on a nullptr ScriptedInstance pointer.

PR #2771, opened 2024-09-17
Native Branch: Add built-in support for proxy-v2 to get the real IP of a player

0blu:
## 🍰 Pullrequest
With this PR it is possible to retrieve the original IP of a client, even if the server is behind a reverse proxy like HAProxy.


0blu:
I would have cleaned up the commit history a bit more, but its going to be squashed anyways.

Issue #2783, opened 2024-09-27
🔧 [Support] SOAP and NodeJS


gajet5:
## 🔧 Support Request
Has anyone tried to implement this?

My test code.
```javascript
const soap = require('soap');
const url = 'http://localhost:7878/';

(async () => {
  soap.createClient(
    url,
    {
      wsdl_options: { method: 'POST' },
      overrideRootElement: {
        namespace: 'urn:MaNGOS'
      }
    },
    (err, client) => {
      if (err) {
        console.log(err);
        return;
      }
      client.executeCommand({ command: 'server info' }, (err, result) => {
        if (err) {
          console.log(err);
          return;
        }
        console.log(result);
      });
    }
  );
})();
``` 

In the console, I get this:
```
"C:\Program Files\nodejs\node.exe" D:\git\s-evil.ru\src\app\lib\soap.js
Error: Invalid WSDL URL: http://localhost:7878/

 Code: 500

 Response Body: <?xml version="1.0" encoding="UTF-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" xmlns:xsi="http://www.w3.org/1999/XMLSchema-instance" xmlns:xsd="http://www.w3.org/1999/XMLSchema" xmlns:ns1="urn:MaNGOS"><SOAP-ENV:Body><SOAP-ENV:Fault><faultcode>SOAP-ENV:Client</faultcode><faultstring>Data required for operation</faultstring></SOAP-ENV:Fault></SOAP-ENV:Body></SOAP-ENV:Envelope>
    at D:\git\s-evil.ru\node_modules\soap\lib\wsdl\index.js:1292:26
    at handleBody (D:\git\s-evil.ru\node_modules\soap\lib\http.js:178:17)
    at D:\git\s-evil.ru\node_modules\soap\lib\http.js:210:24
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Process finished with exit code 0
```

Some data is required for the operation. I don't understand which ones.

Config:
```
SOAP.Enabled = 1
SOAP.IP = 127.0.0.1
SOAP.Port = 7878
```

cpevors:
Are you sure the WSDL URL is correct? Typically it'll end with .wsdl. 

Have you tried navigating to your stipulated URL in your browser, to see if you get a valid WSDL response?

[...]

PR #2784, opened 2024-09-28
Add Node.js SOAP example.

mserajnik:
## 🍰 Pullrequest

Using Node.js to build a SOAP client for VMaNGOS is currently not very straightforward because most people would probably attempt to utilize the [`soap`](https://www.npmjs.com/package/soap) npm package first, which won't work in this case because we don't provide a WSDL file; see #2783.

This PR adds a Node.js SOAP example (in addition to the existing [PHP example](https://github.com/vmangos/core/blob/f6da813251cd1ecd08931cc1d928bacf33607ab7/contrib/soap/example.php)) to iilustrate how a HTTP client like [`fetch`](https://nodejs.org/dist/latest-v22.x/docs/api/globals.html#fetch) can instead be used to construct and send SOAP requests.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/contrib/soap/example.js
+++ b/contrib/soap/example.js
@@ -1,0 +1,82 @@
+/*!
+ * Copyright (C) 2024-present  VMaNGOS  https://github.com/vmangos
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+// Using https://www.npmjs.com/package/soap does not work because it expects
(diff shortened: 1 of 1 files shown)


0blu:
Since this is just an example, maybe we should use the built in `fetch` instead of axios.

mserajnik:
> Since this is just an example, maybe we should use the built in `fetch` instead of axios.

Good idea; I chose axios because it's very common, but fetch has been stable in Node.js since v21, so I guess there's no reason to not use that instead. I'll adjust the PR accordingly.

contrib/soap/example.js
```diff
@@ -0,0 +1,65 @@
...
+        'Authorization': 'Basic ' + Buffer.from(`${username}:${password}`).toString('base64')
+      },
+      body: soapRequest
+    })
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! Status: ${response.status}`)
+    }
+
+    const result = await response.text()
```
0blu: Maybe isolate the result, just the php library does.
```suggestion
    const rawResult = await response.text()
    const resultBeginIdx = rawResult.indexOf('<result>') + '<result>'.length
    const resultEndIdx = rawResult.lastIndexOf('</result>')
    const result = rawResult.slice(resultBeginIdx, resultEndIdx)
```
---
Although I said we dont need any library.
Maybe we can get rid all the HTML encoded entities (like `&#xD;`) with an npm library like [this](https://www.npmjs.com/package/entities).
Or we leave it up the user of this project to figure this out.

0blu: And maybe we should also ensure that the status code is `200` before doing some string manipulation.

mserajnik: I've made adjustments, let me know what you think.

0blu (approved):

PR #2799, opened 2024-10-14
forsaken skills

Gamemechanicwow:
## 🍰 Pullrequest

Allows spell [7054 forsaken-skills](https://www.wowhead.com/classic/spell=7054/forsaken-skills) to proc [7053 forsaken-skill-shadow](https://www.wowhead.com/classic/spell=7053/forsaken-skill-shadow)

### Proof

- From the comments following the code, it looks like 7053 was mistaken for 7052
- I have verified that all 15 spells exist in the clients from 1.2.4 to 1.12.1 - and spell 7052 exists in none of them

### Issues

A different approach of fixing this, could be by using PickRandomValue(). 

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -1457,6 +1457,9 @@ void Aura::TriggerSpell()
         {
             case 7054:
                 spellRandom = urand(0, 14) + 7038;
+                if (spellRandom == 7052)
+                    spellRandom = 7053;
+
                 target->CastSpell(target, spellRandom, true, nullptr, this);
                 return;
                 break;
--- a/src/game/Spells/SpellAuras.cpp
+++ b/src/game/Spells/SpellAuras.cpp
@@ -1115,8 +1115,6 @@ void Aura::TriggerSpell()
     uint32 auraId = auraSpellInfo->Id;
     Unit* target = GetTarget();
 
-    uint32 spellRandom;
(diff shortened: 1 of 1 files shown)


ratkosrb:
Why is there a log statement in this function?

Gamemechanicwow:
For no apparent reason. Let's get rid of it.

ratkosrb:
Same spell is used in Aura::TriggerSpell btw.

ratkosrb:
Can you move the variable declaration to the proper scope btw, no idea why its declared outside the switch case.

src/game/Spells/SpellAuras.cpp
```diff
@@ -1457,6 +1457,9 @@ void Aura::TriggerSpell()
         {
             case 7054:
                 spellRandom = urand(0, 14) + 7038;
+                if (spellRandom == 7052)
+                    spellRandom = 7053;
+
                 target->CastSpell(target, spellRandom, true, nullptr, this);
                 return;
                 break;
```
ratkosrb: no need for break after return

Gamemechanicwow: fixed

0blu:
Maybe adding a comment to those `if`s, so it the future someone still knows why this spellid is not valid and why we need to skip this number.

PR #2803, opened 2024-10-18
Improve cmake install path for extraction tools

schell244:
## 🍰 Pullrequest
Copy map extraction tools into `${BIN_DIR}/Extractors` folder, also copy `offmesh.txt` and `config.json` to the same location.
Added a more consistent naming scheme for the tools:

| Old Name         | New Name           |
|------------------|--------------------|
| `mapextractor`   | `MapExtractor`     |
| `MoveMapGen`     | `MoveMapGenerator` |
| `vmap_assembler` | `VMapAssembler`    |
| `vmapextractor`  | `VMapExtractor`    |

### How to test
- Compile with `USE_EXTRACTORS = 1`
- CMakePredefinedTargets -> INSTALL
- map extraction tools will be copied into Extractors folder


mserajnik:
The executable names (`MoveMapGen` and `MoveMapGen.exe`) will have do be adjusted in this script too then: https://github.com/vmangos/core/blob/development/contrib/mmap/mmap_extract.py

contrib/extractor/CMakeLists.txt
```diff
@@ -23,24 +23,27 @@ include_directories (
...
   endif(ZLIB_FOUND)
 else()
-  target_link_libraries (mapextractor zlib libmpq bz2)
+  target_link_libraries (MapExtractor zlib libmpq bz2)
 endif(UNIX)
 
-install(TARGETS mapextractor DESTINATION ${BIN_DIR})
+install(TARGETS MapExtractor DESTINATION ${BIN_DIR}/Extractors)
+if(WIN32 AND MSVC)
+  install(FILES ${CMAKE_SOURCE_DIR}/bin/Debug/MapExtractor.pdb DESTINATION ${BIN_DIR}/Extractors CONFIGURATIONS Debug)
```
0blu: `BIN_DIR` is correct (which is the install path) but you have to adjust the source, use `${CMAKE_BINARY_DIR}` for that.

contrib/mmap/CMakeLists.txt
```diff
@@ -59,4 +59,10 @@ target_link_libraries(MoveMapGen
                         framework
                         )
 
-install(TARGETS MoveMapGen DESTINATION ${BIN_DIR})
+install(TARGETS MoveMapGenerator DESTINATION ${BIN_DIR}/Extractors)
+install(FILES ${CMAKE_SOURCE_DIR}/contrib/mmap/offmesh.txt DESTINATION ${BIN_DIR}/Extractors)
+install(FILES ${CMAKE_SOURCE_DIR}/contrib/mmap/config.json DESTINATION ${BIN_DIR}/Extractors)
+
+if(WIN32 AND MSVC)
+  install(FILES ${CMAKE_SOURCE_DIR}/bin/Debug/MoveMapGenerator.pdb DESTINATION ${BIN_DIR}/Extractors CONFIGURATIONS Debug)
```
0blu: Same as above

0blu (changes requested):

0blu (approved):
lgtm! I would squash merge it.

mserajnik (approved):
Since the script needs to be placed in the same directory as the `MoveMapGenerator` executable to work anyway, this seems like a good addition to me so the users don't have to do that manually.

__Edit:__ Whoops, this was meant to get added as a comment to [this commit](https://github.com/vmangos/core/pull/2803/commits/06fbf182f3e32eb259718e4b71c2f707fa890cab), not sure why it shows as me approving _all_ the changes in the PR instead.

mserajnik:
Hm, how about also renaming `mmap_extract.py` to something more consistent? Maybe `GenerateMoveMaps.py`; or maybe better `generate_move_maps.py` if we want a different style to visually distinguish the script more from the compiled executables. Not sure what would be preferable here.

PR #2826, opened 2024-11-08
Fix rolling a random item suffix 

Gamemechanicwow:
## 🍰 Pullrequest

This PR fixes an issue where the total chance for an entry in item_enchantment_template is above 100.

The original GetItemEnchantMod() has two issues. 
If the totalchance is below 100 - sometimes a second roll will happen.
If the totalchance is above 100 - there is a risk that some suffixes will never be rolled.

This PR fixes it by summing the totalchance and using that sum for the roll.

Also added a log output for entries that are silently skipped during loading.

### Proof

- None

### Issues

- None

### How2Test

Before:
.debug loottable enchant 4566 1000000
46 items dropped after 1000000 attempts for item Sturdy Quarterstaff.

After:
.debug loottable enchant 4566 1000000
70 items dropped after 1000000 attempts for item Sturdy Quarterstaff.

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/ItemEnchantmentMgr.cpp
+++ b/src/game/ItemEnchantmentMgr.cpp
@@ -109,7 +109,7 @@ uint32 GetItemEnchantMod(uint32 entry)
     float chance = 0;
 
     EnchStoreList const& enchantList = tab->second;
-    for (const auto& ench_iter : enchantList)
+    for (auto const& ench_iter : enchantList)
     {
         chance += ench_iter.chance;
     }
--- a/src/game/ItemEnchantmentMgr.cpp
+++ b/src/game/ItemEnchantmentMgr.cpp
@@ -117,7 +117,7 @@ uint32 GetItemEnchantMod(uint32 entry)
     float const roll = rand_chance() * (chance / 100.f);
     chance = 0.f;
 
-    for (const auto& ench_iter : enchantList)
+    for (auto const& ench_iter : enchantList)
     {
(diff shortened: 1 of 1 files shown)


src/game/ItemEnchantmentMgr.cpp
```diff
@@ -99,26 +106,23 @@ uint32 GetItemEnchantMod(uint32 entry)
         return 0;
     }
 
-    double dRoll = rand_chance();
-    float fCount = 0;
+    float chance = 0;
 
     EnchStoreList const& enchantList = tab->second;
     for (const auto& ench_iter : enchantList)
```
0blu:
```suggestion
    for (auto const& ench_iter : enchantList)
```

src/game/ItemEnchantmentMgr.cpp
```diff
@@ -99,26 +106,23 @@ uint32 GetItemEnchantMod(uint32 entry)
...
+        chance += ench_iter.chance;
     }
 
-    //we could get here only if sum of all enchantment chances is lower than 100%
-    dRoll = (irand(0, (int)floor(fCount * 100) + 1)) / 100.0f;
-    fCount = 0;
+    float const roll = rand_chance() * (chance / 100.f);
+    chance = 0.f;
 
     for (const auto& ench_iter : enchantList)
```
0blu:
```suggestion
    for (auto const& ench_iter : enchantList)
```

PR #2559, opened 2024-03-20
Use concurrent.futures and improve user feedback in mmap_extract.py.

mserajnik:
## 🍰 Pullrequest
This is a continuation of #2541 and improves the code of [mmap_extract.py](https://github.com/vmangos/core/blob/2a615b20721c32eb320d5d71171b95054187c37c/contrib/mmap/mmap_extract.py) further. It uses [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html) for the subprocesses instead of the previous thread-based parallelism, which is more efficient than the thread-keepalive check that existed beforehand and also allows us to easily check whether the function raised any exception (i.e. the `MoveMapGen` executable returned a non-zero exit code), which is hard when subclassing `threading.Thread`. It also determines the platform-specific configuration for running the `MoveMapGen` executable only once and then caches it, instead of running the same logic in every new thread.

The output of the `MoveMapGen` executable is now swallowed and instead the user is presented with a simple overview that makes it easier to follow and see which maps have finished processing and if there were any errors. This should be more appropriate in non-debug scenarios where all the user cares about is the end result instead of seeing the progress of each individual tile. It looks like this:

```
Using 8 worker(s) for map processing
(1/43) Map #34: Processed successfully
(2/43) Map #35: Processed successfully
(3/43) Map #13: Processed successfully
(4/43) Map #29: Processed successfully
(5/43) Map #42: Processed successfully
(6/43) Map #44: Processed successfully
(7/43) Map #43: Processed successfully
(8/43) Map #33: Processed successfully
(9/43) Map #70: Processed successfully
(10/43) Map #48: Processed successfully
(11/43) Map #36: Processed successfully
(12/43) Map #30: Processed successfully
(13/43) Map #37: Processed successfully
(14/43) Map #90: Processed successfully
(15/43) Map #129: Processed successfully
(16/43) Map #109: Processed successfully
(17/43) Map #189: Processed successfully
(18/43) Map #249: Processed successfully
(19/43) Map #229: Processed successfully
(20/43) Map #209: Processed successfully
(21/43) Map #230: Processed successfully
(22/43) Map #289: Processed successfully
(23/43) Map #47: Processed successfully
(24/43) Map #349: Processed successfully
(25/43) Map #269: Processed successfully
(26/43) Map #309: Processed successfully
(27/43) Map #389: Processed successfully
(28/43) Map #409: Processed successfully
(29/43) Map #449: Processed successfully
(30/43) Map #450: Processed successfully
[...]

PR #2827, opened 2024-11-08
MoveMapGenerator: add proper multithreading support

schell244:
## 🍰 Pullrequest
Allow mmap extraction to run properly multi threaded.
No changes have been made to the extraction logic itself, some funtions where just moved into TileWorker to run asynchronous.
Depending on the cpu, mmaps can now be build in <1 hour.

Screenshot 2024-11-08 130912

### How2Test
- start mmap extraction as usual
- select number of threads to be used
- resulting mmaps are the same as without those changes

### Todo / Checklist
- [X] None

### DIFF
--- a/contrib/mmap/mmap_extract.py
+++ b/contrib/mmap/mmap_extract.py
@@ -115,7 +115,7 @@ def get_subprocess_config() -> SubprocessConfig:
 def run_move_map_gen(map_id: int) -> None:
     config = get_subprocess_config()
     subprocess.check_call(
-        (config.executable, str(map_id), "--silent --skipPromtCores", *sys.argv[2:]),
+        (config.executable, str(map_id), "--silent", "--skipPromtCores", *sys.argv[2:]),
         startupinfo=config.startupinfo,
         creationflags=config.creation_flags,
         stdout=subprocess.DEVNULL,
--- a/contrib/mmap/mmap_extract.py
+++ b/contrib/mmap/mmap_extract.py
@@ -115,7 +115,7 @@ def get_subprocess_config() -> SubprocessConfig:
 def run_move_map_gen(map_id: int) -> None:
     config = get_subprocess_config()
     subprocess.check_call(
-        (config.executable, str(map_id), "--silent", "--skipPromtCores", *sys.argv[2:]),
+        (config.executable, str(map_id), "--silent", "--threads", "1", *sys.argv[3:]),
         startupinfo=config.startupinfo,
         creationflags=config.creation_flags,
         stdout=subprocess.DEVNULL,
--- a/contrib/mmap/mmap_extract.py
+++ b/contrib/mmap/mmap_extract.py
@@ -115,7 +115,7 @@ def get_subprocess_config() -> SubprocessConfig:
 def run_move_map_gen(map_id: int) -> None:
     config = get_subprocess_config()
     subprocess.check_call(
-        (config.executable, str(map_id), "--silent", *sys.argv[1:]),
+        (config.executable, str(map_id), "--silent --skipPromtCores", *sys.argv[2:]),
         startupinfo=config.startupinfo,
         creationflags=config.creation_flags,
         stdout=subprocess.DEVNULL,
--- a/contrib/mmap/mmap_extract.py
+++ b/contrib/mmap/mmap_extract.py
@@ -115,7 +115,7 @@ def get_subprocess_config() -> SubprocessConfig:
 def run_move_map_gen(map_id: int) -> None:
     config = get_subprocess_config()
     subprocess.check_call(
-        (config.executable, str(map_id), "--silent", "--threads", "1", *sys.argv[3:]),
+        (config.executable, str(map_id), "--silent", "--threads", "1", *sys.argv[1:]),
         startupinfo=config.startupinfo,
         creationflags=config.creation_flags,
         stdout=subprocess.DEVNULL,
--- a/contrib/extractor/System.cpp
+++ b/contrib/extractor/System.cpp
@@ -84,7 +84,7 @@ enum Extract
 // Select data for extract
 int   CONF_extract = EXTRACT_MAP | EXTRACT_DBC | EXTRACT_CAMERA;
 // This option allow limit minimum height to some value (Allow save some memory)
-// see contrib/mmap/src/Tilebuilder.h, INVALID_MAP_LIQ_HEIGHT
+// see contrib/mmap/src/TerrainBuilder.h, INVALID_MAP_LIQ_HEIGHT
 bool  CONF_allow_height_limit = true;
 float CONF_use_minHeight = -500.0f;
 
--- a/contrib/mmap/CMakeLists.txt
+++ b/contrib/mmap/CMakeLists.txt
@@ -46,6 +46,7 @@ set(SOURCES
     ./src/MapBuilder.cpp
     ./src/TerrainBuilder.cpp
     ./src/VMapExtensions.cpp
+    ./src/TileWorker.cpp
 )
 
 add_executable(${EXECUTABLE_NAME} ${SOURCES})
--- a/contrib/mmap/src/MapBuilder.cpp
+++ b/contrib/mmap/src/MapBuilder.cpp
@@ -27,221 +27,17 @@
 
 using namespace VMAP;
 
-inline void calcTriNormal(const float* v0, const float* v1, const float* v2, float* norm)
-{
-    float e0[3], e1[3];
-    rcVsub(e0, v1, v0);
-    rcVsub(e1, v2, v0);
-    rcVcross(norm, e0, e1);
-    rcVnormalize(norm);
-}
-
-inline unsigned int nextPow2(unsigned int v)
-{
-        v--;
-        v |= v >> 1;
-        v |= v >> 2;
-        v |= v >> 4;
-        v |= v >> 8;
-        v |= v >> 16;
-        v++;
-        return v;
-}
-
-inline unsigned int ilog2(unsigned int v)
-{
-        unsigned int r;
-        unsigned int shift;
-        r = (v > 0xffff) << 4; v >>= r;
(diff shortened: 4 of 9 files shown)


mserajnik:
I guess when it’s run from [`mmap_extract.py`](https://github.com/vmangos/core/blob/development/contrib/mmap/mmap_extract.py) it shouldn’t prompt the user but instead just select 1 thread automatically (since mmap_extract.py already spawns multiple workers).

schell244:
@mserajnik can you maybe give it try and confirm that [mmap_extract.py](https://github.com/vmangos/core/blob/development/contrib/mmap/mmap_extract.py) now works as expected?

mserajnik:
@schell244 Currently not at home, but I’ll give it a try later today or tomorrow at the latest!

mserajnik:
I'm building Docker images off of this PR right now and will test it tomorrow morning.

In the meantime, because it came to my mind: one could make the argument to remove the `mmap_extract.py` script altogether if `MoveMapGenerator` now supports multi-threading itself. However, I would keep it for two reasons:

1. I think the output of `mmap_extract.py` (that I implemented via #2559) is a better overview for the average end user than the elaborate `MoveMapGenerator` output
2. Removing the script now would potentially break VMaNGOS-related projects that rely on it (e.g., I use it for my Docker setup; of course, I monitor VMaNGOS development closely and could easily adjust it, but not everyone might do that)

0blu:
Yeah, maybe instead of `--skipPromtCores` make it `--threads <num>`. So the python script can just forward the count.

mserajnik:
> Yeah, maybe instead of `--skipPromtCores` make it `--threads <num>`. So the python script can just forward the count.

The Python script would need to use `--threads 1` still then, because it determines the maps beforehand and then spawns a `MoveMapGenerator` process for each map, so it already implements multithreading on its own. Changing this would be possible, but that would make it impossible/hard to produce the custom output as it does now, so personally I wouldn't do that.

However, other than that I quite like the idea; so if you pass `--threads <num>` it just uses that number and skips the prompt. That would be a lot more flexible (especially for scripting) than `--skipPromtCores` (should probably be `Prompt`, by the way, there's a typo) because then you can also do `--threads $(nproc)` and such.

contrib/mmap/mmap_extract.py
```diff
@@ -115,7 +115,7 @@ def get_subprocess_config() -> SubprocessConfig:
 def run_move_map_gen(map_id: int) -> None:
     config = get_subprocess_config()
     subprocess.check_call(
-        (config.executable, str(map_id), "--silent", *sys.argv[1:]),
+        (config.executable, str(map_id), "--silent", "--threads", "1", *sys.argv[3:]),
```
mserajnik: `*sys.argv[1:]` has to stay as it is; this has nothing to do with the added `--threads 1`, it just passes along the arguments that the Python script itself was invoked with, e.g., when running it like this to select the configuration and offmesh:
```sh
/opt/vmangos/bin/Extractors/mmap_extract.py \
  --configInputPath /opt/vmangos/bin/Extractors/config.json \
  --offMeshInput /opt/vmangos/bin/Extractors/offmesh.txt
  ```

**Comment by mserajnik** [comment]
@schell244

First of all, I added a comment (see above) because you cut off the Python script arguments with your change (that should be fixed), but that's probably not the root cause of the following:

I just tried it, but unfortunately it's not working when running `mmap_extract.py` now, `MoveMapGenerator` doesn't seem to actually run. I don't have time to properly debug this today, but my guess is that even with the passed `--threads 1` it might be waiting for user input before continuing, because the last output I see from the Python script is [this print](https://github.com/vmangos/core/blob/3f1c5dafdbc5a74e3c94c2b2c34d06aaa933012b/contrib/mmap/mmap_extract.py#L142).

Can you confirm that `MoveMapGenerator` works if you pass `--threads 1` directly (without invoking it through the Python script)?

**Comment by schell244** [comment]
Thanks a lots for testing it! I'll adapt the Python script arguments accordingly. I just re-checked starting the process from windows cmd with `MoveMapGenerator.exe --threads 1` and it works for me:
```
D:\Downloads>MoveMapGenerator.exe --silent --threads 12 --configInputPath "D:\Downloads\config.json" --offMeshInput "D:\Downloads\offmesh.txt"
MMap Generator
====================================
offMeshInputPath = D:\Downloads\offmesh.txt
configInputPath = D:\Downloads\config.json
Discovering maps... found 43.
Discovering tiles... found 2429.

Building map 000:
[Map 000] Creating navMesh [maxTiles=687]
[Map 000] We have 687 tiles.
...
```

**Comment by mserajnik** [comment]
> I just re-checked starting the process from windows cmd with `MoveMapGenerator.exe --threads 1` and it works for me

Thanks for confirming, I'll retest this again then (tomorrow probably) with the `*sys.argv[1:]` fix and see if I can't figure it out.

**Comment by mserajnik** [comment]
@schell244 

I've tried again, with the same result. I can see the `mmap_extract.py` script spawning the `MoveMapGenerator` "workers" (8, in my case):
image
But then nothing happens (and you can see, they basically don't use any CPU or memory); as if they were waiting for user input, which of course shouldn't happen with `--threads 1`.

---

The script effectively spawns each `MoveMapGenerator` worker like this (`34` being an example map ID here):
```
MoveMapGenerator 34 --silent --threads 1 --configInputPath /path/to/config.json --offMeshInput /path/to/offmesh.txt
```
Can you see anything wrong with that in regards to your changes?

---

I'm a bit short on time at the moment, but tomorrow I'll hopefully get around to try not swallowing the output from `MoveMapGenerator` in the Python script (it's getting swallowed for cosmetic reasons, the script generates its own "friendlier" output, as mentioned before) so I can maybe see better what's going on here.

**Comment by mserajnik** [comment]
@schell244 

> I'll hopefully get around to try not swallowing the output from `MoveMapGenerator` in the Python script

I was finally able to try that now and can see the following output:

```
MMap Generator
MMap Generator
MMap Generator
====================================
====================================
====================================
MMap Generator
====================================
MMap Generator
====================================
MMap Generator
====================================
MMap Generator
====================================
MMap Generator
====================================
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
configInputPath = /opt/vmangos/bin/Extractors/config.json
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
configInputPath = /opt/vmangos/bin/Extractors/config.json
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
```
So each of the 8 `MoveMapGenerator` "workers" the Python script starts simulatenously logs:
```
MMap Generator
====================================
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
```
(On a sidenote, it probably shouldn't log anything with the `--silent` flag provided)

If I remove the `--silent` I can actually see that the maps are being processed, e.g.,:
```
MMap Generator
====================================
offMeshInputPath = /opt/vmangos/bin/Extractors/offmesh.txt
configInputPath = /opt/vmangos/bin/Extractors/config.json
Discovering maps... found 43.
Discovering tiles... found 2429.

Building map 034:                                    
[Map 034] Creating navMesh [maxTiles=4]
[Map 034] We have 4 tiles.                          
[Map 034] Complete!    
```
~~__But `MoveMapGenerator` doesn't exit afterwards__, which is likely what's causing the issue here as the Python script is waiting for `MoveMapGenerator` to terminate. I confirmed it also doesn't exit when running it manually (so not spawned by the Python script).~~

~~__Edit 1:__ Nevermind, I think I forgot it should still be processing at this point (so that's why it doesn't exit); but it seems to hang with no resource usage after these log lines (also when running it manually).~~

__Edit 2:__ Actually, looking at the code, it seems like it should be done processing here after all and should exit; so maybe my first statement was right after all. I'm just confusing myself at this point, so I'll leave it at that. 😆
I'm also not sure if it actually processes the map for me because if I try it with map 1 and `--threads 1` it prints `[Map 001] Complete!` basically immediately, when before this used to take like an hour or so for map 1 on the machine I tested this.

So I'm not sure why this is working for you but not for me; it has always worked fine before these changes.
Let me know if you want me to try something specific to debug this.

schell244:
Thanks a lot for the testing efforts! I am currently also a bit time-constrained, but I will take care of it in the next days and will try to figure out whats causing problems.

schell244:
Hi @mserajnik,
I've indeed missed some changes in order to build single maps. Also I've turned off some logs, when --silent option is used.
(It's also possible now to build a single map with more than 1 thread, like when you build a big map like 0 or 1)

mserajnik:
@schell244 

Thanks for the fixes, it seems to work fine now:

image

---

> It's also possible now to build a single map with more than 1 thread, like when you build a big map like 0 or 1

I think that's great for rebuilding specific maps! 👍
Let's leave it at `--threads 1` for the `mmap_extract.py` script though since that already spawns one `MoveMapGenerator` process per logical CPU (which process one map each) and since we're already at 100% CPU utilization it's probably not beneficial for performance to use even more threads here.

schell244:
Cool, thanks a lot for your help getting this right!

PR #2832, opened 2024-11-17
vmap extractor - dont save struct padding to temp files

Gamemechanicwow:
## 🍰 Pullrequest

sizeof(WMOLiquidHeader) returns the size of the struct which is padded to 32 bytes.
WMOLiquidHeader only contains 30 bytes of data. So 2 random unused bytes are being stored in each temporary file.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/contrib/vmap_extractor/vmapextract/wmo.cpp
+++ b/contrib/vmap_extractor/vmapextract/wmo.cpp
@@ -249,7 +249,7 @@ bool WMOGroup::open()
         {
             liquflags |= 1;
             hlq = new WMOLiquidHeader;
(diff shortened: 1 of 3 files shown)


ratkosrb:
Is this a breaking change? Does is necessitate re-extraction of vmaps?

Gamemechanicwow:
nope. Nothing breaking. The 2 random bytes are written by vmap extractor to the temporary files in /buildings/ - and ignored by vmap assembler because it also used sizeof(WMOLiquidHeader).

PR #2833, opened 2024-11-17
Fix an issue were vmap extractor generates duplicate files on some systems

Gamemechanicwow:
## 🍰 Pullrequest

Vmap extractor extract modelnames from references in adt tiles and gameobjectdisplayinfo.dbc. The referenced modelnames can differ in case sensitivity. Since linux has case sensitive filnames, this will - unlike windows - not be caught by fileexists() and 1100~ duplicate files will be generated.

This PR ensures that each modelname is referenced the same way. 

- First letter is uppercase
- First letter after a non-alpha char is uppercase
- Rest is lowercase including the filetype

Also: Doodads will now be stored in the same order in the dir_bin file.

### Proof

- None

### Issues

- None

### How2Test

- Extract maps and vmaps on different platforms
- Diff check all files
- The only difference after this PR should be the floating points stored in dir_bin for position and rotation. At most the difference will be +-0.000001 

### Todo / Checklist

- [X] None

### DIFF
--- a/contrib/vmap_extractor/vmapextract/adtfile.cpp
+++ b/contrib/vmap_extractor/vmapextract/adtfile.cpp
@@ -148,8 +148,6 @@ bool ADTFile::init(uint32 map_num, uint32 tileX, uint32 tileY, StringSet& failed
                 while (p < buf + size)
                 {
                     fixnamen(p, strlen(p));
-                    char* s = GetPlainName(p);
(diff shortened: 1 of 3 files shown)


Gamemechanicwow:
marked as draft because even though - the extracted model files on linux now matches windows. I still need to analyze if it has any impacts on dir_bin and temp_gameobject_models.

PR #2834, opened 2024-11-17
Fix "Only allow one transaction at the same time"

0blu:
This reverts and fixes commit 5a5c7d1405f957e8f722d1196c272b8ee5bcaab2.

### DIFF
--- a/src/shared/Database/Database.cpp
+++ b/src/shared/Database/Database.cpp
@@ -459,9 +459,8 @@ bool Database::BeginTransaction(uint32 serialId)
 {
     if (!m_pAsyncConn)
         return false;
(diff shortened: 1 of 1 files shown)


src/shared/Database/Database.cpp
```diff
@@ -487,7 +486,9 @@ bool Database::CommitTransaction()
         return false;
 
     //check if we have pending transaction
-    ASSERT(m_TransStorage->get()); // if we will get a nested transaction request - we MUST fix code!!!
+    //ASSERT(m_TransStorage->get());
+    if (!m_TransStorage->get())
```
shudza, line 490: this condtion is already asserted

PR #2842, opened 2024-12-07
Some more code cleanup and refactoring

schell244:
## 🍰 Pullrequest
Code cleanup and some fixes based on Clang-Tidy suggestions.
Also resolved some compiler warnings.
No behavioural changes.

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Maps/MapPersistentStateMgr.cpp
+++ b/src/game/Maps/MapPersistentStateMgr.cpp
@@ -491,7 +491,7 @@ void DungeonResetScheduler::ScheduleAllDungeonResets()
     }
 }
 
-void DungeonResetScheduler::ScheduleReset(const bool add, const time_t time, DungeonResetEvent event)
+void DungeonResetScheduler::ScheduleReset(bool add, time_t const time, DungeonResetEvent event)
 {
     MapPersistentStateManager::PersistentStateMap::iterator itr = m_InstanceSaves.m_instanceSaveByInstanceId.find(event.instanceId);
     if (itr == m_InstanceSaves.m_instanceSaveByInstanceId.end())
--- a/src/game/Maps/MapPersistentStateMgr.cpp
+++ b/src/game/Maps/MapPersistentStateMgr.cpp
@@ -491,7 +491,7 @@ void DungeonResetScheduler::ScheduleAllDungeonResets()
     }
 }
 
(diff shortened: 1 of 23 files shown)


src/game/World.h
```diff
@@ -633,8 +633,8 @@ enum RealmType
 
 class SessionPacketSendTask
 {
-    SessionPacketSendTask(const SessionPacketSendTask&) = delete;
 public:
+    SessionPacketSendTask(const SessionPacketSendTask&) = delete;
```
ratkosrb: east const

src/game/Maps/MapPersistentStateMgr.cpp
```diff
@@ -491,7 +491,7 @@ void DungeonResetScheduler::ScheduleAllDungeonResets()
     }
 }
 
-void DungeonResetScheduler::ScheduleReset(bool add, time_t time, DungeonResetEvent event)
+void DungeonResetScheduler::ScheduleReset(bool add, time_t const time, DungeonResetEvent event)
```
ratkosrb: this is not a pointer so no need for the time parameter to be const

src/game/Maps/MapPersistentStateMgr.h
```diff
@@ -300,7 +300,7 @@ class DungeonResetScheduler
             m_resetTimeByMapId[mapId] = t;
         }
 
-        void ScheduleReset(bool add, time_t time, DungeonResetEvent event);
+        void ScheduleReset(bool add, time_t const time, DungeonResetEvent event);
```
ratkosrb: no need for time to be const

src/game/Objects/Creature.cpp
```diff
@@ -215,7 +215,7 @@ Creature::Creature(CreatureSubtype subtype) :
 
 Creature::~Creature()
 {
-    CleanupsBeforeDelete();
+    Unit::CleanupsBeforeDelete();
```
ratkosrb, line 218: why explicitly calling unit version

schell244: Just for readability -  there's no specific implementation for CleanupsBeforeDelete at creature (while at player there is, which then also calls Unit::CleanupsBeforeDelete).

PR #2848, opened 2024-12-12
Skip MinRealmListDelay check on first connection

schell244:
## 🍰 Pullrequest
Fixes error `user %s IP %s is sending CMD_REALM_LIST too frequently` on first login due to value m_lastRealmListRequest not being set before.

### How2Test
- create a new account
- login
- check auth server log

### Todo / Checklist
- [X] None


ratkosrb:
Shouldn't this be a time_t not uint32?

0blu:
We could also use `std::chrono::steady_clock`

PR #2859, opened 2025-01-02
Add missing atiesh quest dialog

srive:
## 🍰 Pullrequest

Add missing atiesh quest dialog
### Proof

- Looking at WoWhead there should be two pieces of dialog for this

### Issues

- None

### How2Test

- Start the Atiesh questline and talk to Anachronos when you are supposed to turn in the cleansed staff. Pre-PR nothing is said in the quest dialog

### Todo / Checklist

- [X] None


ratkosrb:
Do not include the database name in the migration.

srive:
Removed the database name in the migration.

Issue #2871, opened 2025-01-15
🦂 [Bug][AQ] The player can enter AQ ignoring the phase.


gajet5:
## 🦂 Bug report - Dungeon
![1](https://github.com/user-attachments/assets/0bed4735-4ade-4d8f-81b8-4eb1657a3330)
![2](https://github.com/user-attachments/assets/a6143ee7-1d00-4b93-82a7-5a17cab5633b)

A strange raid behavior was noticed on our server.
I did not change the phase through the command.

### Steps to reproduce
1. Go to the raid, 
2. Enter the raid.
3. Profit.

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  

OS Client: Windows
OS Server: Ubunru
Last Update Code Base: 15.01.2025
Last Update world DB: 15.01.2025

### Crashlog

- None

Wall-core:
There is a specific event for the AQ door itself, and I believe the raid locks automatically when enabled (elsewise, just lock it out in map_template by setting level to 61 minimum until it's open).

gajet5:
> There is a specific event for the AQ door itself, and I believe the raid locks automatically when enabled (elsewise, just lock it out in map_template by setting level to 61 minimum until it's open).

I understand that there should be a special event, but no one did it, but raid is access.

and there is no such value in this table.

![image](https://github.com/user-attachments/assets/a85ededa-f8d4-44b3-950e-6db4006fafbe)

GrenderG:
If content patch is set to 10, it's assumed AQ is already open unless you manually tweak it.

gajet5:
> If content patch is set to 10, it's assumed AQ is already open unless you manually tweak it.

![image](https://github.com/user-attachments/assets/68c7aa74-58f5-4071-b004-647d20053ceb)

The value is set to 0. I repeat, I did not change anything on my own, we had a plan to do this and check it.

GrenderG:
> > If content patch is set to 10, it's assumed AQ is already open unless you manually tweak it.
> 
[...]

PR #2884, opened 2025-02-02
Update game event end dates

schell244:
## 🍰 Pullrequest
Set all game event end years to 2037, fix some expired ones.
Kept the other date values (month, day, etc.).

### Issues
- https://github.com/vmangos/core/issues/2871

### Todo / Checklist
- [X] None


mserajnik:
Could columns using `TIMESTAMP` not be changed to `DATETIME` to avoid the 2038 limit?
I know this project wants to keep support for MySQL 5.6, but from what I can find `DATETIME` should work there as well?

Unrelated: I've had zero issues with MariaDB (11.4 LTS currently), so maybe that could be a potential default to target at some point in the future.

PR #2925, opened 2025-02-23
Set BattleBots temporary flag to true by default

ghost:
## 🍰 Pullrequest
Currently, manual added battlebots remain in battlegrounds after players leave because the temporary flag isn't consistently set. This change makes bots leave battlegrounds with their spawning player by default, which is more intuitive behavior.

### Issues
- Bots currently remain in battlegrounds after player leaves
- Battlegrounds can accumulate abandoned bots over time
- Inconsistent bot cleanup behavior
- Manual intervention needed to remove leftover bots

### How2Test
1. Join a battleground and add bots using the command
2. Verify bots join your battleground successfully
3. Leave the battleground
4. Verify that:
   - Bots leave the battleground with you
   - No bots remain behind in the battleground
   - Verify if the battleground is closed
5. Repeat test with multiple bots and different battleground types (WSG, AB, AV)

### Todo / Checklist
- [X] Modified default value of `m_temporary` in BattleBotAI.h
- [X] Updated AddBattleBot calls to pass true for temporary parameter
- [X] Tested bot creation works as expected
- [X] Tested bots properly leave battleground with player and battleground will be closed
- [X] Verified no negative impact on existing battleground functionality
- [X] Confirmed backwards compatibility for cases requiring permanent bots (can be overridden when needed)


mserajnik:
Unrelated to the PR, but I like checking the profiles of people who contribute to VMaNGOS and saw https://github.com/HerrTaeubler/VBots-v2-playerbotaddon; thanks for this, this looks very nice and will surely be appreciated by others as well.

Funnily, I was just today talking to a friend about a different bot management addon I remembered but never ended up using because it took up too much space for my liking (all the actions were combined in a single, large frame). I like how much more compact yours looks in comparison and will definitely give it a try.

ghost:
I hope you like the addon – if you find any errors or have suggestions for improvements, feel free to contact me on GitHub or Discord.

The ADdon is also a reason for the PR, as it automatically fills the battlegrounds with bots, and over time, more and more "zombie" battlegrounds accumulate, causing unnecessary confusion and consuming resources.

ratkosrb:
If this is the default behavior and there is no way to add non temporary bots, then the variable might as well not exist. The bots added with command are meant to be permanent. So people can add them once after booting their server, and they will remain and allow people to play battlegrounds with them until the server is shut off.

PR #2929, opened 2025-02-24
Fix some compiler warnings and explicitly set C++14 as the standard

schell244:
## 🍰 Pullrequest
Be more explicit about the current default compiler version and fix warnings that occur when compiling with newer versions.
Also fixed install path was displayed twice when configuring with CMake on Windows.

### Todo / Checklist
- [X] None


devland:
FYI, compilation on arch linux now fails because the AUR version of ACE requires c++17 but the cmake settings have it set to 14.
Updating line 21 to `set(CMAKE_CXX_STANDARD 17)` fixes it. :)

PR #2931, opened 2025-02-26
Fix CLI thread crash when running as systemd service with console enabled

ghost:
## 🍰 Pullrequest

Fixes a critical issue where the vMaNGOS core crashes when started as a systemd service with console enabled in mangos.conf. The CLI thread attempts to read from stdin even when it's not connected to a terminal, causing the server to crash.
The fix gracefully detects when stdin is not connected to a terminal and safely exits the CLI thread, allowing the server to run properly as a systemd service regardless of console settings.

### Proof

Server no longer crashes when running as a systemd service with console enabled
Logged message confirms proper thread termination
All other server functionality continues to work normally

### Issues

Fixes crash when running server as systemd service with console=1 in mangos.conf

### How2Test

Set console = 1 in your mangos.conf
Run the server as a systemd service: systemctl start mangos
Verify in logs that the CLI thread exits with message "CLI: stdin is not connected to a terminal, CLI thread exiting"
Confirm server continues running without crashing

### Todo / Checklist

[X] Added detection of terminal connection using isatty()
[X] Implemented early exit path for CLI thread when not connected to terminal
[X] Ensured proper database resource cleanup before thread exit

### DIFF
--- a/src/mangosd/CliRunnable.cpp
+++ b/src/mangosd/CliRunnable.cpp
@@ -83,19 +83,23 @@ int kb_hit_return()
 // %Thread start
 void CliRunnable::operator()()
 {
-    // Init new SQL thread for the world database (one connection call enough)
-    WorldDatabase.ThreadStart();                                // let thread do safe mySQL requests
-
     // Check if stdin is connected to a terminal
-    bool stdinIsTerminal = isatty(STDIN_FILENO);
+    bool stdinIsTerminal = false;
+    
+    #if PLATFORM == PLATFORM_WINDOWS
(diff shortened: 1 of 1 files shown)


src/mangosd/CliRunnable.cpp
```diff
@@ -29,6 +29,10 @@
 #include "Util.h"
 #include "CliRunnable.h"
 #include "Database/DatabaseEnv.h"
+#include "Log.h"
+#include <unistd.h>  // For isatty()
```
ratkosrb, line 33: no #ifndef WIN32 ? how is this gonna compile on windows?

Issue #2861, opened 2025-01-05
🐞 [Bug] Wyvern Sting dot gets removed by Immolation Trap dot


twowTurtReynolds:
Migrated report from https://turtle-wow.org/bug-report?id=5666

## 🐞 Bug report - Spell

Wyvern Sting applies a dot that deals nature damage over 12 seconds. As the spell description states, only one sting per hunter can be active on a target. Applying Serpent Sting will remove the Wyvern Sting dot, as it should. Applying Immolation Trap dot to the target also removes the Wyvern Sting dot, which should not happen.

EDIT: Tested with different ranks of Wyvern Sting and Immolation Trap, same result. Tested with Wyvern Sting and Explosive Trap, works as it should, doesnt replace Wyvern Sting dot. Problem thus appears to be with Immolation Trap, not with Wyvern Sting.

https://www.wowhead.com/classic/spell=14304/immolation-trap
https://www.wowhead.com/classic/spell=24132/wyvern-sting

### Expected behavior

Mob should have both the dots of Wyvern Sting and Immolation Trap on it.
### Steps to reproduce

1. Place an Immolation Trap. 
2. Fire a Wyvern Sting on a mob and either wait for it to run out or wake it up with an autoshot/trueshot/arcane shot 
3. See the Wyvern Sting dot applied to the mob after the cc ends. 
4. Have the mob walk into your Immolation Trap. 
5. Application of Immolation Trap dot instantly removes Wyvern Sting dot.

### Version & Environment

Client Version: 1.12.1

Commit Hash:  

OS Client: 
OS Server: 

### Crashlog

- None

ratkosrb:
https://github.com/vmangos/core/commit/92d8659edb09a0815745235257ca0bac8c3d3e77

PR #2936, opened 2025-02-27
Fix Wyvern Sting/Immo Trap

ghost:
## 🍰 Pullrequest
 fixes an issue where the Wyvern Sting DoT effect is incorrectly removed when a target triggers an Immolation Trap. The fix allows both DoT effects to coexist as intended by adding a specific stacking rule for these spells.

### Proof
- Bug report from players experiencing Wyvern Sting DoT being removed by Immolation Trap
- Verified in testing that the issue occurs with all ranks of Wyvern Sting and Immolation Trap

https://github.com/vmangos/core/issues/2861

### How2Test
- Apply Wyvern Sting to a target
- Wait for the cc effect to end and the DoT to be applied
- Have the target trigger an Immolation Trap
- Verify that both the Wyvern Sting doT and Immolation Trap DoT remain active on the target

### Todo / Checklist
- [X] None


mserajnik:
Did this fix #2861 entirely? If yes, then that issue can be closed, I guess.

PR #2941, opened 2025-03-01
Fix: Witherbark Witch Doctor Totem Summoning

ghost:
- Updated Witherbark Witch Doctor to only cast Healing Ward spell (ID: 5605)-
- Removed reference to spell_list_id to ensure direct spell usage

### DIFF
--- a/sql/migrations/20250301081546_world.sql
+++ b/sql/migrations/20250301081546_world.sql
@@ -7,19 +7,19 @@ BEGIN
     IF v=0 THEN
         INSERT INTO `migrations` VALUES ('20250301081546');
         
-        -- Update the Witherbark Witch Doctor to only use Healing Ward spell
+        -- Create a new spell list with only the Healing Ward spell
+        DELETE FROM `creature_spells` WHERE `entry` = 25550;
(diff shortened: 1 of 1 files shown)


Wall-core:
spell_template is never to be edited, unless assigning data to script_name. 
spell_template is data taken directly from the clients.

ghost:
@Wall-core Thanks, Brotalnia already explained it to me too^^

This is a temporary fix for now, as I still need the proper timings from a sniff. However, it's better than having this behavior with the Magma Totem.

Wall-core:
spell_template is never to be edited, unless assigning data to script_name. No exceptions there; that data is 100% accurate. You have to find some other way to accomplish fixes, either by using one of the other tables which can modify it (spell_mod/spell_effect_mod) or handling it in the core.

ghost:
I already have the sniff data and will make the changes now.

ghost:
What's missing now is the probability data to more accurately set the chances of which totem will spawn, since it's not included in the sniff data.

Daribon:
> spell_template is never to be edited, unless assigning data to script_name. No exceptions there; that data is 100% accurate. You have to find some other way to accomplish fixes, either by using one of the other tables which can modify it (spell_mod/spell_effect_mod) or handling it in the core.

There is one exception:

_"Apply custom flags in spell_template table. Only mod other values in the spell_mod table. Since this is a custom field, not coming from dbc, we are not losing any original data by modifying customFlags in spell_template"_

ghost:
I've already adjusted the SQL update to the sniff values. It just needs someone to test how often the magma totem is spawned.

If you take a look at the NPC in its current state, it mostly spawns the magma totem, and if it does cast healing ward, it's usually almost instantly overwritten by the magma totem.

PR #2942, opened 2025-03-01
Remove unused auction house bot configuration options.

mserajnik:
## 🍰 Pullrequest
This removes the unused auction house bot configuration options `AHBot.ah.id`, `AHBot.bot.guid` and `AHBot.bot.account`.

I believe `AHBot.ah.id` will generally not be needed anymore since the auction house(s) is/are now determined through `AHBot.ah.fid` instead.

And while `AHBot.bot.guid` and `AHBot.bot.account` could be used to display a seller name for the listings added by the bot, the code for this is missing/incomplete and thus these configuration options are confusing for users. In addition, if this were to be implemented, a single setting (for the character ID) should probably suffice (instead of having to provide a character ID and an account ID).

The PR also improves the comments for the leftover options to make it a bit clearer how to configure the auction house bot.

### Proof

- None

### Issues

- None

### How2Test

- Start the server with `AHBot.Enable = 1`
- Observe that the AH bot still works correctly

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/AuctionHouse/AuctionHouseBotMgr.cpp
+++ b/src/game/AuctionHouse/AuctionHouseBotMgr.cpp
@@ -66,9 +66,6 @@ void AuctionHouseBotMgr::Load()
     /* CONFIG */
     m_config                 = std::make_unique<AuctionHouseBotConfig>();
     m_config->enable         = sConfig.GetBoolDefault("AHBot.Enable", false);
-    m_config->ahid           = sConfig.GetIntDefault("AHBot.ah.id", 7);
-    m_config->botguid        = sConfig.GetIntDefault("AHBot.bot.guid", 1123);
-    m_config->botaccount     = sConfig.GetIntDefault("AHBot.bot.account", 32377);
     m_config->ahfid          = sConfig.GetIntDefault("AHBot.ah.fid", 120);
     m_config->itemcount      = sConfig.GetIntDefault("AHBot.itemcount", 2);
 
@@ -92,7 +89,7 @@ void AuctionHouseBotMgr::Update(bool force /* = false */)
     if (!(m_config->enable || force))
         return;
 
-    if (m_items.empty() ||  /*m_config->botguid==0 ||*/ m_config->botaccount == 0)
+    if (m_items.empty())
     {
         sLog.Out(LOG_BASIC, LOG_LVL_ERROR, "AHBot::Update() : Bad config or empty table.");
         return;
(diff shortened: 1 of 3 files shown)


gajet5:
I agree, I should have deleted it.
**AHBot.ah.id**

But I find other parameters useful. It was necessary to correct the description in the configuration.
It is useful to know who is selling the lot, a bot or a live player. Yes, you need to create a separate account for this, but it's useful.

```c++
    auctionEntry->owner              = m_config->botguid;
    auctionEntry->ownerAccount       = m_config->botaccount;
``` 

I'll leave this on my server, as people are already used to seeing the bot's name.

gajet5:
If you also want to see the name, I recommend making these changes.

In: https://github.com/vmangos/core/blob/development/src/game/AuctionHouse/AuctionHouseMgr.cpp#L259

add this:
```c++
// owner is AHBot
uint32 ahbot_guid = sConfig.GetIntDefault("AHBot.bot.guid", 0);

// owner exist
if ((owner || owner_accId) && ahbot_guid != auction->owner) //owner not a ahbot
```

ratkosrb:
Ah but were these parameters used or not? I thought they are unused that's why I merged this.

gajet5:
That's right, they weren't used, but it was a bug or a flaw. For a large number of users, this is not important.

The auction will show an empty field in the seller column.

mserajnik:
> It is useful to know who is selling the lot, a bot or a live player.

I think for most players, the current distinction (bot has no seller name compared to actual players) is probably fine.

I agree it would be nice to have the option to have the bot show a seller name too, but that implementation was completely missing, which is why it seemed right to remove those unused variables for now.

We can add the seller name functionality in a separate PR, but maybe so that only a character ID is needed and not also an account ID (as I mentioned in the PR description), as we can easily infer the account ID from the character, right?

gajet5:
> I think for most players, the current distinction (bot has no seller name compared to actual players) is probably fine.

That's right, I think the same.

> We can add the seller name functionality in a separate PR, but maybe so that only a character ID is needed and not also an account ID (as I mentioned in the PR description), as we can easily infer the account ID from the character, right?

Used in the "RemoveAuction" method.
https://github.com/vmangos/core/blob/development/src/game/AuctionHouse/AuctionHouseMgr.cpp#L42
I think it has something to do with account deletion, but I'm not sure. I didn't see where it was needed anymore. Therefore, when adding a name display, this must be taken into account.

1. In fact, you need to do the same thing that was deleted, but add corrections.
https://github.com/vmangos/core/blob/development/src/game/AuctionHouse/AuctionHouseBotMgr.cpp#L166
https://github.com/vmangos/core/pull/2942#issuecomment-2692573700
2. Modify the description in the configuration.
3. Delete things if they belong to a bot.
https://github.com/vmangos/core/pull/2942#issuecomment-2692575172

Issue #2570, opened 2024-04-06
🐛 [Bug] Darkmoon Faire - Tonk Control Console not working correctly


mserajnik:
## 🐛 Bug report
Using any Tonk Control Console at the Darkmoon Faire in Elwynn Forest results in a `Invalid target` message and locks the player in place. `You are afflicted by Summon RC Tank` can be seen in the combat log, but nothing else seems to happen. It looks like the player stays locked in place indefinitely and can't move (until relogging or probably also teleporting). The Tonk Control Console gets put into the activated state and can't be used again for a while.

When playing as a Hunter with a summoned pet, instead of a `Invalid target` message a `You already control a summoned creature` can be seen, but the rest of the described behavior (with locking the player in place) is the same.

I encountered this first while being in a party and assumed it's maybe a bug relating to a party member not being nearby (since a Tonk should supposedly be spawned for each party member as mentioned further below), but the behavior is the same regardless of being in a party or not.

I haven't tested it in Mulgore, but I assume the issue would also exist there.

### Expected behavior
Apparently, using a Tonk Control Console should spawn a Tonk for the player to control with various abilities that allows battling other players who are also controlling a Tonk. I've never done this myself in Vanilla, so I can't confirm if this is accurate, but that's at least how it is in Classic.

When in a party, it should apparently also spawn a Tonk for each party member if they are nearby. I am not sure what is supposed to happen when a party member is offline or not nearby (if the activation then just fails or if only the player who activated and the other nearby party members get to control a Tonk).

### Steps to reproduce

1. Find a Tonk Control Console at the Elwynn Forest Darkmoon Faire and use it
2. Observe the `Invalid target` message (or, when playing as a hunter with a summoned pet, the `You already control a summoned creature` message)
3. Observe that the Tonk Control Console gets put into the activated state and can't be used again for a while
4. Observe that no Tonk gets spawned
5. Observe that the player gets locked in place without being able to move indefinitely

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash: https://github.com/vmangos/core/commit/3febff1d84059c970bbeedd3f1871260bde18e03

OS Client: Arch Linux (using Proton)
OS Server: Ubuntu 22.04

### Crashlog

Not a crash log (the server doesn't crash), but this gets logged each time a player tries to use the Tonk Control Console:

```
[...]

PR #2948, opened 2025-03-03
Unsummon pet when using the Tonk Control Console.

mserajnik:
## 🍰 Pullrequest
This unsummons a potential Hunter or Warlock pet when using the Tonk Control Console, fixing the outstanding issue I described [here](https://github.com/vmangos/core/issues/2570#issuecomment-2688869977) where trying to use the Console with a summoned pet would lock you in place but not let you control a Tonk (requiring you to relog so you can move again).

- **Without the fix:** https://files.catbox.moe/eydu0u.mp4 (note the `You already control a summoned creature` message; also, you can't see it in the video, but I am trying to move and jump, but besides turning nothing works)
- **With the fix:** https://files.catbox.moe/5e5kpq.mp4

__Note:__ there are still at least two other outstanding issues with the Tonk Control Console (both very minor though) that I mentioned in [this comment](https://github.com/vmangos/core/issues/2570#issuecomment-2688869977) that are _not_ addressed by this PR.

### Proof
I can't find much information about how it originally behaved in Vanilla, but according to [this video](https://www.youtube.com/watch?v=YBfGJoqm2Fg&t=1675s) (relevant part timestamped), at least in Classic the pet should be unsummoned when using the Tonk Control Console. The other option would be to not allow Hunters and Warlocks to use the Tonk Control Console with a summoned pet (but in that case, they obviously shouldn't get locked in place).

### Issues

- None

### How2Test

- `.event start 4` to start the Darkmoon Faire event in Elwynn Forest
- Use one of the Tonk Control Consoles at the Darkmoon Faire as a Hunter or Warlock with a summoned pet
- Observe that the pet is unsummoned and you end up controlling a Tonk

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20250303170125_world.sql
+++ b/sql/migrations/20250303170125_world.sql
@@ -1,0 +1,19 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20250303170125');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20250303170125');
+-- Add your query below.
+
+-- Spell script for Tonk Control Console (24934)
+UPDATE `spell_template` SET `script_name`='spell_darkmoon_steam_tonk_control_console' WHERE `entry`=24934;
+
+-- End of migration.
+END IF;
+END??
+DELIMITER ;
+CALL add_migration();
+DROP PROCEDURE IF EXISTS add_migration;
--- a/src/scripts/spells/spell_special.cpp
+++ b/src/scripts/spells/spell_special.cpp
@@ -39,6 +39,24 @@ SpellScript* GetScript_Meteor(SpellEntry const*)
     return new MeteorScript();
 }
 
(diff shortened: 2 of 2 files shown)


ratkosrb:
This is a horrible hack. You should use spell scripts, not add stuff like this anymore.

mserajnik:
Thanks for the feedback. As I said:

> Maybe there is a more elegant (SQL?) solution for the problem, but I thought I'd share this fix regardless.

~~I'll look into implementing a more proper solution using spell scripts.~~

~~__Edit:__ I've changed this PR to a draft in the meantime.~~

__Edit 2:__ Re-implemented.

mserajnik:
I've re-implemented the fix using a spell script now:

- I've put the spell script into [`src/scripts/spells/spell_special.cpp`](https://github.com/vmangos/core/blob/54b6b31378fad7e3f50837349444bcf8b89beaee/src/scripts/spells/spell_special.cpp) since that seemed like the most logical place to me (there is already another Tonk-related spell script in there)
- I used the `OnInit` hook since I realized the `OnCheckCast` would be triggered too late for this case. Maybe a new trigger like `OnBeforeCheckCast` might be good to have, but I didn't want to add that just for this use case

Please let me know if this is acceptable (or not) and if you want any further changes.

PR #2954, opened 2025-03-08
Check Item Level for Heavy Armor Kit after Patch 1.7

### LINKED ISSUE #2388: 🐞 [Bug] Armor Kits have no itemlevel requirement.
They check player level(for cast) but not the itemlevel of the gear itself.

FlagFlayer:
## 🍰 Pullrequest

No longer allows Heavy Armor Kits to work on items with an item level lower than 15 from patch 1.7 onward.

### Issues

- Fixes https://github.com/vmangos/core/issues/2388

### How2Test

- Server on patch 1.7 or later
- .additem 4265
- Attempt to use it on an item with an item level lower than 15
- Spell fizzles because the target item is too low level

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20250308172541_world.sql
+++ b/sql/migrations/20250308172541_world.sql
@@ -1,0 +1,18 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20250308172541');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20250308172541');
+-- Add your query below.
+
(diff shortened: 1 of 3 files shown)


ratkosrb:
Shouldn't this be part of the spell_template data?

ratkosrb:
Actually it should be a spell script.

FlagFlayer:
You'd think it'd be part of spell_template data but the baseLevel for both versions of the spell are 0, the only thing that changes between patches is the description.

I haven't used spell scripts before so correct me if I'm wrong: The spell_template entry for Armor +24 (2833) only has the effect SPELL_EFFECT_ENCHANT_ITEM, and does not have SPELL_EFFECT_SCRIPT_EFFECT. My basic understanding is that SPELL_EFFECT_SCRIPT_EFFECT is required to load spell scripts. Is there a different way a spell script can be used?

Wall-core:
there's a new column on spell template for spellscripts (last column)

FlagFlayer:
So I'd add a script name to spell_template and write the code for this in ~~spell_special.cpp~~ spell_item.cpp?

PR #2961, opened 2025-03-12
Update Majordomo Executus scripts

Wall-core:
Complete overhaul of the Majordomo scripts, based on sniff data.
- Enumerates all things; no more magic numbers
- Implements all the missing spells/abilities
- Properly handles all spell functions
- Adds correct sequencing for his post-fight dialogue
- Adds correct sequencing for pre-Ragnaros dialogue
- Fixes Hot Coals gameobjects mouseover
- Correctly implements Cache of the Firelord handling in case of raid reset or server crash; no more double loots
- Implements despawning of Majordomo after being slain by Ragnaros

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_majordomo_executus.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_majordomo_executus.cpp
@@ -1,43 +1,9 @@
 #include "scriptPCH.h"
 #include "molten_core.h"
 
-enum
-{
-    SAY_AGGRO                   = 7612,
-    SAY_SPAWN                   = 7566,
-    SAY_SLAY                    = 9425,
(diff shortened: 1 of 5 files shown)


src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_majordomo_executus.cpp
```diff
@@ -1,12 +1,37 @@
 #include "scriptPCH.h"
 #include "molten_core.h"
 
+#define GOSSIP_ITEM_1               "Tell me more."
+#define GOSSIP_ITEM_2               "What else do you have to say?"
+#define GOSSIP_ITEM_3               "You challenged us and and we have come. Where is this master that you speak of?"
```
ratkosrb: These texts can be removed since they're in the db. Not used in the script.

PR #2964, opened 2025-03-12
Update Molten Core boss scripts

Wall-core:
Implements Separation Anxiety in Garr scripts and further corrects on-aggro behavior of Majordomo Executus

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
@@ -75,13 +75,10 @@ struct boss_garrAI : ScriptedAI
 
             for (const auto& itr : firesworn)
             {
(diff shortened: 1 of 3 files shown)


src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
```diff
@@ -75,7 +75,13 @@ struct boss_garrAI : ScriptedAI
 
             for (const auto& itr : firesworn)
             {
-                m_lFiresworn.push_back(itr->GetObjectGuid());
+                if (itr->GetObjectGuid())
```
ratkosrb: the guid cant be 0

Wall-core: Updated to remove check

Wall-core:
updated to remove check

PR #2965, opened 2025-03-12
Fix Majordomo being invincible after intro

schell244:
### Todo / Checklist
- [X] None

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_majordomo_executus.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_majordomo_executus.cpp
@@ -351,6 +351,7 @@ struct boss_majordomoAI : public ScriptedAI
                 }
                 break;
             case 76:
+                m_creature->SetInvincibilityHpThreshold(0);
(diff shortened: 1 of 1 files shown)


Wall-core:
is this necessary? He dies automatically from Ragnaros

schell244:
Yes he's supposed to die when Ragnaros casts SPELL_ELEMENTAL_FIRE on him (the script sets his hp to 35 when he turns friendly) but with that flag not removed, he can't die and fights Ragnaros instead

Wall-core:
Elemental Fire is the sniffed spell he uses, and it's an instakill effect which doesn't require this hackfix. The previous spell (Molten Blast) was incorrect and is part of his combat spells, which only deals 6k damage

schell244:
When `SetInvincibilityHpThreshold` is active, he even seems to ignore the instakill effect.
I've tested the script and Domo would not die unless with that line being restored.

Wall-core:
Maybe the way instakill/DealDamage/Kill has changed since I last ran updates, for me it works fine without this but it can be added regardless.

https://github.com/user-attachments/assets/631119b3-4fe6-4ff3-bc83-8682a8d21018

ratkosrb:
Invincibility HP Threshold prevents death. It has to be turned off when he's supposed to die.

PR #2971, opened 2025-03-18
Rework Garr script

schell244:
## 🍰 Pullrequest
Fully rework Garr script.
- corrected some spell ids with sniffs
- timers from cmangos
- ~~added functionality to despawn adds when Garr dies - but commented out, as I don't know if thats correct.~~

### Todo / Checklist
- [X] None

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
@@ -1,167 +1,228 @@
-/*
- *
- */
-
 #include "scriptPCH.h"
 #include "molten_core.h"
+#include "Utilities/EventMap.h"
 
-enum
+enum Garr : uint32
 {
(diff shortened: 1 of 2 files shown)


src/scripts/eastern_kingdoms/burning_steppes/molten_core/boss_garr.cpp
```diff
@@ -1,200 +1,268 @@
...
             m_pInstance->SetData(TYPE_GARR, DONE);
-    }
+        }
 
-    void FireswornJustDied(ObjectGuid fireswornGuid)
-    {
-        m_lFiresworn.remove(fireswornGuid);
-        DoCastSpellIfCan(m_creature, SPELL_ENRAGE, CF_TRIGGERED);
+        // TODO:
+        // When Garr dies, will his adds also die?
```
Daribon: His adds will remain alive when Garr dies, the usual tactic is to kill Garr first and then adds.

PR #2979, opened 2025-03-22
Add Linux Build Github Action

NickTyrer:
## 🍰 Pullrequest
Add linux build to github actions. I have very limited knowledge in this area so this will need reviewing to make sure it is correct.

### DIFF
--- a/.github/workflows/linux-dev-release.yml
+++ b/.github/workflows/linux-dev-release.yml
@@ -27,10 +27,10 @@ jobs:
     - name: Build
       run: |
         sudo apt-get -qq update
-        sudo apt-get -qq install build-essential cmake libace-dev libmysql++-dev libtbb-dev libcurl4-openssl-dev openssl
+        sudo apt-get -qq install build-essential cmake libace-dev libtbb-dev libmysqlclient-dev
         mkdir build
         cd build
-        sudo cmake ../ -DCMAKE_INSTALL_PREFIX=/vmangos -DWITH_WARNINGS=0 -DDEBUG_SYMBOLS=0 -DSUPPORTED_CLIENT_BUILD=5875 -DBUILD_FOR_HOST_CPU=0
+        sudo cmake ../ -DCMAKE_INSTALL_PREFIX=/vmangos -DDEBUG_SYMBOLS=0 -DSUPPORTED_CLIENT_BUILD=5875 -DBUILD_FOR_HOST_CPU=0
         sudo make -j $(nproc)
         sudo make install
         
--- a/.github/workflows/linux-dev-release.yml
+++ b/.github/workflows/linux-dev-release.yml
@@ -10,6 +10,8 @@ on:
       - '.github/PULL_REQUEST_TEMPLATE.md'
       - '.github/workflows/db_check.yml'
       - '.github/workflows/db_dump.yml'
+      - '.github/workflows/dev-release.yml'
+      - '.github/workflows/vmangos.yml'
       - 'sql/**'
       - '.drone.yml'
       - 'README.md'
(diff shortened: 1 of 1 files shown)


ratkosrb:
@mserajnik Can you review this since you've worked on the workflows before?

mserajnik:
@ratkosrb Sure, I'll have a closer look later today.

mserajnik:
The workflow as such seems fine, [here](https://github.com/mserajnik/vmangos-core/actions/runs/14115407914/artifacts/2834406288) is a link to the resulting artifact from a test run I did on my fork.

1. This is compiled on Ubuntu and linked to Ubuntu-specific dependency versions, meaning this won't just run on any Linux  distribution that is not at least Ubuntu-based
2. It's a dynamic build, so even if the user is on Ubuntu, he'll still need to install the runtime dependencies

Static linking would address both of these issues, but I am not sure how feasible this is; I frankly don't have any experience with static linking on Linux.

I think a better approach for Linux would be to offer an AppImage ~~or a Flatpak~~ (__Edit:__ AppImage is probably the saner choice as connecting to a MySQL DB on the host would be a bit of a pain with Flatpak) as those are self-contained and will work on any distro.

Ultimately, it would of course be easiest to just leave the distribution for Linux to the distros/distro communities. E.g., for Arch there is [this AUR package](https://aur.archlinux.org/packages/vmangos-git).

NickTyrer:
Thanks for the review. Here is my reasoning behind adding Linux binaries to the releases, I just want a centralised method to obtain the vmangos binaries without the need to compile or rely on third party sources. I would run the binaries using docker, something like:

```  mangosd:
    image: ubuntu:noble
    container_name: mangosd
    environment:
      - TZ=GB
    ulimits:
      nofile: 1024
    tty: true
    stdin_open: true
    stop_grace_period: 2m
    depends_on:
      mangos_db:
        condition: service_healthy
    restart: unless-stopped
    volumes:
      - mangosd_volume:/vmangos/bin
      - /home/vmangos/mangosd_config:/vmangos/etc
      - /home/vmangos/data:/vmangos/data
      - /home/vmangos/logs:/vmangos/logs
      - /home/vmangos/honor:/vmangos/honor
    command:
      - /bin/bash
      - -c
      - |
        export DEBIAN_FRONTEND=noninteractive
        apt-get -qq update
        apt-get -qq -y install libace-dev libtbb-dev libmysqlclient-dev curl
        [ ! -f /vmangos/bin/mangosd ] && curl -L --create-dirs --output /vmangos/bin/mangosd "https://github.com/vmangos/core/releases/download/latest/mangosd"
        [ ! -f /vmangos/etc/mangosd.conf ] && curl -L --create-dirs --output /vmangos/etc/mangosd.conf "https://github.com/vmangos/core/releases/download/latest/mangosd.conf"
        chmod +x /vmangos/bin/mangosd
        ./vmangos/bin/mangosd
    network_mode: service:mangos_db
```
Maybe its too niche but it would be usefull for me to quickly setup a disposable dev environment (I only do SQL work).

NickTyrer:
Also I looked into creating standalone binaries using static linking but it is way beyond me.

mserajnik:
> I just want a centralised method to obtain the vmangos binaries without the need to compile or rely on third party sources. I would run the binaries using docker

Shameless plug: Maybe [my Docker setup](https://github.com/mserajnik/vmangos-deploy) might be useful for this. You’d still be relying on a third party, but I plan to keep on maintaining it indefinitely.

Other than that, I don’t think there’s a downside to adding this workflow; but maybe the limitations of these builds should be documented somewhere so Linux users don’t just assume the binaries will work on their system without any extra work on their end.

NickTyrer:
Nice work with vmangos-deploy, I just like the guarantee that having the binaries attached to the main repo will result in them always being up to date.

PR #2984, opened 2025-03-27
BWL code cleanup

schell244:
## 🍰 Pullrequest
Improved code style.
Replaced magic numbers with constants and updated naming of some existing ones.
Also replaced two hardcoded texts with matching broadcast texts.

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None


src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/instance_blackwing_lair.cpp
```diff
@@ -14,25 +14,18 @@
...
-SD%Complete: 90
-SDComment:
-SDCategory: Blackwing Lair
-EndScriptData */
-
 #include "scriptPCH.h"
 #include "blackwing_lair.h"
 
-struct RazCoords
+struct EggCoords
```
ratkosrb: Why not just use Position struct?

PR #2986, opened 2025-03-27
Don't perform multiple searches within instanceGroups

Chaosvex:
Insertion causes a lookup before returning false if the value isn't in the set, so performing a search before an insertion is duplicating work that will end up being done again. It's more efficient to try to insert and then clear the icons if the insertion was successful.

Also switched to unordered_set as it'll perform better for this mostly read 'heavy' scenario.

Included the header rather than relying on the container being transitively included as was the case for `set`. Will make no difference to compilation speed if it's already included.

### DIFF
--- a/src/game/AI/CreatureAI.cpp
+++ b/src/game/AI/CreatureAI.cpp
@@ -206,11 +206,11 @@ void CreatureAI::ClearTargetIcon()
     std::unordered_set<Group*> instanceGroups;
 
     // Clear target icon for every unique group in instance
-    for (const auto& player : players)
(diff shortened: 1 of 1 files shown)


Chaosvex:
Used structured bindings and then remembered it's set to C++14. Should have double checked the change. :japanese_goblin:

ratkosrb:
Yes this project will remain on c++14 forever. Also we use east const not west const. Edited your PR to fix this.

Chaosvex:
I didn't change the const, that was existing code or the project's clang-format messed with it. Just for the record. 😄

Issue #2741, opened 2024-08-29
🐛 [Bug] 1.6.1.4544 build failing


mserajnik:
## 🐛 Bug report
Building for 1.6.1.4544 is currently failing:
```
/core/src/game/Objects/Player.cpp: In member function 'InventoryResult Player::CanUnequipItem(uint16, bool) const':
/core/src/game/Objects/Player.cpp:10335:16: error: 'EQUIP_ERR_NOT_WHILE_DISARMED' was not declared in this scope
10335 |         return EQUIP_ERR_NOT_WHILE_DISARMED;
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Full context [here](https://github.com/mserajnik/vmangos-deploy/actions/runs/10606414305/job/29397131632#step:7:3767) (the output is alternating between the `x86_64` and `aarch64` builds which is why it looks so discontinuous).

I assume the error occurs due to this addition in https://github.com/vmangos/core/commit/2f500b9bf6b67a7c20302e9b6f4c4dc820d8f596: https://github.com/vmangos/core/blob/ce50f98c9b3d0e1dfb87d903198a8dd246ffaf1a/src/game/Objects/Player.cpp#L10333-L10335

Because `EQUIP_ERR_NOT_WHILE_DISARMED` is not defined when building for 1.6.1.4544: https://github.com/vmangos/core/blob/ce50f98c9b3d0e1dfb87d903198a8dd246ffaf1a/src/game/Objects/Item.h#L108-L109

Commit Hash:  https://github.com/vmangos/core/commit/ce50f98c9b3d0e1dfb87d903198a8dd246ffaf1a

OS Server: Ubuntu 24.04 LTS

PR #2987, opened 2025-03-27
CI: Test build for all supported client versions.

mserajnik:
## 🍰 Pullrequest
This extends the existing test workflow to build for all currently supported client versions (instead of just 1.12.1) with each compiler (GCC, Clang, MSVC) which will help to prevent compilation errors that affect specific client version builds such as [this one](https://github.com/vmangos/core/issues/2741).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -30,22 +30,33 @@ on:
 
 jobs:
   build:
-    name: Build on ${{ matrix.job_name }}
+    name: Build on ${{ matrix.job_name }} (client build ${{ matrix.client_build }})
     runs-on: ${{ matrix.os }}
     permissions:
       contents: read
     strategy:
       matrix:
+        client_build:
+          - 5875
+          - 5464
+          - 5302
+          - 5086
+          - 4878
+          - 4695
+          - 4544
+          - 4449
+        os:
+          - ubuntu-24.04
+          - windows-2022
         include:
           - os: ubuntu-24.04
             c_compiler: gcc
             cxx_compiler: g++
-            compiler: gcc
             job_name: Ubuntu GCC
           - os: ubuntu-24.04
             c_compiler: clang
             cxx_compiler: clang++
-            compiler: clang
+            install_clang: true
             job_name: Ubuntu Clang
           - os: windows-2022
             job_name: Windows Visual Studio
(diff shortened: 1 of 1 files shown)


ratkosrb:
Won't this overload the CI cause it's too many builds?

mserajnik:
20 jobs can run at the same time. Since there are ~~21~~ **24** (as support for `1.5.1.4449` got added) here now, plus the jobs from the other workflows, some won’t start immediately, but only after others have finished.

So, yes, this will increase the overall time each set of workflow runs takes; imo that tradeoff would be worth it, considering there usually aren’t that many PRs/commits at once that would exhaust the available CI resources.

0blu:
@mserajnik Just as an inspiration for your next PR: can you take a look at removing `TBB` from the GitHub workflows?

0blu:
I mean we are not paying the bills for the server, soooo as long as GitHub allows those kinds of CI's, why not?

mserajnik:
> @mserajnik Just as an inspiration for your next PR: can you take a look at removing `TBB` from the GitHub workflows?

Sure, I didn't notice it didn't get removed from the workflows yet.

I wanted to do some action updates anyway with the upcoming Node.js 20 removal. I'll remove TBB then too.

mserajnik:
> I wanted to do some action updates anyway with the upcoming Node.js 20 removal. I'll remove TBB then too.

I'm changing this PR to a draft for now while I work on that (since it will need rebasing anyway).

mserajnik:
I've rebased this on top of https://github.com/vmangos/core/commit/3e54ff3dac5a88b4e0bcc005df774f2d9f1d7190.

> I mean we are not paying the bills for the server, soooo as long as GitHub allows those kinds of CI's, why not?

One other option would be to run the additional client version builds only for PRs, and keep only `1.12.1.5875` on `development` pushes, as it is today without this PR.

That still would have caught these three issues before https://github.com/vmangos/core/pull/3225 was merged:

- https://github.com/vmangos/core/pull/3308
- https://github.com/vmangos/core/pull/3310
- https://github.com/vmangos/core/pull/3311

So this would preserve most of the value without adding a large number of mostly unnecessary builds to every `development` push, assuming direct pushes to `development` (not coming from a PR merge) are not expected to introduce those kinds of build regressions.

Let me know if you'd prefer that.

PR #2993, opened 2025-03-31
Check player pointer in AuctionHouseHandler.cpp

Wall-core:
Prevent possible crash due to unchecked pointers


ratkosrb:
I don't believe it's possible for the pointer on the session to be null if the packet is being processed. This packet is only processed while in world.

ratkosrb:
https://github.com/vmangos/core/blob/development/src/game/Protocol/Opcodes.cpp#L674

Wall-core:
yeah the player pointer should never be null, but I figured that I'd strengthen it with checks. The auction should be checked though, and I should have included a check for auctionHouse at line 621

Wall-core:
After further review, it seems like this wouldn't be the culprit and these catches won't help much. Closing for now.

Issue #2363, opened 2023-12-20
🐛 [Bug]Opening chests should aggro all nearby enemies while channeling the "open chest" spell


ShiyoKozuki:
## 🐛 Bug report
Opening chests doesn't aggro nearby mobs until AFTER the "open chest" spell is done channeling.

### Expected behavior
Opening a chest should aggro all nearby mobs while channeling the "open chest" spell
https://wowpedia.fandom.com/wiki/Chest#:~:text=Opening%20chests%20will%20usually%20draw,it%20disappear%20after%205%20minutes.

https://youtu.be/xXblOr92hQs?list=PLdYVrYJB1PTHV-apnXYXkRhCI-oFR6QIc&t=826

### Steps to reproduce

1.
2.
3.
4. ...
5. Profit

### Version & Environment

Client Version:  1.12.1

Commit Hash:   current

OS Client:  win10
OS Server:  win10

### Crashlog

- None

GoonBoiBarry:
I don't see anything about yards in that link and that's not really gonna cut it as a reliable source for this project in my experience. If you think it's wrong you need to do proper testing on the official servers and provide evidence. In my experience looting chest do aggro mobs on vmangos, but whether the range is right neither of us can attest to without actual testing.

BDuncanson-17:
The DBC properties for a general chest GO do not suggest any aggro related mechanics. I would guess that article is referring to the fact that opening a chest roots a players and makes a target for proximity aggro of any nearby mobs who are randomly pathing. You can probably close this ticket.

ShiyoKozuki:
> The DBC properties for a general chest GO do not suggest any aggro related mechanics. I would guess that article is referring to the fact that opening a chest roots a players and makes a target for proximity aggro of any nearby mobs who are randomly pathing. You can probably close this ticket.

Nearby mobs are supposed to be aggrod while you are trying to open a chest.

mserajnik:
> The DBC properties for a general chest GO do not suggest any aggro related mechanics. I would guess that article is referring to the fact that opening a chest roots a players and makes a target for proximity aggro of any nearby mobs who are randomly pathing. You can probably close this ticket.

I got curious because to me it also felt like opening chests causes aggro in VMaNGOS. From
[...]

PR #2997, opened 2025-04-05
Apply 10 yard aggro when beginning to open an object that uses "Treasure" faction and lock type

FlagFlayer:
## 🍰 Pullrequest

Self explanatory title.

### Proof

From testing I found there are 2 conditions needed for an object to cause aggro:

- The object uses a lockId that uses LOCKTYPE_TREASURE in Lock.dbc.

- The object must use a faction template that uses faction 77 ("Treasure").

Aggro will result only if both requirements are met.

Results from testing on Classic 1.15 (can supply videos on request, forgot to film the objects that didn't cause aggro):

- Opening an object with a lockId that uses LOCKTYPE_TREASURE but without a faction that doesn't use ID 77, e.g. Cat Figurine (Faction 0, lockId 57) or Venture Co. Engineering Plans (Faction 84 - uses Alliance Generic, lockId 57) does not aggro nearby mobs.

- Opening an object with a faction that uses ID 77 but a lockId that doesn't use LOCKTYPE_TREASURE, e.g. Food Crate (Faction 94, lockId 43) does not aggro nearby mobs.

- Opening an object with faction 94 and lockId 57 results in aggro:

https://i.imgur.com/wDSV9do.mp4

- Opening a locked object with a lock that uses LOCKTYPE_TREASURE with faction that uses ID 77 (via any method) results in aggro:

https://i.imgur.com/eLaEJCB.mp4
 
Aggro occurs based on line of sight on any creatures that can attack the player (even if they're already in combat with a different player, the player will be set in combat with them on beginning to open the object)

### Issues

- https://github.com/vmangos/core/issues/2363

### How2Test

- Open a chest -> Aggros nearby mobs
- Open an object that has a different lockId and/or faction -> Does not aggro nearby mobs

### Todo / Checklist

- [X] None


ratkosrb:
The indentation looks totally whack.

FlagFlayer:
Looked fine on the file itself. I'll try and adjust it

FlagFlayer:
~~Currently has an issue with handling factions 101 and 102 since the objects are openable by the object faction's friendly mask and should aggro based on the faction's hostile mask.~~ Hostile marks are to prevent players from the same faction as the object from opening it (e.g. preventing Alliance players opening Alliance chests). The aggro is sent to any valid attacker in range.

FlagFlayer:
> Currently has an issue with handling factions 101 and 102 since the objects are openable by the object faction's friendly mask and should aggro based on the faction's hostile mask.

Fixed now. Should be good to go.

ShiyoKozuki:
I've been waiting for this for like 2 years!
Any chance you can merge in the current core to this? It has tons of merge conflicts if I try to take it as a PR.

FlagFlayer:
> I've been waiting for this for like 2 years! Any chance you can merge in the current core to this? It has tons of merge conflicts if I try to take it as a PR.

Done

ShiyoKozuki:
Thank you! Works wonderfully.
However, this still has the bug where opening rogue lockers that require lock picking skill aggro everything. (i.e. lockers in wetlands, NOT locked TREASURE chests)

FlagFlayer:
> However, this still has the bug where opening rogue lockers that require lock picking skill aggro everything.

Yes. That is intended and correct behaviour. See [this video](https://i.imgur.com/eLaEJCB.mp4) from Classic.

ShiyoKozuki:
Hi
This crashes the server without an error when I attempt to open "Practice Lockbox" in Redridge on a level 20 rogue human rogue with level 0 skill lockpicking
Reverting this commit fixes the crash

FlagFlayer:
@ShiyoKozuki can you send the error log from the crash?

ShiyoKozuki:
How do I do that?

FlagFlayer:
There's a txt file in the vmangos folder

ShiyoKozuki:
> There's a txt file in the vmangos folder

Please let me name what folder and the name, I can't find it.

FlagFlayer:
Errors before the crash are logged in the Server.log file

What I'm asking for is that you launch your server, cause it to crash via the bug you found, and then send the Server.log file.

FlagFlayer:
The issue was that practice locks use faction 0, so GetFactionTemplateEntry returns null which the code then attempts to use as a pointer. It's a quick fix.

ShiyoKozuki:
Can you pull the main repo in again? Tons of merge conflicts again

FlagFlayer:
@ShiyoKozuki can't you just copy these changes to your own repo? It's literally removing one paragraph at one place, and adding one at another that you can copy paste

mserajnik:
~~I don’t even see any conflicts, the patch applies just fine.~~

__Edit:__ Never mind, it actually doesn't apply on `development`, I had a different ref checked out, so my previous suggestion here won't work.

codestothestars:
This branch merges with the development branch just fine. See above the comment box, "No conflicts with base branch". There may be a conflict somewhere, but it isn't here.

mserajnik:
The branch merges fine, but it won’t apply [as a Git patch](https://github.com/vmangos/core/pull/2997.patch) because it contains outdated references. That is because GitHub only uses its auto-resolution for the actual PR merge, but it doesn’t update the generated patch file.

I assume Shiyo tried applying the patch or manually merging the branch from this PR which would present the same conflicts.

codestothestars:
Thanks @mserajnik I wasn't aware that GitHub even generated such a file. But yeah anyone wanting to pull these changes in, I would recommend just going with a git merge, as it doesn't result in a conflict with the development branch.

ShiyoKozuki:
image
Is what I'm doing

codestothestars:
@ShiyoKozuki That's fine to do, but the conflicts are only with your own fork's development branch, not the one in the vmangos repo. You'll need to resolve the conflicts on your side.

ShiyoKozuki:
Ok thanks
Why isn't this merged into release yet?

ratkosrb:
1. It hardcodes a faction id which i do not like.
2. The lack of { } around multiple huge nested if blocks makes it look ugly.
3. I suspect this should be tied to specific Opening spells that are used for the lock types in question, and thus could be a spell script attached to them.
4. It arbitrarily checks only effect index 0, which seems weird.
5. It manually casts the caster to Unit instead of simply using m_casterUnit like everywhere else in the code base.
6. It's checking the default faction on the gameobject_template instead of the gameobject's current faction written in the update fields.
7. Uses an any unit searcher, which will cause even players to aggro if under some kind of charm, as well as yellow neutral mobs and the pets of enemy faction players.
8. No check if the target is already in combat, and thus will override current aggro and any threat to force the unit to attack whoever opens the chest. It should add threat instead of calling AttackStart to not override current target, unless there is evidence threat is ignored like this on classic.

PR #3002, opened 2025-04-10
Improve group member position updates across maps

ghost:
🍰 Pullrequest
This PR improves the group member position update system to ensure reliable position tracking across different maps and visibility ranges. The changes maintain proper visibility rules for non-position updates while ensuring group members can always see each other's positions.

#Proof
Tested with group members on different maps
Verified position updates work beyond visibility range

#Issues
fixes Group member positions not updating across maps

#How2Test
1. Create a group with multiple players
2. Have group members move to different maps
3. Verify that:
Position updates are received by all group members

#Todo / Checklist
[X] Tested with different map scenarios

### DIFF
--- a/src/game/Objects/Player.cpp
+++ b/src/game/Objects/Player.cpp
@@ -4262,6 +4262,32 @@ void Player::BuildCreateUpdateBlockForPlayer(UpdateData& data, Player* target) c
 
 void Player::DestroyForPlayer(Player const* target) const
 {
+    // Clean up visibility lists when players are on different maps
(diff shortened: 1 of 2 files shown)


ratkosrb:
Why would the other players be in the visible list across maps?

ghost:
Fixed an important visibility tracking issue: Players on different maps are now properly removed from visibility lists.

Previously, player IDs remained in visibility lists even when players were on different continents/maps. Now we correctly clean up the visibility lists when players are on different maps while still ensuring position updates get sent to all group members.

PR #3004, opened 2025-04-10
Prevent long distance following

Wall-core:
Prevent bots from following beyond reasonable distances, as following outside visibility distance will cause units to fly across the screen and can lead to crashes on grid unload


ratkosrb:
Bots already do not follow at long distances. They teleport beyond visibility range. Also as you can see in his crash log, it crashed upon using a gm command, which actually does not even exist in vmangos, so he coded a custom command badly.

Wall-core:
Ah, I thought he performed a .tele across the same continent map and it attempted to find a path.

PR #3007, opened 2025-04-13
Prevent crash in MoveMap.cpp

Wall-core:
Adds crash protection if mmap fails to load.

Happened today while updating Chrome with server running, likely some form of memory corruption which caused a failure to load the mmap properly before passing it on. This adds checks to instead return if data is null.


ratkosrb:
Isn't the assert after this pointless if we check for null beforehand?

Wall-core:
updated to remove ASSERT

PR #3013, opened 2025-04-17
Improve waypoint movement

### LINKED ISSUE #2713: 🚀 [Feature] Improve waypoint movement
As it is now the creatures reset their animation after reaching a new waypoint node. To prevent this a new spline has to be sent to the client while the old one is still running. My old project was based on an AzerothCore fork from 2020 which was itself based on SunwellCore. They fixed this by starting a new spline 200ms before reaching the next node, which prevented the animation "hiccups", but was only a workaround. CMaNGOS seems to have implemented a better solution by sending a new spline 1.5s before reaching the end of the old one and ensuring that a spline is at least 6s long by including multiple nodes if necessary:
https://github.com/cmangos/mangos-classic/blob/master/src/game/MotionGenerators/WaypointMovementGenerator.cpp

I tried to implement this in my own project but gave up eventually (well, actually I gave up the entire project recently). Perhaps something similar could be implemented in VMaNGOS.

ghost:
## 🍰 Pullrequest
Adjust the algorithm for waypoint movement similar to CMaNGOS:
- Try to generate movement splines which are at least 6s long
- 1.5s before the spline is finished start a new spline

This prevents the animation resets occuring at each waypoint. I tested this with all kinds of different waypoint types with and without creature groups (see below). Please take a look at the code changes, I'm not a C++ professional.

### Issues
Closes #2713

### How2Test
Watch different creatures which use waypoint movement, e.g.:
- Thora Feathermoon (ID 5798, GUID 15169)
- Theresa (ID 5697, GUID 41840)
- Lord Kazzak (ID 12397, GUID 156950)
- Felguard Elite (ID 8717, GUID 2791)
- Pridewing Soarer (ID 6141, GUID 32374)
- Westfall Woodworker (ID 6670, GUID 45521)
- Thomas Booker (ID 1437, GUID 9468)
- Gordo (ID 10666, GUID 28413)
- Gazelle (ID 4166, GUID 14049)
- Wooly Kodo (ID 3237, GUID 15144)
- Warpwood Tangler (ID 11464, GUID 300943)
- Grunt Gargal (ID 9086, GUID 6883)

Also watch creature formations without waypoint movement, e.g. Emeraldon Boughguard (GUID 33246, ID 12474).


src/game/Movement/WaypointMovementGenerator.cpp
```diff
@@ -181,64 +203,111 @@ bool WaypointMovementGenerator<Creature>::OnArrived(Creature& creature)
...
-    m_isArrivalDone = false;
     creature.AddUnitState(UNIT_STATE_ROAMING_MOVE);
 
-    WaypointNode const& nextNode = currPoint->second;
     Movement::MoveSplineInit init(creature, "WaypointMovementGenerator<Creature>::StartMove");
-
-    if (WaypointPath const* pSubpath = (nextNode.path_id ? sWaypointMgr.GetPathFromOrigin(nextNode.path_id, 0, 0, PATH_FROM_SPECIAL) : nullptr))
+    PointsArray genPath; // will contain the generated path
+    genPath.reserve(20); // little optimization
+    genPath.push_back(Vector3(creature.GetPositionX(), creature.GetPositionY(), creature.GetPositionZ()));
```
Gamemechanicwow: Consider using emplace_back when applicable. Here it avoids an unnecessary copy.
```suggestion
    genPath.emplace_back(creature.GetPositionX(), creature.GetPositionY(), creature.GetPositionZ());
```

src/game/Movement/WaypointMovementGenerator.cpp
```diff
@@ -137,16 +158,17 @@ bool WaypointMovementGenerator<Creature>::OnArrived(Creature& creature)
...
-        return true;
-
     creature.ClearUnitState(UNIT_STATE_ROAMING_MOVE);
-    m_isArrivalDone = true;
 
     WaypointPath::const_iterator currPoint = i_path->find(i_currentNode);
-    MANGOS_ASSERT(currPoint != i_path->end());
+    if (currPoint == i_path->end())
+        return false;
+    else
```
Gamemechanicwow: No need for else after a return

src/game/Movement/WaypointMovementGenerator.cpp
```diff
@@ -181,64 +203,111 @@ bool WaypointMovementGenerator<Creature>::OnArrived(Creature& creature)
...
+            {
+                if (!m_repeating)
+                    break;
+                else
+                {
+                    currPoint = i_path->begin();
+                    ++loops;
+                }
+            }
+        }
```
Gamemechanicwow: If you avoid else after breaks, you can make the code a bit cleaner to look at.
```suggestion
        if (nextDelay)
            break;

        ++currPoint;

        if (currPoint == i_path->end()) 
        {
            if (!m_repeating)
                break;

            currPoint = i_path->begin();
            ++loops;
        }
```

ghost:
To be honest, I don't know if this is even worth to be merged. I installed it on my server and it looks better, but it makes the code more complicated and needs more calculations, so costs more performance, just for a small visual improvement.

mserajnik:
> To be honest, I don't know if this is even worth to be merged. I installed it on my server and it looks better, but it makes the code more complicated and needs more calculations, so costs more performance, just for a small visual improvement.

Since VMaNGOS tries to replicate Vanilla as closely as possible: if this is closer to the original blizz-like behavior than the current solution, it’s probably worth it, no?

ghost:
> Since VMaNGOS tries to replicate Vanilla as closely as possible: if this is closer to the original blizz-like behavior than the current solution, it’s probably worth it, no?

Yes, of course. There are probably also ways to improve performance, that's why I wrote that I'm not a C++ professional, just doing this as a hobby and wanted to share the stuff I played around with for the past few weeks. I also tested if there are any crashes by setting all creatures with WP movement active, preloading all maps and letting the server run for a few days.

ghost:
Thanks, Gamemechanic, for your feedback. I pushed a new commit with your suggestions.

Daribon:
Big improvement over current waypoint movement.

Here is before:
https://github.com/user-attachments/assets/14768b1f-64ba-4805-bf89-1c100940f9ce
And after:
https://github.com/user-attachments/assets/c4435f0c-2635-4409-9090-cd624c50437f

Do note that the NPC in above example is using custom waypoints, but it does a good job worst case scenario.
(I will add the sniffed waypoints when this is merged, since the animation would look ugly without this PR.)

NickTyrer:
Can this be reopened? Seemed like a good PR

PR #3019, opened 2025-04-24
Add bot skin generation

Wall-core:
Adds proper bot skin generation.

### DIFF
--- a/sql/migrations/20250612195802_world.sql
+++ b/sql/migrations/20250612195802_world.sql
@@ -1,0 +1,973 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20250612195802');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20250612195802');
+-- Add your query below.
+
+
+-- Azshara
+SET @GGUID = 20238;
+SET @PTEMPLATE = 2900;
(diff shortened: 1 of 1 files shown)


gajet5:
I think this way is easier.
https://github.com/vmangos/core/pull/3010

Wall-core:
that loop is unchecked and can theoretically loop infinitely. the best way to do it would be accessing CharSections dbc data from the cache but this is straightforward and does the job just as well.

gajet5:
I'm sorry, but how can this loop be unchecked?

It uses the verification function.
https://github.com/vmangos/core/blob/2dae0ef85799fc34c2f9b62a1df3008cc069ebd0/src/game/Objects/Player.cpp#L362
This verification function is used when creating a character.

How can this cycle run indefinitely?

Wall-core:
it randomly rolls a skin, attempts to validate it, and if it's not valid it re-runs the loop. it could fail to roll a valid skin infinitely, in theory.

gajet5:
It sounds almost impossible. 
The chance that the combination will be incorrect is small, only 1-2 values per certain race. 
The cycle will run no more than 10 times. 
This can be ignored.

Wall-core:
using a loop here isn't clean, it's unnecessary cycles when the data is static and can be hardcoded or pulled from the DBC storage. And there should never be any potential for infinite loops. Non-deterministic behavior leads to crashes.

gajet5:
Well, that's your right to think so.
I don't see anything wrong with reducing the amount of code.

Wall-core:
Well, if you're interested in being the one to fix this issue, then you should look at how DBC data is stored. All the valid combinations exist already but aren't currently accessed by the bot generator. Write a function that does this and PR it so I can close mine.

PR #3034, opened 2025-05-24
Fix missing loot for Jade

Bennylavaa:
## 🍰 Pullrequest
[Jade](https://www.wowhead.com/classic/npc=1063/jade) is missing its random green drop loot table, only has a chance to drop trash currently. 

I ended up checking the reference_loot_template for all the items wowhead shows and it looks like it is a perfect match to 30072 which also some other npcs around that same level share as well. 

### Proof
https://www.wowhead.com/classic/npc=1063/jade

### Issues

- None

### How2Test
- Kill Jade a few times :)

### Todo / Checklist

- [X] None


ratkosrb:
If you add up the drop chances for all the greens on Wowhead it comes up to around 40%. Why did you give it only a 1% chance?

Bennylavaa:
That's what everyone other npc has for this reference entry.

ratkosrb:
Other npcs are unrelated...

Bennylavaa:
I changed it to 39 as that is what the total added drop chance is for all the items added together.

PR #3054, opened 2025-05-31
Optimize GetContinentInstanceId

Gamemechanicwow:
## 🍰 Pullrequest

Optimizes GetContinentInstanceId() by precalculating boundaries needed to do early returns.

Example of existing lines that are used to calculated ContinentInstanceId:
image
(Crude mock up. Boundary lines might be off)

Image of orc/troll starting zone, with its own continentinstanceid - show with lines that indicate boundaries and their early return values
image

Test show high potential for performance gains
image

**Test explained:**

- loaded all positions from creature table in map 0/1
- made a vector with all positions repeated 50 times
- timed old function
- timed new function
- after timing; a 2nd iteration was done verifying that the new and old implementation returned the same value for all coords in the vector 

**Notes:**
Could use better codeformatting. 
Maybe better naming of structs, vars etc.

Alternative solution
Remove struct and makeboundary - and just add the boundary as constants to function call:
            if (IsNorthTo(x, y, topNorthSouthBoundary, yyy, YYY, xxx, XXX))
                return MAP0_TOP_NORTH;
Upside: less code needed.
Downside: if someone is to edit limits, they also manually need to update the numbers in the function call

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [ ] Better code formatting
- [ ] Better naming
- [ ] Code comments

### DIFF
--- a/src/game/Maps/MapManager.cpp
+++ b/src/game/Maps/MapManager.cpp
@@ -644,7 +644,7 @@ static float const topNorthSouthLimit[] = { 2032.048340f, -6927.750000f,
                                             1577.923706f, 1293.419922f,
                                             1458.520264f, 1727.373291f,
                                             1591.916138f, 3728.139404f};
-static const InstantiatedContinentBoundary topNorthSouthBoundary(MakeBoundary(topNorthSouthLimit, sizeof(topNorthSouthLimit) / (2 * sizeof(float))));
+static InstantiatedContinentBoundary const topNorthSouthBoundary(MakeBoundary(topNorthSouthLimit, sizeof(topNorthSouthLimit) / (2 * sizeof(float))));
 
 static float const ironforgeAreaSouthLimit[] = {-5806.26f, -5864.11f,
                                                 -5989.37f, -4312.29f,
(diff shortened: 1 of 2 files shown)


ratkosrb:
Why move those to the header file? They are not used outside the file. No need to expose them. Also now if anyone edits the boundaries it will recompile most of the core because it's in a header.

Gamemechanicwow:
Ye, from that perspective I guess it makes sense to keep them in the .cpp file.

I have moved the constants to the .cpp file. Also marked this PR as a draft, since there might be an edgecase with the orgrimmar boundary. It's a small area that doesn't contain any creatures, so my initial test didn't catch it.

Gamemechanicwow:
- Added a better check for when a player is further north than the most northen point in the boundary line.
- All poly lines now go from east to west. The first Y coordinate must be less than the last Y coordinate. Added comment and assert to document this new restriction

Diagram showing the 3 types of early outs:

image

1 - Player is more south than the most southern point in the boundary line, so the player can't be in this area
2 - Player is to the east or west of the boundary.
3 - Player is above the boundary. We know that boundaries are checked from south to north, so we don't have to worry about the player being in a different boundary. So if the player position is between the starting and end position of the boundary he is in this instance.

src/game/Maps/MapManager.cpp
```diff
@@ -625,19 +625,218 @@ BattleGroundMap* MapManager::CreateBattleGroundMap(uint32 id, uint32 InstanceId,
...
+                                        -173.00f, -4064.00f,
+                                        -263.00f, -3836.00f,
+                                        -75.00f, -3742.00f,
+                                        737.00f, -3782.00f,
+                                        1184.00f, -3915.00f,
+                                        1564.00f, -3943.00f,
+                                        1762.00f, -3746.00f,
+                                        2225.00f, -3596.00f,
+                                        2755.00f, -3766.00f};
+static const InstantiatedContinentBoundary durotarSouthBoundary(MakeBoundary(durotarSouthLimit, sizeof(durotarSouthLimit) / (2 * sizeof(float))));
```
ratkosrb: east const

Gamemechanicwow: ye, I should have known 😅 It's fixed now

I made another diagram to remind myself
image

PR #3073, opened 2025-06-17
Add config for M2 line-of-sight

Wall-core:
Adds a config to allow M2 objects such as trees and rocks to provide line-of-sight.


kuramajssken001:
@Wall-core thank u

PR #3080, opened 2025-06-27
Update windows runner to windows-latest

coolzoom:
## 🍰 Pullrequest
windows 2019 environment will be obsolete soon.
ACE and TBB compile are tricky, try to stick with the version i tested

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -19,8 +19,8 @@ on:
 
 jobs:
   build:
-    runs-on: windows-2019
-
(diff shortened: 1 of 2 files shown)


mserajnik:
@coolzoom

Thanks for tackling this!

And good catch removing the unneeded* `-DBUILD_FOR_HOST_CPU=0` option.

<sub>* It doesn't matter if `-march=native`—which is what that option controls—is present or not here since that is specific to GCC and Clang and has no effect when MSVC is used.</sub>

PR #3095, opened 2025-07-23
Fix for BattleBotAI chasing target entered stealth

ivgrekow:
## 🍰 Pullrequest
Battlebots and player pets keep chasing victim target after it enters stealth mode and is not visible for them anymore. For example, if you are playing as a rogue on the battleground and aggro a bot, then upon successful entry into stealth mode the bot will see you regardless of the distance. The same behavior occurs with player pets. The pull request adds code that checks visibility of victim targets.

### How2Test

- Enter battleground as a rogue/cat druid
- Aggro some battlebot and try to leave combat state preparing to enter stealth (or immidiately enter stealth being out of combat when bot targets you)
- Enter stealth and ensure bot can't see/attack/target/chase you.

### DIFF
--- a/src/game/AI/PetAI.cpp
+++ b/src/game/AI/PetAI.cpp
@@ -84,11 +84,7 @@ bool PetAI::_needToStop() const
         }
     }
 
-    bool isVisible = false;
-    // Prevent creature pets from chasing target that is not visible for owner
-    if (pOwner)
-        isVisible = !m_creature->GetVictim()->IsVisibleForOrDetect(pOwner, pOwner, false);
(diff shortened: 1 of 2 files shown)


Wall-core:
It is blizzlike behavior for pets to chase targets who have entered stealth from what I recall

ivgrekow:
Just read few threads on Blizzard forum and can say... well, sometimes it is hard to tell the difference between bug and feature. Judging by the whining of rogues in the comments, pets could do this right up until the TBC update.

ratkosrb:
Remove the change for pets. This is not blizzlike.

ivgrekow:
Done. Check for `PetAI::_needToStop` is removed.

PR #3098, opened 2025-07-31
Pin Windows runner images to Windows Server 2022.

mserajnik:
`windows-latest` currently references Windows Server 2022, but [will be migrated to Windows Server 2025 in September](https://github.blog/changelog/2025-07-31-github-actions-new-apis-and-windows-latest-migration-notice/#windows-latest-image-label-migration).

Since using Windows Server 2025 will likely break things (yet again) without further changes to the workflow configurations, adjusting them to use `windows-2022` explicitly instead for now should keep things working (until someone finds the time to make the necessary adjustments to be able to use Windows Server 2025).

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -19,8 +19,8 @@ on:
 
 jobs:
   build:
-    runs-on: windows-latest
-    
+    runs-on: windows-2022
+
     steps:
     #git checkout
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


0blu:
Thx for doing that.
I saw the runner version change to `latest` in a recent PR and I figured that this piece of code would need to be adjusted eventually.
Pinning a specific version is really important if it comes down to consistently.

PR #3114, opened 2025-08-29
Anticheat: Wallclimb - optimization

Gamemechanicwow:
## 🍰 Pullrequest

**Calculation change** 
Instead of calculating atan every tick, we can eliminate it by calculating tan of the wallclimb angle once

Math proof:
atan(deltaZ / deltaXY) > sWorld.getConfig(CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE)

tan(atan(deltaZ / deltaXY)) > tan(sWorld.getConfig(CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE))

deltaZ / deltaXY > tan(sWorld.getConfig(CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE))

Now we multiply both sides with deltaXY

deltaZ > tan(sWorld.getConfig(CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE)) * deltaXY

**get map height change**
We have two calls to GetHeight(..) that calls GetDynamicTreeHeight(..) with the exact same parameters.
image

### Proof

- None

### Issues

- None

### How2Test
- I personally tested it by having both old and new version of the function being called each time. I then logged positions and what caused each function to return
- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Anticheat/MovementAnticheat/MovementAnticheat.cpp
+++ b/src/game/Anticheat/MovementAnticheat/MovementAnticheat.cpp
@@ -16,6 +16,9 @@
 
 using namespace Geometry;
 
+float MovementAnticheat::m_wallSlope = 0.f;
+float MovementAnticheat::m_wallSlopeHigh = 0.f;
+
 char const* GetMovementCheatName(CheatType flagId)
 {
     switch (flagId)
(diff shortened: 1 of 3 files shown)


src/game/Anticheat/MovementAnticheat/MovementAnticheat.cpp
```diff
@@ -1158,20 +1168,27 @@ bool MovementAnticheat::CheckWallClimb(MovementInfo const& movementInfo, uint16
...
 
-    float const maxClimbAngle = sWorld.getConfig(CONFIG_FLOAT_AC_MOVEMENT_CHEAT_WALL_CLIMB_ANGLE);
-    if (angleRad > maxClimbAngle)
+    if (deltaZ > m_wallSlope * deltaXY)
     {
-        if (angleRad > (maxClimbAngle + 0.2f))
+        if (deltaZ > m_wallSlopeHigh * deltaXY)
             return true;
 
+        ASSERT(MaNGOS::IsValidMapCoord(movementInfo.pos.x, movementInfo.pos.y, movementInfo.pos.z));
```
ratkosrb: Assert in function called from packet handlers? Won't this allow players to crash the server if they craft a packet with invalid coordinates?

Gamemechanicwow: I added it because the assert was in GetHeight(..). But i see now its not needed at all. All paths to the function are guarded by VerifyMovementInfo(..)

PR #3122, opened 2025-09-12
mmaps: fix some memory leaks and refactor code for async tile processing

schell244:
## 🍰 Pullrequest
- refactor code for async tile processing
- fixed some memory leaks (found by AI)
- no changes which impact the resulting mmaps, just process improvements

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None


contrib/mmap/src/TerrainBuilder.cpp
```diff
@@ -170,8 +182,18 @@ namespace MMAP
...
+                {
+                    delete[] m_V8;
+                    m_V8 = new float[V8_SIZE_SQ];
+                }
                 if (!m_V9)
                     m_V9 = new float[V9_SIZE_SQ];
+                else
+                {
+                    delete[] m_V9;
+                    m_V9 = new float[V9_SIZE_SQ];
```
ratkosrb: Why do you delete it and reallocate it when its already allocated?

PR #1705, opened 2022-11-21
Implement Book Soothsaying for dummies + Profession fixes + Trainer overhaul

FlagFlayer:
## 🍰 Pullrequest

This pull request implements the intended ways to change profession specialisations, as well as generally reworking profession handling to use cmangos implementation.

## Server hosts be warned: merging this PR will cause respecialisation to be broken on old characters

The blizz-like implementation uses hidden "specialisation reputation" (details below) that would not have been rewarded to characters that have previously completed specialisation quests. Once this PR is ready for merge there will be another PR linked here that uses quest completion instead of the reputation to prevent any bugs.

### Remove QUEST STATUS NONE update for abandoned professions

* Fixes https://github.com/vmangos/core/issues/1681

* Allows for implementation of Book Soothsaying for dummies

### Implement Book Soothsaying for dummies for Engineering and Leatherworking specialisations

* Fixes https://github.com/vmangos/core/issues/1685
[Added patch 1.10](https://warcraft.wiki.gg/wiki/Patch_1.10.0), before this there was no way to switch specialisations for Engineering and Leatherworking.

> Professions
> Players that unlearn Engineering or Leatherworking may now respecialize if they decide to relearn the professions.

[Video from original WotLK](https://www.youtube.com/watch?v=9ffNV2ZRI7M)

* [Adds relevant gossip to trainer NPCs](https://www.youtube.com/watch?v=u9oOnczxaW8&t=65s)

### Remove profession specialisation ranks in spell_chain

Implied by SkillLineAbility.dbc: There's no previous rank for profession specialisations. They are not part of the profession spell chain.

* Fixes https://github.com/vmangos/core/issues/1686

* Fixes https://github.com/vmangos/core/issues/1683

> From Thottbot By naiad [on 2005/10/27](https://www.wowhead.com/classic/quest=5143/tribal-leatherworking#comments:id=2738463) (Patch 1.8.1)		
> Subject: "225->300 (artisan)"
> there is NO correlation between tribal (or any of the three specs) and getting your artisan (225 to 300 max) leatherworking.  to learn artisan lwing, you will need to go to the master leatherworking trainer.  for the alliance he is in aerie peak, in the inn downstairs.  you must be level 35 and have 200+ lwing.  it will cost you 5 gold.
> 
> for the horde, i believe you have to go to feralas (camp mojache).

[...]

PR #3128, opened 2025-09-26
Remove trainer_... fields from creature_template and move trainerType to DB

FlagFlayer:
### Main changes:

* Remove deprecated trainer_... fields from creature_template (except for trainer_id, not all trainer_ids are sniffed).

* Move trainerType assignment to DB and evaluate based on sniffed values.

* Handle trainer gossip options via conditions in `gossip_menu_option`. All missing gossip options and conditions have been added and applied.

### Miscellaneous

* Clean up deprecated code: "msg" field in IsTrainerOf and CanInteractWithBattleMaster (has no function)

### How2Test

- Talk to a trainer. Trainer gossip option appears when the correct conditions are met.
- Use `.learn all_trainer`. It correctly teaches all available trainer spells.

### Todo / Checklist

- [X] Remove relevant fields from `creature_template` and replace code that is reliant on them
- [X] Move trainerType assignment to DB (trainer greeting type)
- [X] Adjust the Core to work with the new DB assignment
- [X] Reimplement `.learn all_trainer` command in the new environment


ratkosrb:
What happened with the Soothsaying fixes, did that implementation require these changes first?

mserajnik:
> What happened with the Soothsaying fixes, did that implementation require these changes first?

I think he just decided to split up the PR, see his comment here: https://github.com/vmangos/core/pull/1705#issuecomment-2699672335

FlagFlayer:
> What happened with the Soothsaying fixes, did that implementation require these changes first?

They did not require this, but since this part of the fixes makes the `trainer_spell` column become completely empty I bundled these fixes with this change. 

What I'm missing for the Soothsaying fixes is sniffs of the gossip submenus when going through the option "I would like to relearn my leatherworking focus", as well as the correct order of entries when talking to Book Soothsaying for Dummies and having both the leatherworking and the engineering options available. I haven't been able to get around to getting these sniffs, I hope someone else might be able to obtain them at some point.

FlagFlayer:
Ready for merge

sql/migrations/20260825234315_world.sql
```diff
@@ -0,0 +1,568 @@
...
+11178);
+
+-- Velma Warnam has a non-empty trainer greeting (already correct in other localisations)
+UPDATE `npc_trainer_greeting` SET `content_default` = 'Hello!  Can I teach you something?' WHERE `entry` = 4773;
+
+-- Add 'type' to npc_trainer_greeting instead of calculating in core:
+-- This is a formatter for training UI that is opened after sending SMSG_TRAINER_LIST to the client (completely unrelated to trainer_type from creature_template)
+-- Stored in DB together with trainer greetings, valid values are 0 and 2
+ALTER TABLE `npc_trainer_greeting`
+	ADD COLUMN `trainerType` TINYINT(2) UNSIGNED NOT NULL DEFAULT '2' AFTER `entry`;
```
ratkosrb: we use underscore naming style in the db

sql/migrations/20260825234315_world.sql
```diff
@@ -0,0 +1,568 @@
...
+5040,
+5041);
+
+-- from `npc_trainer`
+DELETE FROM `npc_trainer` WHERE `entry` IN (
+5032,
+5037,
+5038,
+5040,
+5041);
```
ratkosrb: These are not custom npcs, Given their name it's safe to assume they are trainers for the listed professions. They should have appropriate training data.

FlagFlayer: I'll remove this from this PR 👍

My point here was that these trainers are ancient code from alpha. Consider that there's World Trainers for [Cartography](https://www.wowhead.com/classic/npc=5035/world-cartography-trainer) and [Tracking](https://www.wowhead.com/classic/npc=5039/world-tracking-trainer). I don't know if it's possible to pinpoint what sort of trainer spell list they'd have had (assuming they were functional trainers), but the current list and greetings are custom-picked.

PR #3133, opened 2025-10-26
Fix floating-point drift in PLAYER_FIELD_MOD_DAMAGE_DONE_PCT modifiers

Onixaz:
## 🍰Pullrequest

This PR fixes floating-point drift in the per-school damage done percentage modifiers (PLAYER_FIELD_MOD_DAMAGE_DONE_PCT[school]) when modified via **Object::ApplyModSignedFloatValue**.
These fields are 1.0-based multipliers. When temporary effects repeatedly add and remove small percentage values (e.g., talents, consumables, procs), floating-point rounding error can accumulate and cause the stored value to drift away from the expected neutral baseline of 1.0f.

This leads to situations where, after all modifiers are removed, the final value may be slightly below or above 1.0 (e.g., 0.99999994 instead of 1.0).
This results in tiny unintended buffs/nerfs and, more visibly, **client sheet desync:** the character sheet may display “100%” in red, indicating a hidden modifier still being applied.

Seen in: #1236 

This PR normalizes values that end up within 2 ULPs of 1.0f back to exactly 1.0f, and prevents NaN/Inf propagation.
The correction is only applied to the known 1.0-based array range.

### Proof
Reproducible by repeatedly applying/removing small damage % auras and observing drift from 1.0 in PLAYER_FIELD_MOD_DAMAGE_DONE_PCT. 

### Issues
None (fix is self-contained). Primarily resolves a correctness and UI desync issue.

### How2Test

1. Create a Warrior with **Enrage (Rank 1)** talented.
2. Trigger Enrage (e.g., take a critical hit).
   - Character sheet shows **+5% damage** in green.
3. Cancel Enrage.
   - **Before the fix:** damage returns to 100% but appears **red**, indicating hidden modifier drift.
   - **After the fix:** damage returns to **clean white 100%**, with no hidden modifier remaining.

### Todo / Checklist
None

### DIFF
--- a/src/game/Objects/Object.cpp
+++ b/src/game/Objects/Object.cpp
@@ -1232,9 +1232,24 @@ void Object::ApplyModInt32Value(uint16 index, int32 val, bool apply)
 
 void Object::ApplyModSignedFloatValue(uint16 index, float  val, bool apply)
 {
-    float cur = GetFloatValue(index);
-    cur += (apply ? val : -val);
-    SetFloatValue(index, cur);
+    const float cur = GetFloatValue(index);
+    const float delta = apply ? val : -val;
+    float next = cur + delta;
+
+    // Normalize floating-point drift for per-school damage modifiers
+    if (index >= PLAYER_FIELD_MOD_DAMAGE_DONE_PCT &&
+        index <  PLAYER_FIELD_MOD_DAMAGE_DONE_PCT + MAX_SPELL_SCHOOL)
+    {
+        // Within 2 ULPs of 1.0 (FLT_EPSILON is 1 ULP near 1.0)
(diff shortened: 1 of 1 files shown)


killerwife:
Tons of fields never go back to 1.f for players, and this is a systemic issue across many fields.

Likely different "fix" is that we currently do not recalculate everything on teleporting to a diff map and thats what instead should be done, vs reference servers. (since on official its going across nodes)

Onixaz:
@killerwife yeah I would suspect thats the case. The reason I only changed this particular one is because its rather easy to test and very obvious one, as client screams at you: "Hey something is wrong with calculations on your server"

killerwife:
Mhm, anyway, up to brot what he dors with this

PR #3141, opened 2025-10-30
Fix: Prevent insta-kill spell from crashing client on pet

DealsBeam:
The EffectInstaKill function did not properly handle cases where a player casts an insta-kill spell on their own pet. The existing check only prevented self-casting, leading to a client crash when a player targeted their own pet.

This commit extends the check to also prevent a player from insta-killing their own pet, resolving the crash. A test script has been added to verify the fix.

## 🍰 Pullrequest
This pull request fixes a critical bug that causes the game client to crash when a player uses an insta-kill spell on their own pet. The fix is a simple and targeted change to the `EffectInstaKill` function in `src/game/Spells/SpellEffects.cpp`.

### Proof
The fix has been verified by a custom spell script located in `src/scripts/spells/spell_test.cpp`. This script can be attached to an in-game spell to test the fix. When the spell is cast on the caster's pet, it will log a "pass" message to the server console if the pet survives, and a "fail" message if it does not.

### Issues
- None

### How2Test
1.  Apply this patch and rebuild the server.
2.  Create a new spell in the database and attach the `spell_insta_kill_pet_test` script to it.
3.  Log into the game with a character that has a pet.
4.  Learn the test spell.
5.  Cast the test spell on your pet.
6.  Check the server console for the "Insta-kill pet test passed." message.

### Todo / Checklist
- [X] Find and fix a verifiable bug.
- [X] Provide a detailed bug report.
- [X] Implement a targeted fix.
- [X] Verify the fix through testing.

### DIFF
--- a/src/game/Spells/SpellEffects.cpp
+++ b/src/game/Spells/SpellEffects.cpp
@@ -267,8 +267,11 @@ void Spell::EffectInstaKill(SpellEffectIndex /*effIdx*/)
     if (!unitTarget || !unitTarget->IsAlive())
         return;
 
-    if (m_caster == unitTarget)                             // prevent interrupt message
+    if (m_caster == unitTarget || (unitTarget->IsPet() && unitTarget->ToPet()->GetOwner() == m_caster))
+    {
         finish();
+        return;
+    }
 
 #if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_11_2
     WorldPacket data(SMSG_SPELLINSTAKILLLOG, (8 + 4));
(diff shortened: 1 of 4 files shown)


Wall-core:
this will cause a crash if `unitTarget->ToPet()` is null. This should be checked first, and ToPet() checks IsPet() inherently.

```
    if (m_caster == unitTarget || (unitTarget->ToPet() && unitTarget->ToPet()->GetOwner() == m_caster))
    {
        finish();
        return;
    }
```

Daribon:
AI slop.

ratkosrb:
https://github.com/user-attachments/assets/2c0edb49-18e4-4cc4-b86a-188ec4df76fe

There is no crash when killing your own pet. This change is unnecessary.

PR #3145, opened 2025-11-04
Fix migration file name.

### LINKED ISSUE #3144: 🐛 [Bug] wrong id in latest commit 0c990fe13c26907b4ee658513d9495de10f048ac
## 🐛 Bug report
mangos failed to launch due to the wrong migration id
"Database `mangos` is missing the following migrations:
	20251103125545
"

the sql file name in commit 0c990fe13c26907b4ee658513d9495de10f048ac is  20251103125545 :
`sql/migrations/20251103125545_world.sql`

but the id insert into migrations table is 20251103132426
`SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20251103132426');`

20251103132426 is not same be 20251103125545.

These should be same.

mserajnik:
## 🍰 Pullrequest
This fixes the incorrect file name of the migration introduced in https://github.com/vmangos/core/commit/0c990fe13c26907b4ee658513d9495de10f048ac which currently prevents the server from starting:
```
Database `mangos` is missing the following migrations:
  20251103125545
```

### Proof

- None

### Issues

- Fixes #3144

### How2Test

- None

### Todo / Checklist

- [X] None


Daribon:
My bad, it was due to the very odd encoding error I had to fix.

PR #3149, opened 2025-11-18
Crashfix with underwater giant NavMesh

Wall-core:
Fixes a dereferenced unit pointer crash


ratkosrb:
Did you actually get a crash here? Cause if IsCreature is true then pUnit can't be null, because all creatures are units, it's their base class.

Wall-core:
frankly I'm not sure, I'm helping debug AshenWoW and don't have a stack trace but there's a crash occurring somewhere in findRandomPointAroundCircle, according to the crashlog:

image

so I just browsed through these associated functions and my visual studio was throwing a warning for dereferenced null pointer, which I dealt with here.

I'll be helping them rebuild with gdb this week so I can get an actual stack trace, but if you want to investigate further, here is an additional crashlog (same problem):

image

ratkosrb:
Oh I know what causes this. You need to set CleanupTerrain to 0.

PR #3152, opened 2025-11-19
Fix parsing of a log-related parameter

micttyoid:
## 🍰 Pullrequest

In the mangosd configuration, the parameter`GmCriticalCommands` has been parsed as `CriticalCommands`. The following commit fixes it.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/mangosd/mangosd.conf.dist.in
+++ b/src/mangosd/mangosd.conf.dist.in
@@ -588,7 +588,7 @@ DebuffLimit = 0
 #        GM Log file of gm commands
 #        Default: "" (Disable)
 #
-#    LogFile.CriticalCommands
+#    LogFile.GmCriticalCommands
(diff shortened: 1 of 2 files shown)


ratkosrb:
Why does it have to be renamed if it has the same name in the cpp file as in the conf file?

mserajnik:
> Why does it have to be renamed if it has the same name in the cpp file as in the conf file?

Hm, no, he is right, it doesn’t: https://github.com/vmangos/core/blob/abbb1a8d0c51f13a66a12c8043d4fa9ec51c0d9c/src/mangosd/mangosd.conf.dist.in#L690

It is called `CriticalCommands` in the comment (which he renames), but the actual setting is called `GmCriticalCommands`.

PR #3170, opened 2025-12-22
Implement Nefarians AoE version of Shadow Flame

schell244:
## 🍰 Pullrequest

Implement special version of spell Shadow Flame:
When Nefarian lands at the start of Phase 2 of his encounter he will use an AoE Shadowflame ~~that ignores LoS~~, this spell does about 1000 initial shadow damage, but applies a deadly DoT if an Onyxia Scale Cloak is not equipped.

Also fixed one missplaced text in Razorgores script.

### Proof
- None

### Issues
- None

### How2Test
- Start Nefarian encounter, progress till phase 2 untill he lands, do it one time with Onyxia Clock equipped and one time without.

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20251222091514_world.sql
+++ b/sql/migrations/20251222091514_world.sql
@@ -1,0 +1,32 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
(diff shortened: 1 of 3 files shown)


Daribon:
Does shadow flame hit you behind the throne with this applied, without onyxia scale cloak?

schell244:
Just checked it, behind throne players don't get hit by the spell.

Daribon:
> Just checked it, behind throne players don't get hit by the spell.

Good, then that part is working as intended. So many times people on classic would forget about cloak and then rush to get behind throne when they realized.

PR #3171, opened 2025-12-22
Fix problems spotted by code analysis

schell244:
### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Objects/Creature.cpp
+++ b/src/game/Objects/Creature.cpp
@@ -111,7 +111,7 @@ bool AssistDelayEvent::Execute(uint64 /*e_time*/, uint32 /*p_time*/)
                     // When nearby mobs aggro from another mob's initial call for assistance
                     // their leash timers become linked and attacking one will keep the rest from evading.
                     if (assistant->GetVictim())
-                        assistant->SetLastLeashExtensionTimePtr(m_owner.ToCreature()->GetLastLeashExtensionTimePtr());
+                        assistant->SetLastLeashExtensionTimePtr(static_cast<Creature*>(&m_owner)->GetLastLeashExtensionTimePtr());
                 }
             }
         }
--- a/src/game/Objects/Creature.cpp
+++ b/src/game/Objects/Creature.cpp
@@ -111,7 +111,7 @@ bool AssistDelayEvent::Execute(uint64 /*e_time*/, uint32 /*p_time*/)
                     // When nearby mobs aggro from another mob's initial call for assistance
                     // their leash timers become linked and attacking one will keep the rest from evading.
                     if (assistant->GetVictim())
-                        assistant->SetLastLeashExtensionTimePtr(static_cast<Creature*>(&m_owner)->GetLastLeashExtensionTimePtr());
+                        assistant->SetLastLeashExtensionTimePtr(m_owner.ToCreature()->GetLastLeashExtensionTimePtr());
                 }
             }
         }
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -5821,21 +5821,21 @@ bool Unit::IsImmuneToSpellEffect(SpellEntry const* spellInfo, SpellEffectIndex i
             }
         }
 
-        if (mechanic == MECHANIC_NONE)
-            return false;
-
-        AuraList const& immuneAuraApply = GetAurasByType(SPELL_AURA_MECHANIC_IMMUNITY_MASK);
-        for (const auto& iter : immuneAuraApply)
+        if (mechanic != MECHANIC_NONE)
         {
-            if (iter->GetModifier()->m_miscvalue & (1 << (mechanic - 1)))
+            AuraList const& immuneAuraApply = GetAurasByType(SPELL_AURA_MECHANIC_IMMUNITY_MASK);
+            for (const auto& iter : immuneAuraApply)
             {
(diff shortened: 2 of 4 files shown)


src/game/Objects/Creature.cpp
```diff
@@ -111,7 +111,7 @@ bool AssistDelayEvent::Execute(uint64 /*e_time*/, uint32 /*p_time*/)
                     // When nearby mobs aggro from another mob's initial call for assistance
                     // their leash timers become linked and attacking one will keep the rest from evading.
                     if (assistant->GetVictim())
-                        assistant->SetLastLeashExtensionTimePtr(static_cast<Creature*>(&m_owner)->GetLastLeashExtensionTimePtr());
+                        assistant->SetLastLeashExtensionTimePtr(m_owner.ToCreature()->GetLastLeashExtensionTimePtr());
```
ratkosrb: why change this? we don't need a type check here

schell244: Just for readability reasons 😅

src/game/Objects/Unit.cpp
```diff
@@ -737,7 +737,7 @@ uint32 Unit::DealDamage(Unit* pVictim, uint32 damage, CleanDamage const* cleanDa
     uint32 health = pVictim->GetHealth();
     // duel ends when player has 1 or less hp
     bool duel_hasEnded = false;
-    if (pVictim->IsPlayer() && ((Player*)pVictim)->m_duel && damage >= (health - 1))
+    if (pVictim->IsPlayer() && ((Player*)pVictim)->m_duel && health > 0 && damage >= (health - 1))
```
ratkosrb: why prevent duel from ending if victim died? instead just handle the 0 hp case by setting damage also to 0

src/game/Objects/Unit.cpp
```diff
@@ -1612,6 +1612,8 @@ void Unit::DealMeleeDamage(CalcDamageInfo const* damageInfo, bool durabilityLoss
             else if (offtime > percent60)
             {
                 offtime -= 2.0f * percent20;
+                if (offtime < 0.0f)
+                    offtime = 0.0f;
```
ratkosrb, line 1616: can it really go below 0 given that the value is checked above that

src/game/Objects/Unit.cpp
```diff
@@ -5739,6 +5743,9 @@ bool Unit::IsImmuneToSpell(SpellEntry const* spellInfo, bool /*castOnSelf*/) con
             }
         }
 
+         if (mechanic == MECHANIC_NONE)
```
ratkosrb: put the part below inside the if instead of returning

src/game/Objects/Unit.cpp
schell244: If health is 0, then health - 1 wraps to 4294967295 as it's a uint

src/game/Objects/Unit.cpp
schell244: Floating-point precision can cause offtime to become slightly negative near the boundary, even when offtime > percent60 should prevent it.

src/game/Objects/Creature.cpp
ratkosrb: i don't like it, static cast is perfectly readable, do not introduce needless type checks please

src/game/Objects/Unit.cpp
```diff
@@ -5814,6 +5821,9 @@ bool Unit::IsImmuneToSpellEffect(SpellEntry const* spellInfo, SpellEffectIndex i
             }
         }
 
+        if (mechanic == MECHANIC_NONE)
+            return false;
```
ratkosrb: here too same as above

src/game/Objects/Unit.cpp
ratkosrb: this is what i suggest
```
    if (pVictim->IsPlayer() && ((Player*)pVictim)->m_duel && (damage + 1) >= health)
    {
        // prevent kill only if killed in duel and killed by opponent or opponent controlled creature
        if (((Player*)pVictim)->m_duel->opponent == this ||
            ((Player*)pVictim)->m_duel->opponent->GetObjectGuid() == GetOwnerGuid()
            // World of Warcraft Client Patch 1.7.0 (2005-09-13)
            // - Fixed bug where you could kill someone in a duel with spell reflection.
#if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_6_1
            || pVictim == this && reflected
#endif
            )
            damage = health ? health - 1 : 0;

        duel_hasEnded = true;
    }
```

src/game/Objects/Creature.cpp
schell244: but then why static_cast<Creature*>(&m_owner) when we can also use (Creature*)(&m_owner) so we safe another check! 😄

Gamemechanicwow: static_cast doesnt do a runtime check. We know its always a creature, thats why static. 

static_cast also have the benefit of readability, we know the intent. So does the compiler; a static_cast can produce an error at compile time.

schell244: Was meant to be a joke😅 - I simply like how expressive m_owner.ToCreature() is, even though it hides a additional typeId check. My brain got bendy from too much bad readable code in the past.

PR #3180, opened 2025-12-27
Converted the logic for Warlock, Rogue, and Mage class-call spells to use spell scripts

schell244:
## 🍰 Pullrequest
Converted the logic for Warlock, Rogue, and Mage spells.

Changes also include:
- Nefarian doesn't consider GM's class for class calls (if this class is not in the raid group facing him).
- Rogue spell: use GetFirstCollisionPosition to avoid teleporting the player into walls/obstacles
- Mage spell: iterate through the ~~group~~ map players instead of Nefarian's threat list for target selection (no need to check for totems, pets, ...)

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/sql/migrations/20251227092730_world.sql
+++ b/sql/migrations/20251227092730_world.sql
@@ -1,0 +1,20 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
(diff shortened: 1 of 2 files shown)


ratkosrb:
Doesn't this allow bypassing the mechanic if you drop group right before?

schell244:
Oh wow, did not think about that👀

Issue #2858, opened 2025-01-01
🐛 [Bug]Massive pathing issue


ShiyoKozuki:
I press charge on an enemy, it paths me around the boat for 15 seconds.
Pets also path incorrectly through this boat, and so do enemies. They both won't path through the narrow little "broken" part of the boat and go all the way around like my charge did.

https://drive.google.com/file/d/1HzxE9-UU68A2Ji9WmfFoPNY1QUqoTxKC/view?usp=sharing

### Version & Environment

Client Version:  1.12.1.5875

mserajnik:
> I press charge on an enemy, it paths me around the boat for 15 seconds.

I can reproduce this issue with Charge using the latest commit (https://github.com/vmangos/core/commit/41b78fe791d1cef0696414dfe0100e73541a6ec9 at the time of writing) and freshly extracted client data: https://files.catbox.moe/htxt6y.mp4

> Pets also path incorrectly through this boat, and so do enemies. They both won't path through the narrow little "broken" part of the boat and go all the way around like my charge did.

This seems to be a bit inconsistent: https://files.catbox.moe/nvntdb.mp4

It looks like that sometimes mobs manage to path through that gap fine, other times they try to go around (and then clip partially through the ship). I didn't try in regards to pets (yet).

Issue #2549, opened 2024-03-15
🐛 [Bug]Navmesh / Pathing error spot inside of RFD


ShiyoKozuki:
## 🐛 Bug report

![image](https://github.com/vmangos/core/assets/58101246/f8248e2d-7ef0-43e0-93e0-f47d1b84985f)
Video: (Too big to upload here):
https://drive.google.com/file/d/1Nn0ELfJVtmynZm_rJJO9JuaFRqWkNN_P/view?usp=sharing
Mobs get stuck here and can't properly path to players.

### Expected behavior
Should properly path.

### Version & Environment

Client Version:  1.12.1

ratkosrb:
Is this with the repack mmaps or ones you compiled yourself?

mserajnik:
Is this the same issue as #2356 (also opened by you) or does this behave differently?

Looks like the video sharing settings are wrong, or at least I can't view it (logged out or logged in):
<img width="400" alt="image" src="https://github.com/vmangos/core/assets/80664890/5168b34d-db09-4841-b55e-39c81f38d2b0">

ShiyoKozuki:
> Is this with the repack mmaps or ones you compiled yourself?

Bropack maps

ShiyoKozuki:
> Is this the same issue as #2356 (also opened by you) or does this behave differently?
> 
[...]

Issue #2516, opened 2024-02-17
🦂 [Bug] Razorfen Downs. The creatures can't find their way to. Map error.


gajet5:
## 🐛 Bug report
https://www.youtube.com/watch?v=KJGkiU8dT_E

![image](https://github.com/vmangos/core/assets/3825412/009df268-560f-4343-b945-9eb544f8f969)
0:10 video

![image](https://github.com/vmangos/core/assets/3825412/4980be9f-4698-4260-b5c3-98a37dc6c7d7)
2:20 video

![image](https://github.com/vmangos/core/assets/3825412/09174668-33ea-44ba-8a11-256c32a80a2a)
5:09 video

![image](https://github.com/vmangos/core/assets/3825412/a5ffa80c-4e69-4e8a-9a0e-c2d89aff5f8d)
5:43 video

**The maps have been extracted from the client: 31.01.2024**

### Expected behavior
Creatures find their way.

### Steps to reproduce
1. go to Razorfen Downs
2. Profit

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  No

OS Client: Win
OS Server: Ubuntu

### Crashlog

- None

ratkosrb:
Have you tried with the repack mmaps?

gajet5:
I repackaged the map on January 31, as far as I can see in the commitsthere was no work with the map extractor.

mserajnik:
Just as a quick note (I didn't want to create a new issue for this):

The Azurelode Mine in Hillsbrad Foothills has a very similar problem where mobs can't path over a spot (let's call it an "invisible wall" for reading clarity) near the main entrance.

Instead, if they are aggroed and on the opposite side of the invisible wall compared to the player, they will run through and out of the cave via the entrance/exit on their side of the invisible wall and then back inside through the other cave entrance to get to the player; this seems to behave the same in both directions.

This is with freshly extracted mmaps.

PR #3192, opened 2026-01-02
Improvements for extraction tools (vmap/mmap)

schell244:
## 🍰 Pullrequest
The extractor was logging warnings for models referenced in DBC files that don't actually exist in MPQ archives.
Added `MPQFileExists()` check before attempting to open model files, so only actual extraction failures are reported. 

Also removed the misleading "A few of these warnings are expected" 
message since warnings now indicate real problems.

- Add MPQFileExists to prevent false warning messages
- Improve single tile mmap building
- Add fallback logic for model paths with space/underscore variants
- ~~Add interactive prompts for vmap resolution~~
- Add "Press enter to close" prompts to prevent console from closing on error
- fix tileY,tileX filename order in IntermediateValues.cpp (revert previous change)
- buildSingleTile now always rebuilds the tile (overwrite existing)
- improve `.mmap connect` command to record offmesh entries
- added new offmesh to fix path between Mr. Smite bridge <-> ship

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None


mserajnik:
> Add interactive prompts for vmap resolution
> Add "Press enter to close" prompts to prevent console from closing on error

Will this still allow for running the extraction unsupervised/without manual input? That's important to me for scripting purposes. 😄

contrib/vmap_extractor/vmapextract/vmapexport.cpp
```diff
@@ -431,19 +431,6 @@ int main(int argc, char** argv)
...
-    std::getline(std::cin, userInput);
-    highRes = userInput.empty() || userInput.compare("y") == 0;
-    std::cout << "High resolution = " << highRes << std::endl;
-    std::cout << "Press enter to start extracting vmaps." << std::endl;
-    std::cout << "=====================================" << std::endl;
-    std::cin.get();
-
-    // Overwrite due to user input
-    preciseVectorData = highRes;
-
```
mserajnik: If you want to keep the prompts, to me it would also be fine to introduce a flag `--non-interactive` or something instead to be able to skip them.

schell244: Thanks for pointing that out. I might just open a second PR, adding promts for both maps/vmaps similar to how it's currently done in the mmap tool 👍 (diff is already big enough here 😅)

schell244: Would be very much appreciated if you want to give it a try 😇. A good point to check is `.go c Glutton` and see if he manages to walk through his 'narrow passage' – when Razorfen Leanto03 was broken, he just disappeared in the walls instead.

mserajnik: I'm unfortunately on vacation and don't really have the time to test it, but I'll gladly give it a spin when I'm back next week!

schell244: Thanks and no worries, enjoy your vacation!🎉

mserajnik:
Nice work with the Deadmines fix! Looking forward to give that a try.

Do you happen to have an idea how any of the following other pathing bugs could be fixed?
I assume offmesh connections won't be the solution in these cases.

- [Seemingly invisible walls in RFD](https://github.com/vmangos/core/issues/2516) (duplicate issue exists [here](https://github.com/vmangos/core/issues/2549))
- [Seemingly invisible wall in Azurelode Mine](https://github.com/vmangos/core/issues/2516#issuecomment-3440207582) (my comment, didn't want to create a new issue)
- [Pathing issue with a shipwreck in Wetlands](https://github.com/vmangos/core/issues/2858)

schell244:
"RFD invisible wall issue" could already be fixed, since the patch that fixed parsing of “razorfen leanto03”, but have not checked that certain spot yet.

Using offmesh connections is the simplest way of patching up broken mmap spots without having to touch the overall logic. You can "force" a path between two points, which you declare in `offmesh.txt`.
The easiest way to record it is using `.mmap connect` (records point 1), then move to the point where you want to create a connection, run `.mmap connect` again, and it will create an offmesh entry in a file called `offmesh_connections.txt`, which you can then copy to the actual offmesh file that the mmap generator reads from. Afterwards just rebuild the tile where you added the path patch.

A useful command to check NPC pathing between you and another NPC is `.mmap path` - it will spawn those blue spirit waypoint NPCs along the path from your character to the selected NPC. If there is a "hole" or problem in between, the spawned waypoints will show that. Also what works well is spawning a pet and watching if it takes weird extra walks at certain spots.

ShiyoKozuki:
> Nice work with the Deadmines fix! Looking forward to give that a try.
> 
> Do you happen to have an idea how any of the following other pathing bugs could be fixed? I assume offmesh connections won't be the solution in these cases.
> 
>     * [Seemingly invisible walls in RFD](https://github.com/vmangos/core/issues/2516) (duplicate issue exists [here](https://github.com/vmangos/core/issues/2549))
> 
>     * [Seemingly invisible wall in Azurelode Mine](https://github.com/vmangos/core/issues/2516#issuecomment-3440207582) (my comment, didn't want to create a new issue)
> 
>     * [Pathing issue with a shipwreck in Wetlands](https://github.com/vmangos/core/issues/2858)

Gnomergan also has pathing issues in the big room on the elevator after Crowd Pummeler 9-60. The big room that has the elevator, but at the bottom level.

contrib/mmap/src/MapBuilder.cpp
```diff
@@ -147,7 +147,11 @@ namespace MMAP
         m_cancel.store(false);
         buildMap(mapID);
         processQueuedTiles();
-        printf("Done.");
+
+        char mapFileName[255];
+        sprintf(mapFileName, "mmaps/%03u.mmap", mapID);
```
ratkosrb: why are you formatting this separately and not in the printf

contrib/mmap/src/MapBuilder.cpp
```diff
@@ -260,32 +266,25 @@ namespace MMAP
...
-
-        worker->WaitCompletion();
-
         dtFreeNavMesh(navMesh);
 
-        printf("Building single tile finished.");
+        processQueuedTiles();
+
+        char outputFileName[255];
+        sprintf(outputFileName, "mmaps/%03u%02u%02u.mmtile", mapID, tileY, tileX);
```
ratkosrb: again why not do this inside the printf

PR #3194, opened 2026-01-09
Don't include time in realmd console output by default.

mserajnik:
## 🍰 Pullrequest

This default value got changed via 28818aabd89c8d34df696edc7e49808865fffe97 and now doesn't match the default in the description and the fallback in the code anymore:
https://github.com/vmangos/core/blob/28818aabd89c8d34df696edc7e49808865fffe97/src/realmd/realmd.conf.dist.in#L67-L70
https://github.com/vmangos/core/blob/28818aabd89c8d34df696edc7e49808865fffe97/src/shared/Log.cpp#L84

Of course, if the change was intentional, the description and the code fallback should be changed instead.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


0blu:
Good catch.  
Should we rather change the default?
A simple timestamp might help in a lot of cases.

mserajnik:
> Should we rather change the default?

I wouldn't be against it; if it's done, probably best to adjust the mangosd default too.
But it's up to ratkos. 😄

PR #3196, opened 2026-01-09
Add missing effect index checks in aura scripts

schell244:
## 🍰 Pullrequest
Follow up for: https://github.com/vmangos/core/commit/f669c5038a8fcca7ad4a737ef9caf5eb0f144f26
Adds missing `GetEffIndex() != EFFECT_INDEX_0` checks to aura script callbacks to prevent logic from executing multiple times when spells have multiple effects.
Also fixed a bug in Thaddius charge auras where caster was incorrectly used for distance calculations.

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_nefarian.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_nefarian.cpp
@@ -741,6 +741,9 @@ struct NefarianPolymorphAuraScript : public AuraScript
         if (!apply)
             return;
 
+        if (aura->GetEffIndex() != EFFECT_INDEX_0)
+            return;
+
         Unit* target = aura->GetTarget();
 
         // Randomly select one of three display IDs for the polymorph transform
--- a/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_nefarian.cpp
+++ b/src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_nefarian.cpp
@@ -741,7 +741,7 @@ struct NefarianPolymorphAuraScript : public AuraScript
         if (!apply)
             return;
 
-        if (aura->GetEffIndex() != EFFECT_INDEX_0)
+        if (aura->GetEffIndex() != EFFECT_INDEX_1)
             return;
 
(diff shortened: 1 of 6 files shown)


src/scripts/eastern_kingdoms/burning_steppes/blackwing_lair/boss_nefarian.cpp
```diff
@@ -741,6 +741,9 @@ struct NefarianPolymorphAuraScript : public AuraScript
         if (!apply)
             return;
 
+        if (aura->GetEffIndex() != EFFECT_INDEX_0)
```
ratkosrb: i think you want EFFECT_INDEX_1 here since transform is the second effect?

src/scripts/spells/spell_special.cpp
```diff
@@ -336,6 +339,9 @@ struct SilithystAuraScript : public AuraScript
 
     void OnAfterApply(Aura* aura, bool apply) final
     {
+        if (aura->GetEffIndex() != EFFECT_INDEX_0)
```
ratkosrb: Was on EFFECT_INDEX_1 in the old code

src/scripts/spells/spell_special.cpp
```diff
@@ -502,6 +517,9 @@ struct StoneformAuraScript : public AuraScript
 {
     void OnAfterApply(Aura* aura, bool apply) final
     {
+        if (aura->GetEffIndex() != EFFECT_INDEX_0)
```
ratkosrb, line 528: Was on EFFECT_INDEX_1 in the old code

src/scripts/spells/spell_special.cpp
```diff
@@ -336,6 +339,13 @@ struct SilithystAuraScript : public AuraScript
 
     void OnAfterApply(Aura* aura, bool apply) final
     {
+        // Silithyst PvP was added in patch 1.12, FORCE_REACTION effect on INDEX_1
+#if SUPPORTED_CLIENT_BUILD < CLIENT_BUILD_1_12_1
+        return;
```
ratkosrb: why like this, why not wrap the whole script in it

PR #3198, opened 2026-01-10
Add DNS resolution timeout for realm list updates

schell244:
## 🍰 Pullrequest
Adds configurable DNS resolution timeout to prevent the realm server from hanging when DNS servers are unreachable or slow to respond.

| Scenario | Before | After |
|----------|--------|-------|
| DNS server unreachable | Server hangs indefinitely | Returns after timeout, realm temporarily skipped |
| DNS server slow (>5s) | Waits for OS timeout (30-120s) | Returns after configured timeout |
| Repeated DNS failures | Unlimited blocking threads | Limited to 10 concurrent threads |
| DNS recovers | N/A | Realm automatically returns on next update cycle |

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/shared/IO/Networking/DNS.cpp
+++ b/src/shared/IO/Networking/DNS.cpp
@@ -4,6 +4,11 @@
 #include "Errors.h"
 #include "Util.h"
 #include "IO/SystemErrorToString.h"
+#include <thread>
+#include <mutex>
+#include <condition_variable>
+#include <atomic>
+#include <memory>
 
 #if defined(WIN32)
 #include <WinSock2.h>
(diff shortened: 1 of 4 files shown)


src/shared/IO/Networking/DNS.cpp
```diff
@@ -42,58 +127,92 @@ std::vector<IO::Networking::IpAddress> IO::Networking::DNS::ResolveDomainAll(std
...
-        if (ptr->ai_family == AF_INET)
+        try
         {
-            sockaddr_in* sockaddr_ipv4 = reinterpret_cast<sockaddr_in*>(ptr->ai_addr);
-            IpAddress ip = IO::Networking::Internal::inet_ntop(&(sockaddr_ipv4->sin_addr));
-            list.emplace_back(ip);
+            std::rethrow_exception(state->exceptionPtr);
         }
-        else if (ptr->ai_family == AF_INET6)
+        catch (const std::exception& e)
```
0blu: east const

src/shared/IO/Networking/DNS.cpp
```diff
@@ -42,58 +127,92 @@ std::vector<IO::Networking::IpAddress> IO::Networking::DNS::ResolveDomainAll(std
...
+    }
+    g_pendingDnsResolutions.fetch_add(1);
+
+    // Use shared_ptr for state so it survives thread detachment on timeout
+    auto state = std::make_shared<DnsResolverState>();
 
-    addrinfo* dnsResult = nullptr;
-    if (::getaddrinfo(domainName.c_str(), nullptr, &hints, &dnsResult) != 0)
+    // Worker thread that performs the DNS resolution
+    std::thread resolverThread([state, domainName, type]()
```
0blu: I am not sure if this is good practice. The other solution would be to use our IoContext and then `::GetAddrInfoExW`

0blu (changes requested):

src/shared/IO/Networking/DNS.cpp
0blu: btw we dont use `std::thread` to spawn threads. Prefer to use `IO::Multithreading::CreateThread`

schell244: Thanks for the feedback! Went with the lazy `IO::Multithreading::CreateThread` solution for now - need to look into `::GetAddrInfoExW`, but this would mean platform specific implemenations, correct?

0blu: Yeah, its OS specific. In linux/macos it would be `getaddrinfo_a`.  
Although your current solution works, it feels wrong to create a heavy `std::thread` for just a simple IO task and letting a timeout run in the main thread.

And what happens if a DNS requests gets stuck forever and never resolves?  
`MAX_PENDING_DNS_RESOLUTIONS` would be reached and is denying further requests that might work.

With `GetAddrInfoExW`/`getaddrinfo_a` you don't need to spawn a new thread and correctly cancel timed'out requests.

src/shared/IO/Networking/DNS.cpp
```diff
@@ -29,11 +34,91 @@ std::string IO::Networking::DNS::GetOwnHostname()
     return hostname;
 }
 
-std::vector<IO::Networking::IpAddress> IO::Networking::DNS::ResolveDomainAll(std::string const& domainName, IpAddress::Type type)
+namespace
 {
-    MANGOS_ASSERT(type == IpAddress::Type::IPv4); // TODO: this function is only tested with IPv4. `inet_ntop` will fail
+    // Limit concurrent pending DNS resolutions to prevent thread accumulation on DNS failures
+    std::atomic<int> g_pendingDnsResolutions{0};
+    constexpr int MAX_PENDING_DNS_RESOLUTIONS = 10;
```
0blu, line 41: east const 🙃

0blu:
Have you experienced a blocking DNS requests on your system or what's the reason for this PR?

schell244:
I saw reports about auth server freezes, after DNS being unstable. Closing this for now, as the solution is not ideal - as you said above using `::GetAddrInfoExW` would be best.

PR #3199, opened 2026-01-10
Exclude cpptrace and its deps from install to avoid copying lib/include

schell244:
## 🍰 Pullrequest
Skip install rules for cpptrace and its dependencies to avoid copying unnecessary `lib` and `include` folders.

- cpptrace: Disable install rules (statically linked, not needed at runtime)
- zstd: Use add_subdirectory(..., EXCLUDE_FROM_ALL) instead of
  FetchContent_MakeAvailable so its install rules are not executed
- libdwarf: Same change as zstd

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/dep/cpptrace/CMakeLists.txt
+++ b/dep/cpptrace/CMakeLists.txt
@@ -426,7 +426,11 @@ if(CPPTRACE_GET_SYMBOLS_WITH_LIBDWARF)
           DOWNLOAD_EXTRACT_TIMESTAMP TRUE
           URL "${CPPTRACE_ZSTD_URL}"
       )
-      FetchContent_MakeAvailable(zstd)
+      FetchContent_GetProperties(zstd)
+      if(NOT zstd_POPULATED)
+        FetchContent_Populate(zstd)
+        add_subdirectory(${zstd_SOURCE_DIR}/build/cmake ${zstd_BINARY_DIR} EXCLUDE_FROM_ALL)
+      endif()
     endif()
     # Libdwarf itself
     set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
@@ -438,7 +442,11 @@ if(CPPTRACE_GET_SYMBOLS_WITH_LIBDWARF)
       GIT_TAG ${CPPTRACE_LIBDWARF_TAG}
       GIT_SHALLOW ${CPPTRACE_LIBDWARF_SHALLOW}
     )
-    FetchContent_MakeAvailable(libdwarf)
+    FetchContent_GetProperties(libdwarf)
+    if(NOT libdwarf_POPULATED)
+      FetchContent_Populate(libdwarf)
+      add_subdirectory(${libdwarf_SOURCE_DIR} ${libdwarf_BINARY_DIR} EXCLUDE_FROM_ALL)
+    endif()
     target_include_directories(
       dwarf
       PRIVATE
@@ -601,9 +609,10 @@ endif()
 
 # =============================================== Install ===============================================
 
-if(NOT CMAKE_SKIP_INSTALL_RULES)
-  include(cmake/InstallRules.cmake)
-endif()
+# Disabled: cpptrace is statically linked and not needed at runtime
+# if(NOT CMAKE_SKIP_INSTALL_RULES)
+#   include(cmake/InstallRules.cmake)
+# endif()
 
 # =============================================== Demo/test ===============================================
 
--- a/dep/cpptrace/README.md
+++ b/dep/cpptrace/README.md
@@ -41,3 +41,54 @@ README.md -> README_original.md
  
  # create config file that points to targets file
 ```
+
+### Disabled install rules to prevent cmake_install.cmake generation
+```diff
+--- a/dep/cpptrace/CMakeLists.txt
++++ b/dep/cpptrace/CMakeLists.txt
+@@ -598,6 +598,9 @@ endif()
+
+ # =============================================== Install ===============================================
+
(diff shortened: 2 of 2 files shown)


dep/cpptrace/CMakeLists.txt
```diff
@@ -601,9 +601,10 @@ endif()
 
 # =============================================== Install ===============================================
 
-if(NOT CMAKE_SKIP_INSTALL_RULES)
-  include(cmake/InstallRules.cmake)
-endif()
+# Disabled: cpptrace is statically linked and not needed at runtime
+# if(NOT CMAKE_SKIP_INSTALL_RULES)
```
0blu, line 613: Can we just `CMAKE_SKIP_INSTALL_RULES=TRUE` inside `<root>/dep/CMakeLists.txt:24`, so we don't modify the `cpptrace` tree.
It's hard to keep track of those changes if we upgrade the version of `cpptrace`.

schell244: That's what I've tried with the first commit, but it broke the install process.

0blu: Then let us try this in `dep/CMakeLists.txt`
```cmake
set(PREV_SKIP_INSTALL_RULES ${CMAKE_SKIP_INSTALL_RULES})
set(CMAKE_SKIP_INSTALL_RULES TRUE)
add_subdirectory(cpptrace)
set(CMAKE_SKIP_INSTALL_RULES ${PREV_SKIP_INSTALL_RULES})
```
( And also please add a comment explaining why we are doing that :P )

schell244: Setting `CMAKE_SKIP_INSTALL_RULES = TRUE` prevents CMake from generating  `cpptrace/cmake_install.cmake`. But it seems the parent install script `dep/cmake_install.cmake` (auto-generated) always tries to include install scripts from all subdirectories

ratkosrb: So what are we settling on? Is the current solution the best approach?

0blu: @schell244 can you please document your current change in `dep/cpptrace/README.md` under "## Manual Changes"

mserajnik:
Hmm, this doesn't seem to be working as intended for me; but maybe I'm misunderstanding something.

I am currently already manually [removing `include` and `lib`](https://github.com/mserajnik/vmangos-deploy/blob/de20cdff9e45f8151dcdabe6d60c18ce41a3539f/docker/server/Dockerfile#L155-L158) in my Docker setup. (Yes, instead I should've just used `--exclude` with the [`COPY`](https://github.com/mserajnik/vmangos-deploy/blob/de20cdff9e45f8151dcdabe6d60c18ce41a3539f/docker/server/Dockerfile#L146), but I only realized this afterwards and I was too lazy to change it 😄)

In preparation for this PR, I [added some logging](https://github.com/mserajnik/vmangos-deploy/blob/82f0a66f0a8b8106e27a23f08d39a3efb22c9bab/docker/server/Dockerfile#L156-L161) to confirm my manual cleanup would no longer be necessary when it gets merged. But when I build a Docker image from your [PR branch](https://github.com/schell244/core/tree/cpptrace_skip_install), `include` and `lib` are still present for me until I manually delete them:
```
###BEFORE_CLEANUP###
total 32
drwxr-xr-x 8 vmangos vmangos 4096 Jan 18 15:14 .
drwxr-xr-x 1 root    root    4096 Jan 18 15:04 ..
drwxr-xr-x 3 vmangos vmangos 4096 Jan 18 15:14 bin
drwxr-xr-x 2 vmangos vmangos 4096 Jan 18 15:04 config
drwxr-xr-x 2 vmangos vmangos 4096 Jan 18 15:14 etc
drwxr-xr-x 2 vmangos vmangos 4096 Jan 18 15:14 include <-----------
drwxr-xr-x 4 vmangos vmangos 4096 Jan 18 15:14 lib <-----------
drwxr-xr-x 5 vmangos vmangos 4096 Jan 18 15:04 storage
###AFTER_CLEANUP###
total 32
drwxr-xr-x 1 vmangos vmangos 4096 Jan 18 15:14 .
drwxr-xr-x 1 root    root    4096 Jan 18 15:04 ..
drwxr-xr-x 3 vmangos vmangos 4096 Jan 18 15:14 bin
drwxr-xr-x 2 vmangos vmangos 4096 Jan 18 15:04 config
drwxr-xr-x 2 vmangos vmangos 4096 Jan 18 15:14 etc
drwxr-xr-x 5 vmangos vmangos 4096 Jan 18 15:04 storage
```

I am using a [custom`CMAKE_INSTALL_PREFIX`](https://github.com/mserajnik/vmangos-deploy/blob/de20cdff9e45f8151dcdabe6d60c18ce41a3539f/docker/server/Dockerfile#L74), but that shouldn't matter for this, right?

schell244:
Oh ok, thanks for pointing that out. The `include/` and `lib/` directories are most likely coming from zstd (and potentially libdwarf). These dependencies are dynamically fetched on Linux, which is probably why this issue only occurs there and not on Windows. Sorry for the delayed response, I'm currently on vacation and traveling.

mserajnik:
> The `include/` and `lib/` directories are most likely coming from zstd (and potentially libdwarf).

Ah, sorry, that would make sense. I somehow didn't consider checking what is actually inside these directories, otherwise I would've probably realized myself that what's left in there is not targeted by your original changes here. 😅

Have a nice vacation!

PR #3200, opened 2026-01-11
Pool EPL Plaguebloom

NickTyrer:
## 🍰 Pullrequest
Untitled Diagram


NickTyrer:
Can pooling be automated using sniffs? Because this was painful to complete.

ratkosrb:
I don't see how it would decide what pool to place what spawn in other than whole per zone pools.

PR #3207, opened 2026-01-18
Improve waypoint movement

### LINKED ISSUE #2713: 🚀 [Feature] Improve waypoint movement

Stoabrogga:
Continuation of #3013, which was closed because I deleted my old account.

## 🍰 Pullrequest
Adjust the algorithm for waypoint movement similar to CMaNGOS:

- Try to generate movement splines which are at least 6s long
- 1.5s before the spline is finished start a new spline

This prevents the animation resets occuring at each waypoint.

### Issues
Closes #2713

### How2Test
Watch different creatures which use waypoint movement, e.g.:

- Thora Feathermoon (ID 5798, GUID 15169)
- Theresa (ID 5697, GUID 41840)
- Lord Kazzak (ID 12397, GUID 156950)
- Felguard Elite (ID 8717, GUID 2791)
- Pridewing Soarer (ID 6141, GUID 32374)
- Westfall Woodworker (ID 6670, GUID 45521)
- Thomas Booker (ID 1437, GUID 9468)
- Gordo (ID 10666, GUID 28413)
- Gazelle (ID 4166, GUID 14049)
- Wooly Kodo (ID 3237, GUID 15144)
- Warpwood Tangler (ID 11464, GUID 300943)
- Grunt Gargal (ID 9086, GUID 6883)
- Sorcerer Ashcrombe (ID 3850, GUID 16262)

Also watch creature formations without waypoint movement, e.g. Emeraldon Boughguard (GUID 33246, ID 12474).

Sometimes there are small "jumps" visible while the NPC follows its path, example:
https://github.com/user-attachments/assets/221bbc42-1844-428f-92bb-bbf1c689cc4d

This is not related to this PR, it is caused by the client using low precision linear splines (3 float values packed into one integer value) and the server using high precision. The low precision linear spline coordinates are generated here for the client: https://github.com/vmangos/core/blob/82edaee41485778920586fcb97e3e80e2ad493f0/src/game/Movement/spline/packet_builder.cpp#L108


Stoabrogga:
Concerning the small "jumps" or "teleports" I did a bit more testing:

### Use high precision linear splines on the client

This needs client and server patching:
https://codeberg.org/Stoabrogga/vmangos_misc#use-high-precision-linear-splines

Result: Looks much better, but there is still a small "teleport" forward, like at the beginning of this video:
[video_high_precision_linear_splines](https://github.com/user-attachments/assets/82a0c5f0-63b6-4339-b5c6-8abdf15b823b)

### Use low precision linear splines on the server

Server patch:
[vmdev_compressed_linear_splines.patch](https://github.com/user-attachments/files/24992746/vmdev_compressed_linear_splines.patch)

Please note that this is just a quick-and-dirty implementation, which also does not take partial linear splines into account (see mangosd parameter "Movement.MaxPointsPerPacket"). But it suffices for a few tests.

Result: More or less the same as with high precision. The small "teleport" forward is still there:
[video_low_precision_linear_splines](https://github.com/user-attachments/assets/dbbc37ff-5e8e-42b8-805c-35b8582df2bd)

### Conclusion

It seems that the movement looks better if the server and the client use the same precision for their linear splines. I still don't know what causes the small "teleport" forward at the beginning. Perhaps wrong calculation of the spline duration?
https://github.com/vmangos/core/blob/29fb21c57c6c831bea4d3475afc16c5484c294d7/src/game/Movement/spline/packet_builder.cpp#L147-L148

PR #3220, opened 2026-02-06
Fix memory leaks in GameEventMgr, MotionMaster, ObjectMgr and Pet

schell244:
## 🍰 Pullrequest

Fix some memory leaks and uninitialized member variables discovered during Valgrind profiling.
- GameEventMgr: Implement destructor to clean up mGameEventHardcodedList entries
- MotionMaster: Fix memory leak in ClearType() by deleting non-static movement generators
- ObjectMgr: Add cleanup methods for item prototypes and creature auras
- Pet: Add ClearCharmInfo() call in destructor
- MasterPlayer: Initialize unReadMails and m_nextMailDelivereTime in constructor
- battleground_alterac: Initialize Event_Timer and Point in Reset()
- MotionMaster: Remove stray semicolon
- PlayerBotAI: Add cleanup on error path and PvP flag initialization

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/src/game/Chat/MasterPlayer.cpp
+++ b/src/game/Chat/MasterPlayer.cpp
@@ -11,7 +11,8 @@
 
 MasterPlayer::MasterPlayer(WorldSession* s):
     m_speakTime(0), m_speakCount(0), m_social(nullptr), 
-    m_session(s), m_mailsUpdated(false)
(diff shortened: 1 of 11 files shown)


ratkosrb:
I've applied the fixes that are actually legit.

PR #3222, opened 2026-02-06
Check itr in DeleteCharacterPetById

Wall-core:
It should never be invalid, but ensure it is so we don't crash.


Gamemechanicwow:
i have a slight issue with this type of fix. A crash would indicate that there is a bug somewhere else in the code - and this just masks the real issue.

Wall-core:
Yeah it should never be invalid at this part in the code, but if the two maps somehow desynced, it would crash if this were ever called. It's just more proper to actually check both

PR #3225, opened 2026-02-15
Remove packet parsing logic out of PacketHandler

0blu:
This MR will remove the old packet parsing logic out of PacketHandlers (manual using `WorldPacket`) and introduces new classes for each packet type.

Notable example
```diff
-        GetPlayer()->TeleportTo(mapId, positionX, positionY, positionZ, orientation);
+        GetPlayer()->TeleportTo(packet.location);
```

- [X] Test build of all supported client builds


src/game/Server/Packets/Trade.h
```diff
@@ -14,6 +14,44 @@ namespace WorldPackets { namespace Trade
...
+        uint8 tradeSlot = 0;
+
+        explicit ClearTradeItem() : ClientPacket(CMSG_CLEAR_TRADE_ITEM) {}
+        void ReadFromWorldPacket(WorldPacket& recv_data) override;
+    };
+
+    class AcceptTrade final : public ClientPacket
+    {
+    public:
+        uint32 unknown = 0; // not used by server
```
ratkosrb: can't we figure out what this is

src/game/Handlers/ChatHandler.cpp
```diff
@@ -733,12 +732,9 @@ void WorldSession::HandleTextEmoteOpcode(WorldPacket& recv_data)
...
 
-    uint32 textEmote, emoteNum;
-    ObjectGuid guid;
-
-    recv_data >> textEmote;
-    recv_data >> emoteNum;
-    recv_data >> guid;
+    uint32 textEmote = packet.textEmote;
+    uint32 emoteNum = packet.emoteNum;
+    ObjectGuid guid = packet.guid;
```
ratkosrb: i don't like all of this copying, can't we just use the value in the packet directly

0blu: I don't want to touch the packet handlers too much.

ratkosrb: The packet handlers were written when there were no packet classes, you had to create local variables and manually read from the packet into them. If we are switching to packet class these local variables shouldn't exist.

src/game/Server/Packets/Battleground.cpp
```diff
@@ -19,3 +19,19 @@ void WorldPackets::Battleground::BattlemasterHello::ReadFromWorldPacket(WorldPac
 {
     recv_data >> guid;
 }
+
+void WorldPackets::Battleground::BattleFieldPort::ReadFromWorldPacket(WorldPacket& recv_data)
+{
+#if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_8_4
+    recv_data >> mapId >> action;
+#else
+    recv_data >> action;
```
ratkosrb: reading of action can be outside of the build check

src/shared/ByteBuffer.h
```diff
@@ -82,6 +76,24 @@ class ByteBuffer
...
+        }
+
+        static ByteBuffer from(std::vector<uint8> const&& v)
+        {
+            ByteBuffer buf;
+            buf._storage = std::move(v);
+            buf._rpos = 0;
+            buf._wpos = buf._storage.size();
+            return buf;
+        }
```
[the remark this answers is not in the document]
0blu: @copilot apply changes based on [this feedback](https://github.com/vmangos/core/pull/3225#discussion_r2937234334) really copy by using non-const. add a single  line comment above this function.

ratkosrb:
Whats up with this new class SpellCastTargetsInfo, why do we need this?

src/game/Spells/SpellCastTargetsInfo.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+void SpellCastTargetsInfo::ReadFromBuffer(ByteBuffer& data)
+{
+    data >> m_targetMask;
+
+    if (m_targetMask == TARGET_FLAG_SELF)
+        return; // Will be statefully filled when using FromSpellCastTargetsInfo
+
+    // TARGET_FLAG_UNIT_MINIPET is used for non-combat pets, maybe other?
+    if (m_targetMask & (TARGET_FLAG_UNIT | TARGET_FLAG_UNIT_MINIPET))
```
ratkosrb: We already found out this is not even possible, the target mask is uint16 and it can't represent 0x00010000.

0blu:
> Whats up with this new class SpellCastTargetsInfo, why do we need this?

The current SpellCastTargets is stateful, you need a reference the player.

Which we might not have in the future when parsing will be handled in a different thread.

src/game/Spells/SpellCastTargetsInfo.cpp
0blu: I forgot to change that after rebase

src/game/Chat/Chat.h
```diff
@@ -91,6 +91,12 @@ enum PlayerChatTag
     CHAT_TAG_GM                 = 3,
 };
 
+enum class ParseCommandResult
```
ratkosrb, line 94: why do we need an enum when there are only 2 possible return values, using bool is simpler

0blu: A boolean in this case is ambiguous.

I agree that booleans can be used for simple checks like `isEmpty()` or something like this.

src/game/PlayerBots/CombatBotBaseAI.cpp
```diff
@@ -3078,51 +3078,49 @@ void CombatBotBaseAI::SendBattlefieldPortPacket()
 {
     for (uint32 i = BATTLEGROUND_QUEUE_AV; i <= BATTLEGROUND_QUEUE_AB; i++)
     {
-        if (me->IsInvitedForBattleGroundQueueType(BattleGroundQueueTypeId(i)))
+        if (me->IsInvitedForBattleGroundQueueType(static_cast<BattleGroundQueueTypeId>(i)))
         {
             WorldPacket data(CMSG_BATTLEFIELD_PORT);
```
ratkosrb: shouldn't this line be removed

0blu: Good catch

ratkosrb:
image

This pull request does not compile. Goes on like this. The errors are in Opcodes.cpp.

ratkosrb:
I have fixed the build. Such a big function like BuildOpcodeList cannot be constexpr.

PR #3245, opened 2026-03-11
Added a new console account command to set the email

p4kkabx:
## 🍰 Pullrequest
Implemented the `.account email` console command. This feature allows administrators to manage account emails with built-in security hierarchy checks and automatic cache synchronization.

### Proof
Tests performed via console:
- **Success:** `mangos>account set email admin mail@mail.com` -> `Account ADMIN (Id: 1) email set to: mail@mail.com`
- **Missing Args:** `mangos>account set email` -> `Sintaxis incorrecta.`
- **Partial Args:** `mangos>account set email admin` -> `Sintaxis incorrecta.`
- **Invalid Account:** `mangos>account set email invented mail@mail.com` -> `Cuenta no existe: INVENTED`

### How2Test
1. **Valid Update:** Run `.account set email <name> <email>`. Verify the DB `account` table and the console output.
2. **Argument Validation:** Run the command with 0 or 1 arguments to ensure the "Incorrect Syntax" message triggers.
3. **Database Integrity:** Provide a non-existent account name to ensure the "Account does not exist" error is handled.
4. **Security:** Log in as a lower-security GM and attempt to change a higher-security account's email to verify the hierarchy check.

### Todo / Checklist
- [x] SQL migration `20260310190604_world.sql` added with string `1035`.
- [x] Command registered in `Chat.cpp` under `SEC_CONSOLE`.
- [x] Logic implemented in `AccountCommands.cpp`.
- [x] Hierarchy safety check via `HasLowerSecurityAccount`.
- [x] Real-time cache update via `sAccountMgr.UpdateAccountData`.
- [x] Basic email format validation implemented.

### DIFF
--- a/src/game/Handlers/MiscHandler.cpp
+++ b/src/game/Handlers/MiscHandler.cpp
@@ -601,25 +601,19 @@ void WorldSession::HandleReclaimCorpseOpcode(WorldPackets::Misc::ReclaimCorpse c
 
 void WorldSession::HandleResurrectResponseOpcode(WorldPackets::Misc::ResurrectResponse const& packet)
 {
-    if (!packet.guid) // Cheating attempt
-    {
-        ProcessAnticheatAction("PassiveAnticheat", "Instant resurrect hack detected", CHEAT_ACTION_LOG | CHEAT_ACTION_REPORT_GMS);
-        return;
-    }
-
     if (GetPlayer()->IsAlive())
         return;
 
-    if (packet.status == 0)
+    if (!packet.accept)
     {
-        GetPlayer()->ClearResurrectRequestData();           // reject
+        GetPlayer()->ClearResurrectRequestData(); // player denied rezz attempt
         return;
     }
 
-    if (!GetPlayer()->IsRessurectRequestedBy(packet.guid))
+    if (!GetPlayer()->IsRessurectRequestedBy(packet.resurrectorGuid))
         return;
 
-    GetPlayer()->ResurectUsingRequestData();                // will call spawncorpsebones
+    GetPlayer()->ResurrectUsingRequestData();     // will call SpawnCorpseBones
 }
 
 void WorldSession::HandleAreaTriggerOpcode(WorldPackets::Misc::AreaTrigger const& packet)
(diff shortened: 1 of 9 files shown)


src/game/Commands/AccountCommands.cpp
```diff
@@ -207,6 +207,48 @@ bool ChatHandler::HandleAccountSetLockedCommand(char* args)
...
+    // Basic email validation  
+    if (email.find('@') == std::string::npos)  
+    {  
+        SendSysMessage("Invalid email format.");  
+        SetSentErrorMessage(true);  
+        return false;  
+    }  
+  
+    // Update the email using AccountMgr  
+    LoginDatabase.PExecute("UPDATE `account` SET `email` = '%s' WHERE `id` = '%u'", email.c_str(), account_id);
```
ratkosrb: you should escape the string first

p4kkabx: Corrected. I've added the escape_string call right after the validation to secure the input before the SQL update. Thanks for pointing that out!

ratkosrb: You need to create a copy of the string and escape that and use it only in the query, because like this you will save the escaped string into the server memory when calling UpdateAccountData instead of the original email.

p4kkabx: Done, thanks for the detailed explanation.

src/game/Chat/Chat.cpp
```diff
@@ -1485,6 +1486,14 @@ AccountTypes ChatHandler::GetAccessLevel() const
 
 bool ChatHandler::isAvailable(ChatCommand const& cmd) const
 {
+    // Allow party bot commands when SkipChecks is enabled  
+    if (sWorld.getConfig(CONFIG_BOOL_PARTY_BOT_SKIP_CHECKS) &&   
+        (strcmp(cmd.Name, "partybot") == 0 ||   
+         (!cmd.FullName.empty() && cmd.FullName.find("partybot") != std::string::npos)))  
+    {  
+        return true;  
+    }
```
mserajnik: This seems unrelated to this PR?

mserajnik (changes requested):

src/game/Chat/Chat.cpp
p4kkabx: Yes, you are absolutely right. I forgot I had this PR still open, excuse me.

PR #3250, opened 2026-03-13
Treat Undercity slime as water

### LINKED ISSUE #3246: 🐛 [Bug] Green slime river in Undercity have damage.
## 🐛 Bug report
Image

### Expected behavior
It's just a river, it shouldn't cause damage.

### Steps to reproduce

1. Go to Green slime river in Undercity.
2. Wait
3. Profit

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  https://github.com/vmangos/core/commit/f54db6b4a82f40929c13868f61f0fa1e40c974c7

OS Client: Windows
OS Server: Ubuntu

### Crashlog

- None

Stoabrogga:
## 🍰 Pullrequest
Treat the slime in the Undercity as water, so it won't hurt the players anymore. Don't know if this is the correct way to fix it, but it works.

Related commit:
f54db6b4a82f40929c13868f61f0fa1e40c974c7 (PR #3118)

### Proof
None

### Issues
Closes #3246

### How2Test
- Extract maps/vmaps
- Swim in the slime in Undercity

### Todo / Checklist
None

### DIFF
--- a/contrib/vmap_extractor/vmapextract/wmo.cpp
+++ b/contrib/vmap_extractor/vmapextract/wmo.cpp
@@ -474,12 +474,18 @@ int WMOGroup::ConvertToVMAPGroupWmo(FILE* output, WMORoot* rootWMO, bool pPrecis
                     liquidEntry = 3;        // magma
                     break;
                 case 3:
-                    if (rootWMO->RootWMOID == 4489) // Stratholme_raid.wmo WMOID == 4489
+                    switch (rootWMO->RootWMOID)
                     {
-                        liquidEntry = 21;   // Naxxramas slime
+                        case 4489:            // Stratholme_Raid.wmo, WMOID 4489
+                            liquidEntry = 21; // Naxxramas slime
+                            break;
+                        case 1150:            // Undercity.wmo, WMOID 1150
+                            liquidEntry = 1;  // Water
+                            break;
+                        default:
+                            liquidEntry = 4;  // Normal slime
+                            break;
                     }
-                    else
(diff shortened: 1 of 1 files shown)


Wall-core:
the issue is with the other commit that was added. It fixed the checks for naxxramas slime, but for some reason removed the checks for IsInSlime which excluded undercity slime. the IsInSlime() function should be fixed instead. Slime should not apply environmental damage; instead, it applies a debuff which causes the damage.

Stoabrogga:
@Wall-core There was no special handling of Undercity, the code parts were commented out:
https://github.com/vmangos/core/blob/efa0d8b043d4798db4e409a2486c71e8885cfd2e/src/game/Objects/Player.cpp#L1038-L1040

But in commit f54db6b4a82f40929c13868f61f0fa1e40c974c7 the comment was deleted and this was added instead:
https://github.com/vmangos/core/blob/f54db6b4a82f40929c13868f61f0fa1e40c974c7/src/game/Objects/Player.cpp#L1038-L1039

Wall-core:
what I'm saying though, is that IsInSlime() shouldn't even exist. There's no slime damage in vanilla that isn't applied through a debuff, as shown in the liquid dbcs. There only exists slime in Undercity and Naxxramas, and Naxxramas slime does not apply environmental damage either. This function should be removed entirely.

Stoabrogga:
Ok, thanks for the explanation. I'll close this PR and open a new one removing the "IsInSlime" function.

PR #3255, opened 2026-03-14
Fix cmake warning

Stoabrogga:
## 🍰 Pullrequest
Fix this warning:
```
CMake Warning (dev) at /usr/share/cmake-3.31/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(zstd) is deprecated, call
  FetchContent_MakeAvailable(zstd) instead.  Policy CMP0169 can be set to OLD
  to allow FetchContent_Populate(zstd) to be called directly for now, but the
  ability to call it with declared details will be removed completely in a
  future version.
Call Stack (most recent call first):
  dep/cpptrace/CMakeLists.txt:431 (FetchContent_Populate)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

Related commit:
efa0d8b043d4798db4e409a2486c71e8885cfd2e (PR #3199)

### Proof
None

### Issues
None

### How2Test
I used Cmake 3.31.6 on Debian 13:
```bash
mkdir ~/vmangos/build
cd ~/vmangos/build
cmake ~/vmangos/core -DCMAKE_C_FLAGS="-w" -DCMAKE_CXX_FLAGS="-w" -DUSE_PCH=OFF -DUSE_STD_MALLOC=1 -DUSE_EXTRACTORS=0 -DCMAKE_INSTALL_PREFIX=~/vmangos-srv -DCMAKE_BUILD_TYPE=Release
```

### Todo / Checklist
None

### DIFF
--- a/dep/cpptrace/CMakeLists.txt
+++ b/dep/cpptrace/CMakeLists.txt
@@ -425,11 +425,11 @@ if(CPPTRACE_GET_SYMBOLS_WITH_LIBDWARF)
           SOURCE_SUBDIR build/cmake
           DOWNLOAD_EXTRACT_TIMESTAMP TRUE
           URL "${CPPTRACE_ZSTD_URL}"
+          EXCLUDE_FROM_ALL
       )
       FetchContent_GetProperties(zstd)
       if(NOT zstd_POPULATED)
-        FetchContent_Populate(zstd)
-        add_subdirectory(${zstd_SOURCE_DIR}/build/cmake ${zstd_BINARY_DIR} EXCLUDE_FROM_ALL)
+        FetchContent_MakeAvailable(zstd)
       endif()
     endif()
     # Libdwarf itself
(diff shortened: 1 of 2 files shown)


GrenderG:
Maybe a better approach could be found to still keep support for older cmake versions, @Stoabrogga. That way we can honor:

```
cmake_minimum_required(VERSION 3.12)
```

0blu:
We could upgrade to `3.13.5` it's the last Windows XP supported version

https://cmake.org/files/v3.13/cmake-3.13.5-win32-x86.zip

Edit: I made a typo and wrote 3.15 instead of 3.13

GrenderG:
Is `FetchContent_MakeAvailable(zstd)` supported in that version?

Stoabrogga:
> Is `FetchContent_MakeAvailable(zstd)` supported in that version?

Hmm, the last cmake version which supports Windows XP is `3.13.5`, not `3.15.5`. They dropped support for Windows XP in `3.14`:
https://cmake.org/cmake/help/latest/release/3.14.html#deprecated-and-removed-features

`FetchContent_MakeAvailable` needs at least 3.14:
https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable

ratkosrb:
Does this need to be reverted then?

Stoabrogga:
Perhaps both ways can be implemented, one for cmake < 3.14, and one for cmake >= 3.14. I'll take a look on this.

PR #3256, opened 2026-03-14
Add build option "ENABLE_CPPTRACE"

Stoabrogga:
## 🍰 Pullrequest
Add build option "ENABLE_CPPTRACE". Default is "ON". If off, cpptrace is disabled. This can be useful if it is not needed, e.g. if gdb is used instead. Also speeds up the build process because cpptrace has a dependency to libdwarf which also needs zstd. Both are downloaded during the build process.

### Proof
None

### Issues
None

### How2Test
I used Cmake 3.31.6 on Debian 13:
- Preparation:
  ```bash
  mkdir ~/vmangos/build
  cd ~/vmangos/build
  ```
- Build with default ("ENABLE_CPPTRACE" is "ON"):
  ```
  cmake ~/vmangos/core -DCMAKE_C_FLAGS="-w" -DCMAKE_CXX_FLAGS="-w" -DUSE_PCH=OFF -DCMAKE_INSTALL_PREFIX=~/vmangos-srv -DCMAKE_BUILD_TYPE=Release
  make -j $(($(nproc)+2))
  ```
- Turn off "ENABLE_CPPTRACE":
  ```
  cmake ~/vmangos/core -DCMAKE_C_FLAGS="-w" -DCMAKE_CXX_FLAGS="-w" -DENABLE_CPPTRACE=OFF -DUSE_PCH=OFF -DCMAKE_INSTALL_PREFIX=~/vmangos-srv -DCMAKE_BUILD_TYPE=Release
  make -j $(($(nproc)+2))
  ```

### Todo / Checklist
None


cmake/options.cmake
```diff
@@ -5,6 +5,7 @@ option(USE_SCRIPTS                    "Build scripts"
 option(USE_EXTRACTORS                 "Build extractors"                                                                              OFF)
 option(USE_REALMMERGE                 "Build helper tool for merging character databases"                                             OFF)
 option(ENABLE_MAILSENDER              "Enables support for sending emails via sendgrid.com (requires libcurl)"                        OFF)
+option(DISABLE_STACK_TRACE            "Disables cpptrace stack tracing if not needed (e.g. if GDB is used)"                           OFF)
```
0blu: Maybe we should call it `ENABLE_STACK_TRACE` and make it `ON` by default.
Just to prevent double negation and metal overhead.

0blu: Maybe `ENABLE_RUNTIME_STACKTRACE` (although a stack trace is usually runtime only)

0blu: Or `ENABLE_CPPTRACE_STACKTRACE`

Stoabrogga: Or just `ENABLE_CPPTRACE`?

Stoabrogga: I now use `ENABLE_CPPTRACE` because it is shorter, hope that's ok.

PR #3258, opened 2026-03-15
Fix Tranquil air totem 

Gamemechanicwow:
## 🍰 Pullrequest

When playing on a 1.12 client on a server with patch set below 1.9 - tranquil totem will not spawn.

Tranquil air totem was added in patch 1.9, so its current patch set to 7 is technically correct, but it breaks the spell on 1.12.

I know.. This smells like a hackfix. 

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


ratkosrb:
We will not erase progression data to accommodate mismatched client and content versions. Set the content patch to 10 if you want all the 1.12 content to work. Everything in the database should have it's correct patch assigned.

PR #3263, opened 2026-03-16
Feature/ahbot buyout

ivgrekow:
## 🍰 Pullrequest

The PR is intended to improve current implementation of auction house bot.

### Features
- Alters `auctionhousebot` table and fills with reasonable values almost for every item
- Adds buyout feature for ahbot that makes decision based on its internal listing `auctionhousebot`
- Price of items with random enchantments is scaled due to its suitability
- Number of items placed by bot is customizable as well but the quantity of each item category is approximately the same

### Issues

- None

### How2Test

Not sure how to provide the right testlist, so much lines were added. Generally, to ensure it works:
- Type `ahbot reload` and `ahbot update` to load ahbot config and fill auction house; 
- Place item on AH with cheapest buyout price; there is 10% chance for every `Update` call to buyout your item

### Todo / Checklist

- The price data is scrapped from website that stores auction info for various servers. That means some values in `auctionhousetable` are weird. Possibly, it needs more migrations to fix the values.


mserajnik:
I like this in principle (especially the added buyout functionality), but I am not sure if we really should have this many items by default, since that makes it quite a large table for an optional, non-blizzlike feature that probably only a small percentage of the userbase will enable.

Maybe it would be better to keep a much smaller default set of the most useful items (similar to how it currently is) and provide these "complete" sets of the various item categories (if I saw it correctly how you split the migrations) as optional SQL files people can import if they want to; e.g., in the [`sql/custom`](https://github.com/vmangos/core/tree/49eebd079dd5bf666956271a461b5583c1aba2c1/sql/custom) directory.

Accordingly, I would also keep the `AHBot.itemcount` low by default.

Instructions could then be added [to the wiki](https://github.com/vmangos/wiki/wiki/auctionhousebot) on how to add the larger sets or how to insert specific items (side note: a lot of information in the wiki is outdated, especially how to compile VMaNGOS; might be good to update it in regards to the ACE removal etc. at some point).

Just my personal opinion, maybe ratkos sees this differently. 😄

gajet5:
It seems to me like a bad idea to set a buyout price. It seems to me that it is better to use the existing figures in the "item_template" table for the "sell_price" value with a random modifier of the acceptable buyout price.

I think this modification should work with all things, not just the ones in the table.

ivgrekow:
Thanks to all participants who left a comment here. I'll try to explain in detail all things you mentioned.

> Maybe it would be better to keep a much smaller default set of the most useful items (similar to how it currently is) and provide these "complete" sets of the various item categories (if I saw it correctly how you split the migrations) as optional SQL files people can import if they want to; e.g., in the [sql/custom](https://github.com/vmangos/core/tree/49eebd079dd5bf666956271a461b5583c1aba2c1/sql/custom) directory.

I didn't think of that. The idea of buyout feature came to my mind when I was solo leveling a character on my server running locally. I thought it would be good if auction house bot could compare every item that possibly can be placed on AH with some internal listing. Inspecting ahbot codebase I just tried to extend current implementation i.e. "larger means better". To be honest, I don't understand which items are more often used than others. How can we distinguish which is more important?

> It seems to me like a bad idea to set a buyout price. It seems to me that it is better to use the existing figures in the "item_template" table for the "sell_price" value with a random modifier of the acceptable buyout price.

There are plenty of items that have sell prices equal to 0. For example [Nexus Crystal](https://classicdb.ch/?item=20725). How should we scale that price? Moreover, the database sell price doesn't reflect the actual value of tradable items.

mserajnik:
> I don't understand which items are more often used than others. How can we distinguish which is more important?

I would say profession-related items are among the most useful if you don't have other players with the respective professions to provide things you might need (e.g., leather for Engineering), because there is often no other good (or even no legitimate) way to acquire them. The current set is also mostly such items.

Personally, I wouldn't want most items to be purchasable from the AH on my private server. I would rather gather/earn items I can get by myself, because that feels more fulfilling (even if it can take a while). But I can also understand the other opposite way of thinking here. 😄

ivgrekow:
> I would say profession-related items are among the most useful if you don't have other players with the respective professions to provide things you might need (e.g., leather for Engineering), because there is often no other good (or even no legitimate) way to acquire them. The current set is also mostly such items.

Okay, now I get it. If I'm correct, you mean that the table should have only items which can be produced from crafting and which are required by other professions. If so I think you're right about moving all migrations into 'custom' folder. However, countrary to what you said, Light Leather, for example, is required for tailoring and it can be obtained by fishing! You just have to find floating crates and get lucky.

In more detail, I wrote a simple scraper script that requests html-pages and parses their tags containing gold/silver/copper values. The purpose is to store _real_ average prices for all tradeable items like it is on real server. Yeah, [Copper Bar]x20 can cost 40s on one server and 20s on another. But generally its market value lies between 20s and 40s. I thought it would be great if the AHBot could cover the needs for items whether they are Thorium Bars or Shredder Operating Manual pages.

mserajnik:
> If I'm correct, you mean that the table should have only items which can be produced from crafting and which are required by other professions. If so I think you're right about moving all migrations into 'custom' folder.

I have to stress: this is just my opinion/preference. If ratkos or any of the other maintainers think otherwise, their opinion obviously outweighs mine here. 😉

But yes, you understood what I meant: I think it is better to keep the default set of items rather small, since it would mainly serve as an example. And if we define that the default set should contain, let's say, 100 items, then I believe it makes sense to have items that come from crafting and are otherwise impossible or hard to obtain part of that set.

If someone then decides they want the AH bot to essentially sell "all the items", we could have clear instrucions how to import that custom SQL.

> However, countrary to what you said, Light Leather, for example, is required for tailoring and it can be obtained by fishing! You just have to find floating crates and get lucky.

You are right, leather was a poor example, since it can be obtained from various sources outside of Skinning (even if tedious). A better one would be Silver Rods (needed for Enchanting and only obtainable from Blacksmithing), which are currently also part of the set of items the AH bot sells.

ivgrekow:
Since my last post, I’ve done several personal reviews of how auction bots are structured in other projects. I think the current version of the “improved” auction bot is still somewhat homebrew. For my own private needs, this functionality works great, but for a project that requires a certain level of “general reliability,” it’s not suitable. In short, I’m closing this pull request for now and will think about how to refine it.

PR #3267, opened 2026-03-19
Added Batch for Windows Extraction (for easier of use)

TheSCREWEDSoftware:
## 🍰 Pullrequest

Prior to this PR you would have to run each of them in order and 1 by 1 (MapExtractor, Vmap Extracotr, VampAssemeblr and than MMAP generator)

With this (gets copied with windows extractor tool if they ticked in cmake) and matter of moving the batch file alongside the exectuables

### Proof

Extractor CLI
image

Without Extractors
image

With Extractors
Screenshot_1

https://github.com/Ryan-s-Projects/1.12.1-5875-Vanilla/releases/tag/Data
(this was the data extracted using the extractorBat + the extractors at the current pre-compiled binaries from release, based on a private server client (unsure if it's or not modified) but it's the proof it worked (the extraction).

### Issues

- None

### How2Test

- Be on windows
- Tick the USE_EXTRACTORS
- Compile
- See the batch file with the extractors
- Run the batch file with the extractors in the same place
- See it working

### Todo / Checklist

- [X] I don't use Linux but having a shell file equivalent for linux and mac (unsure if this supported by vmangos) woudln't be a bad addition.


contrib/extractor/extractorWindows.bat
```diff
@@ -0,0 +1,99 @@
...
+)
+if exist VMapAssembler.exe (
+    echo Assembling vmaps...
+    VMapAssembler.exe
+) else (
+    echo VMapAssembler.exe not found!
+)
+if exist MoveMapGenerator.exe (
+    echo Extracting mmaps...
+    MoveMapGenerator.exe
```
mserajnik, line 90: You might want to pass `--configInputPath` and `--offMeshInput` here. E.g.:
```sh
MoveMapGenerator.exe --configInputPath "config.json" --offMeshInput "offmesh.txt"
```
(the files should be in `${BIN_DIR}/Extractors` by default iirc, so the same directory as the extractor executables)

mserajnik:
> I don't use Linux but having a shell file equivalent for linux and mac (unsure if this supported by vmangos) woudln't be a bad addition.

I have this [in my Docker setup](https://github.com/mserajnik/vmangos-deploy/blob/4a9485f47daf1c72987e107401e9a8e8da135620/docker/server/docker-cmd-extract-client-data.sh), might serve as inspiration for a script.

Daribon:
I made a Linux extractor script myself a few months back (pre native merge):
https://gist.github.com/Daribon/03a696faf50c70ef692ca9803843d7ac#file-install-vmangos-sh-L328

It is very integrated with the entire install script I made, but it could be modified to be standalone.

Daribon:
> > I don't use Linux but having a shell file equivalent for linux and mac (unsure if this supported by vmangos) woudln't be a bad addition.
> 
> I have this [in my Docker setup](https://github.com/mserajnik/vmangos-deploy/blob/4a9485f47daf1c72987e107401e9a8e8da135620/docker/server/docker-cmd-extract-client-data.sh), might serve as inspiration for a script.

Already made new version, which grabs build number from the wow.exe, so user doesn't have to move the dbc files manually: https://gist.github.com/Daribon/4072493ffd0a4f2c6fc54a1895fdc743

Already tested and works as intended. It will extract maps, dbc, vmaps and mmaps.

With the power of ai slop, I also have a windows version of the above script:
https://gist.github.com/Daribon/f2148639837476d7b4f02153daf2d5b7
Untested, but probably works.

In none of the above scripts is mmaps optional, it is forced. However I could easily change it to ask the user.

contrib/extractor/extractorWindows.bat
TheSCREWEDSoftware: I can add an input to ask the files within the same directoy otherwise by default mmaps generator already uses them by default

image

mserajnik: Ah, right, I forgot because I install thise files to custom locations. 😄

Don't think it's necessary to prompt the user then.

PR #3276, opened 2026-03-20
Add Conjure Water Level 7 to Mage Trainer on GM Island

IamTirion:
## 🍰 Pullrequest
As title says

https://www.wowhead.com/classic/spell=10143/conjure-water

### How2Test
Create a Mage
Speak to the Mage Trainer on GM Island
Learn Conjure Crystal Water

### DIFF
--- a/sql/migrations/20260320183126_world.sql
+++ b/sql/migrations/20260320183126_world.sql
@@ -1,0 +1,18 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20260320183126');
+IF v = 0 THEN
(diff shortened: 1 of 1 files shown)


ratkosrb:
This is custom, it shouldn't be added to the main db.

IamTirion:
> This is custom, it shouldn't be added to the main db.

Why is it custom? Mage learn this from a quest in Dire Maul normally. The trainers on the GM Island can train spells such as Berserker Stance, which is normally learned from a quest. So I think this is the same.

This is how it is learned normally
https://www.wowhead.com/classic/news/how-to-learn-conjured-crystal-water-wow-classic-fresh-363219

Wall-core:
This is a GM Island vendor, which is custom. It should be placed in the add_gm_island_vendor.sql scripts instead of a direct commit to the main db.

IamTirion:
> This is a GM Island vendor, which is custom. It should be placed in the add_gm_island_vendor.sql scripts instead of a direct commit to the main db.

Okay. That means I shouldn't add migration files, but should add these SQL Insert lines in the add_gm_island_vendor.sql?

PR #3291, opened 2026-03-26
Fix rep loss for Blackwater Deckhand

Wall-core:
should be 5 gained, -25 lost for this NPC

### DIFF
--- a/sql/migrations/20260406103411_world.sql
+++ b/sql/migrations/20260406103411_world.sql
@@ -1,0 +1,40 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20260406103411');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20260406103411');
+-- Add your query below.
+
+-- Gordo
+-- Gloom Weed spawns are pooled and thus sometimes can't be found and despawned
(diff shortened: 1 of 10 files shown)


Wall-core:
we need sniffs/screenshots for all these NPCs, their reputation values seem incorrect (5 gain, -125 lost).

Wall-core:
Moving this to bugtracker instead of PR until sniffs or proofs are provided

ellingtonsantos:
screenshot
5 gain / 25 lost

Wall-core:
awesome, thanks for confirming.
Can you check all the other npcs and log their rep gain/loss? also, i assume you're level 60?

ellingtonsantos:
Human Warrion, level 60. Do you prefer screenshots or sniffs?

Wall-core:
screenshots are fine, as long as we get all of them

ratkosrb:
Be aware the player's level affects reputation gain. If the creature is gray the reputation amount will be reduced.

ratkosrb:
I have one TBC sniff in which this mob is killed and the player loses 25 reputation but he is level 70. On vmangos if you set your level to 70 and kill this mob you lose the same amount.

ratkosrb:
Here is a parse of reputation gain data from classic sniffs. Only includes cases where the mob is not grey to the player. Unfortunately the reputation gain packet does not include the source of the rep, so false positives are possible. The parser tracks what is the last creature that died, and when it reads a reputation change packet it assigns the reputation to that creature. If you kill a mob at the same time you turn in a quest for example, it will think the creature is the source of the quest reputation reward. And in cases where multiple mobs die at once, it can also assign reputation wrongly, as it has no way to know which one is the source of the rep change. In Alterac Valley its particularly difficult to assign reputation, as you are constantly gaining rep from things other players in your team are doing. Treat this data sceptically.

https://pastebin.com/JdiyVnXJ

PR #3314, opened 2026-03-30
Fix some packets

0blu:


src/game/Handlers/ChatHandler.cpp
0blu, line 167: We are in `!= LANG_ADDON` already.

That's why:
```diff
- if (packet.lang != LANG_ADDON && GetMasterPlayer())
+ if (GetMasterPlayer())
```

PR #3316, opened 2026-03-31
CI: Clean up workflows.

mserajnik:
## 🍰 Pullrequest
This PR cleans up and improves the GitHub Actions workflows. I ended up effectively rewriting most things.

Changes include:

- Removed the obsolete TBB handling (since it's no longer supported)
- Updated all used actions to their latest major versions (so we are ready for the [Node.js 20 removal](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/))
- Replaced the unmaintained (last commit in [2023](https://github.com/crowbarmaster/GH-Automatic-Releases/commits/main/)) `crowbarmaster/GH-Automatic-Releases` with a custom shell script
- Removed `Kaven-Universe/github-action-current-date-time@` as well (no longer needed with the new shell script); this means all actions left are official `actions/*` ones maintained by GitHub
- Extracted some of the repeated DB shell logic into reusable scripts to reduce duplicated code
- Pinned `ubuntu-latest` to `ubuntu-24.04` instead (which is currently `ubuntu-latest`), so we won't silently get upgraded to `ubuntu-26.04` next month
- Added concurrency guards to the release workflows to prevent overlapping pushes from publishing `latest` / `db_latest` out of order
- Fixed the compiler matrix to make the Ubuntu Clang build actually work
- Changed the workflows to use all available build cores/threads instead of just 2
- Cleaned up the Ubuntu build dependencies
- Cleaned up the CMake flags
- Used explicit PowerShell cmdlets such as `Set-Location` and `Copy-Item` instead of shell aliases to make the Windows workflow steps less alias-dependent
- Added descriptive workflow, job, and step names
- Used more descriptive names for the workflow files
- Did some other minor cleanup (formatting, comments, removal of dead code etc.)

I kept the current `latest` and `db_latest` release names, even though I don't love them (to not break things for anyone who runs scripts/automations that rely on those names). Likewise, the actual artifact names (`dev-<sha>.zip` etc.) are also unchanged for the same reason.

Everything was tested in my fork, including the updated release processes.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


Daribon:
You don't need to hardcode the database filename world_full_14_june_2021 or clone the entire database repo.

Instead you can just make it download whatever is the latest using the github API, in case there ever is a new DB release. As the file names contain month names, gotta use date -d.

Here is what I've used myself:
`LATEST_DB=$(curl -s https://api.github.com/repos/brotalnia/database/contents/ | grep -o '"name": "world_full_[^"]*\.7z"' | sed 's/"name": "\(.*\)"/\1/' | while read f; do date_part=$(echo "$f" | sed 's/world_full_\(.*\)\.7z/\1/'); echo "$(date -d "$(echo "$date_part" | sed 's/_/ /g')" +%s 2>/dev/null) $f"; done | sort -n | tail -n1 | cut -d' ' -f2-) && wget -qO "/tmp/$LATEST_DB" "https://github.com/brotalnia/database/raw/master/$LATEST_DB"`

Just an explanation what it does:
```
# 1. Fetch directory listing from GitHub API
LATEST_DB=$(
  curl -s https://api.github.com/repos/brotalnia/database/contents/ \
  
  # 2. Extract filenames matching world_full_*.7z
  | grep -o '"name": "world_full_[^"]*\.7z"' \
  | sed 's/"name": "\(.*\)"/\1/' \
  
  # 3. Convert each filename into: "<epoch> <filename>"
  | while read f; do
      date_part=$(echo "$f" | sed 's/world_full_\(.*\)\.7z/\1/')
      epoch=$(date -d "$(echo "$date_part" | sed 's/_/ /g')" +%s 2>/dev/null)
      echo "$epoch $f"
    done \
  
  # 4. Sort by epoch, pick the newest
  | sort -n \
  | tail -n1 \
  
  # 5. Strip the epoch, keep only the filename
  | cut -d' ' -f2-
)

# 6. Download the newest DB archive
wget -qO "/tmp/$LATEST_DB" \
  "https://github.com/brotalnia/database/raw/master/$LATEST_DB"
```

mserajnik:
> You don't need to hardcode the database filename 

This is actually by choice here. The workflow needs to use a reproducible base world DB, as there might be a new dump that is not reflected in the code here yet (in regards to migrations), which would cause the workflow runs to fail.

Yes, that might be rare (I actually asked ratkos about this [recently](https://github.com/vmangos/core/commit/f14d110264719ac8d9fe76f5d2ba367afe3a4cc0#commitcomment-172328917)), but I monitor this actively anyway, in case ratkos forgets to adjust the dump name accordingly then. 😄

> or clone the entire database repo.

That is true, good suggestion. I will adjust this.

mserajnik:
~~Found some things that can be simplified that I missed when cleaning up my code for the PR yesterday.
Will tackle them later today, marking as draft in the meantime.~~
Should be good now, added concurrency guards for the release workflows as well.

NickTyrer:
Please could you add linux binaries to the releases?

mserajnik:
> Please could you add linux binaries to the releases?

Like we talked in #2979?
We can do it, but I don't think we can statically link all the dependencies, realistically. I'll look into it.

NickTyrer:
That would be great, dont need dependencies linked though. Can just run the binaries in a container with pinned ubuntu version.

PR #3318, opened 2026-03-31
Fix the Hunter bot from returning to melee range after getting into it once

IamTirion:
## 🍰 Pullrequest
In the current version, once a Hunter bot gets into melee range, it retreats, then tries to get back to melee range. This PR fixes that. The Hunter bot will now stay at proper range after retreating. The range they stay at can be lower or higher, but that's up for debate.

The problem is that somewhere in the code, the Hunter bot is made to execute the MoveChase function without specifying the distance, which brings it to melee range. And then it executes the RunAwayFromTarget function again, making it a loop.

### How2Test
Bring a Hunter bot to melee range to a Training Dummy on GM Island, then order it to attack.


ratkosrb:
This should be resolved now.

PR #3329, opened 2026-04-04
Fix Imp Fire Shield owner targeting while grouped.

### LINKED ISSUE #2916: 🐞 [Bug] Imp doesn't seem to cast Fire Shield on Warlock that summoned him
## 🐞 Bug report - Imp doesn't cast Fire Shield on Warlock that summoned him

__Edit: This issue probably only happens when using a GM account; see a bit further below for an explanation.__

### Expected behavior
A Warlock's Imp doesn't seem to auto-cast [Fire Shield (Rank 1)](https://www.wowhead.com/classic/spell=2947/fire-shield) on the Warlock that summoned him. When Fire Shield is set to auto-cast, it should cast it on the Warlock (or a party member of the Warlock in range) when they get melee attacked.

Being in a party or not doesn't seem to alter this behavior (as in, the Imp won't cast it on the Warlock in either case). However, the Imp __does__ cast Fire Shield on party members of the Warlock as soon as a mob is about to attack the party member in melee (__Edit:__ actually, this doesn't seem to be limited to melee attacks; I've also observed the Imp casting Fire Shield on party members that were attacked by ranged and ranged magic).

I've only tested this with Fire Shield (Rank 1) so far, so I can't say if higher ranks behave the same.

---

__Edit:__ This seems to be [related to the usage of a GM account](https://github.com/vmangos/core/issues/2916#issuecomment-2678361454), or more specifically the GM visibility setting, with either a straight up bug in that system or a behavior discrepancy between Fire Shield and other things that are affected by GM visibility:

- There seem to be two issues/bugs when playing with a GM account, that might or might not be connected
- The first one is that (with default `GM.*` settings in the config) auto-casting Fire Shield on the summoning Warlock doesn't work unless `.gm visible on` is explicitly used, which doesn't seem necessary for other gameplay systems affected by GM visibility (such as mobs being able to attack you). This makes the setting inconsistent and confusing and implies a "third" (default) state in between an explicit use of `.gm visible on` or `.gm visible off` exists
- The second one is that even after using `.gm visible on` explicitly  (or setting `GM.Visible=1` in the config so it becomes the default), once the summoning Warlock joins a party, auto-casting Fire Shield on him stops working again, which persists even after leaving the party; relogging makes it work again, implying that joining the party changes something that isn't reset until a relog

---

### Steps to reproduce
1. Make sure all `GM.*` settings in the configuration are set to their defaults (you can set `GM.CheatGod=0` so mobs can actually damage you, but that doesn't have any impact on reproducing this issue)
2. Create a Warlock character and __don't use__ `gm. <on>/<off>` or `.gm visible <on>/<off>` at this point
3. Observe that everything during normal gameplay seems to behave as if you were not using a GM account; mobs can attack and damage you, other players can see and buff you etc.
4. Summon an Imp and make sure he has learned Fire Shield (Rank 1) via [Grimoire of Fire Shield (Rank 1)](https://www.wowhead.com/classic/item=16326/grimoire-of-fire-shield-rank-1)
5. Set Fire Shield to auto-cast
6. Engage any mob that will attack in melee
7. Observe that the Imp doesn't auto-cast Fire Shield on you
8. Use `.gm visible on`
9. Observe that the Imp will now correctly auto-cast Fire Shield on you
10. Join a party
11. Observe that the Imp now stops auto-casting Fire Shield on you again (but will auto-cast Fire Shield just fine on party members)
12. Leave the party again
13. Observe that the Imp will still not auto-cast Fire Shield on you (implying that joining the party changed something that is not reset even after leaving it)
14. Relog
15. Observe that the Imp will once again correctly auto-cast Fire Shield on you (which is due to the explicit usage of `.gm visible on` before)

### Version & Environment
Client Version: 1.12.1.5875

Commit Hash: https://github.com/vmangos/core/commit/fe6fcb493687c61f0c6e57e0f96670c99cd5af0c

OS Client: Win 11 VM on macOS
OS Server: Ubuntu 24.04 LTS 

### Crashlog

- None

mserajnik:
## 🍰 Pullrequest
Currently, `PetAI::UpdateAllies()` skips the owner (= the summoning Warlock in this case) in the grouped-member loop here: https://github.com/vmangos/core/blob/09a9a84ae949b6016053258624f7498a6e3b3a02/src/game/AI/PetAI.cpp#L382-L383
That would be fine if the owner were added unconditionally elsewhere. However, in the current code the owner is only inserted in the non-grouped branch here: https://github.com/vmangos/core/blob/09a9a84ae949b6016053258624f7498a6e3b3a02/src/game/AI/PetAI.cpp#L388-L389
That makes the Imp stop considering its summoning Warlock as a valid friendly autocast target while grouped.

In addition, the cached `m_AllySet` can also stay stale after leaving a 2-player party, because a 2-entry set can still be `{pet, former party member}` (since the owner was never added in the grouped branch) versus `{pet, owner}` after leaving the party, and the size-based `m_AllySet.size() == 2 && !group` check will incorrectly still pass.

This change always inserts the owner before processing grouped allies. This fixes both the grouped case and the stale cache after leaving a 2-player party, because even with a 2-player party the size of `m_AllySet` will be 3 (`{pet, owner, party member}`) and thus not hit the early return anymore.

Note that I could not reproduce the GM visibility issue around Fire Shield that I first reported in #2916 anymore, so I probably initially misinterpreted it while simultaneously hitting this party bug.

### Proof
- None

### Issues
- Fixes #2916

### How2Test
- Create a Warlock character using a GM account
- `.levelup 13` (need to be level 14 to learn Fire Shield)
- `.learn all_myclass`
- `.additem 16326 1`
- Cast `Summon Imp`
- Learn Fire Shield by right-click the `Grimoire of Fire Shield (Rank 1)` in the inventory
- Enable Fire Shield autocasting by right clicking on it in the pet action bar
- `.partybot add priest`
- Keep the party bot out of combat and attack a mob
- Observe that the Imp will still cast Fire Shield on you, despite being in a party (this is not the case without this fix)
- Leave the party, wait at least 10 seconds, retest without relogging, and confirm you can still receive Fire Shield (this also does not work without relogging without this fix)

### Todo / Checklist
- Fire Shield currently also autocasts when not attacked in melee (as soon as you get aggro) because its conditional in `Spell::CanAutoCast()` uses `GetAttackerForHelper()`. I am not entirely sure, but I think this is not fully blizz-like. I didn't touch this in this PR though.

### DIFF
--- a/src/game/AI/PetAI.cpp
+++ b/src/game/AI/PetAI.cpp
@@ -371,7 +371,7 @@ void PetAI::UpdateAllies()
 
     m_AllySet.clear();
     m_AllySet.insert(m_creature->GetObjectGuid());
-    m_AllySet.insert(owner->GetObjectGuid()); // Owner must always be a valid friendly target
+    m_AllySet.insert(owner->GetObjectGuid()); // The pet owner must always be a valid friendly target.
 
     if (group)
     {
--- a/src/game/AI/PetAI.cpp
+++ b/src/game/AI/PetAI.cpp
@@ -366,12 +366,14 @@ void PetAI::UpdateAllies()
         return;
 
     //owner is in group; group members filled in already (no raid -> subgroupcount = whole count)
-    if (group && !group->isRaidGroup() && m_AllySet.size() == (group->GetMembersCount() + 2))
+    if (group && !group->isRaidGroup() && m_AllySet.size() == (group->GetMembersCount() + 1))
         return;
 
     m_AllySet.clear();
     m_AllySet.insert(m_creature->GetObjectGuid());
-    if (group)                                             //add group
+    m_AllySet.insert(owner->GetObjectGuid()); // Owner must always be a valid friendly autocast target
+
+    if (group)
     {
         for (GroupReference* itr = group->GetFirstMember(); itr != nullptr; itr = itr->next())
         {
@@ -385,8 +387,6 @@ void PetAI::UpdateAllies()
             m_AllySet.insert(target->GetObjectGuid());
         }
     }
-    else                                                    //remove group
-        m_AllySet.insert(owner->GetObjectGuid());
 }
 
 void PetAI::KilledUnit(Unit* victim)
@@ -523,7 +523,7 @@ std::pair<Unit*, ePetSelectTargetReason> PetAI::SelectNextTarget() const
     Unit* owner = m_creature->GetCharmerOrOwner();
     if (!owner)
         return std::make_pair(nullptr, PSTR_FAIL_NO_OWNER);
-    
+
     if (Creature const* pOwnerCreature = owner->ToCreature())
     {
         // Owner is creature and is evading. We must not re-aggro.
@@ -552,7 +552,7 @@ std::pair<Unit*, ePetSelectTargetReason> PetAI::SelectNextTarget() const
     {
         if (Unit* pVictim = owner->GetVictim())
         {
-            if (!pVictim->HasAuraPetShouldAvoidBreaking() && 
+            if (!pVictim->HasAuraPetShouldAvoidBreaking() &&
                (!m_creature->GetCharmInfo()->IsAtStay() || m_creature->CanReachWithMeleeAutoAttack(pVictim)))
                 return std::make_pair(pVictim, PSTR_SUCCESS_OWNER_VICTIM);
         }
@@ -727,7 +727,7 @@ bool PetAI::CanAttack(Unit* target)
     // CC - mobs under crowd control can be attacked if owner commanded
     if (target->HasAuraPetShouldAvoidBreaking())
         return m_creature->GetCharmInfo()->IsCommandAttack();
-        
+
     // Returning - pets ignore attacks only if owner clicked follow
     if (m_creature->GetCharmInfo()->IsReturning())
         return !m_creature->GetCharmInfo()->IsCommandFollow();
@@ -801,4 +801,3 @@ void PetAI::AttackedBy(Unit* attacker)
     // Continue to evaluate and attack if necessary
     AttackStart(attacker);
 }
-
--- a/src/game/AI/PetAI.cpp
+++ b/src/game/AI/PetAI.cpp
@@ -369,9 +369,11 @@ void PetAI::UpdateAllies()
     if (group && !group->isRaidGroup() && m_AllySet.size() == (group->GetMembersCount() + 1))
         return;
 
+    // Cache potential friendly targets here. Charmed owner/group members are
+    // filtered later by spell target validation before the pet actually casts.
     m_AllySet.clear();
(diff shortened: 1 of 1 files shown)


src/game/AI/PetAI.cpp
```diff
@@ -371,7 +371,7 @@ void PetAI::UpdateAllies()
 
     m_AllySet.clear();
     m_AllySet.insert(m_creature->GetObjectGuid());
-    m_AllySet.insert(owner->GetObjectGuid()); // Owner must always be a valid friendly target
+    m_AllySet.insert(owner->GetObjectGuid()); // The pet owner must always be a valid friendly target.
```
ratkosrb: If the owner is charmed it will not be a friendly target.

mserajnik: Good point, I forgot about that.

This makes the set size checks questionable because a lot can change within 10 seconds in regards to party composition and party members getting charmed.

I will have a look at this later and try to come up with a good solution.

mserajnik: Had a few minutes and checked how CMaNGOS does it and it looks like they also add the owner unconditionally: https://github.com/cmangos/mangos-classic/blob/db3c5cba390b8f02c709675780bc086dc88fec37/src/game/AI/BaseAI/PetAI.cpp#L429

But maybe they check if the potential ally is charmed later; will investigate once I am home, if they have a good solution it may be worth it aligning our handling here.

mserajnik: I looked into this:

Unless I misunderstood something when reading the code, with this PR we are essentially matching current CMaNGOS behavior: the owner is added unconditionally to `m_AllySet` and a charmed owner or party members are not filtered there. Instead, the charm state is validated later at cast time. I added a comment in `UpdateAllies()` to make that explicit.

In VMaNGOS, friendly-target pet casts go through `Spell::CheckPetCast()`, which rejects hostile targets via `IsHostileTo()`, so a charmed target is filtered there rather than when adding them to `m_AllySet`. CMaNGOS handles this in the same way conceptually as well, but through `CanAssistSpell()` / `CanAssist()` instead.

If we wanted to, we could additionally rebuild `m_AllySet` earlier when a charmed party member is detected and explicitly exclude charmed targets from the set, but I am not sure if that is worth adding here; since `m_AllySet` is only rebuilt every 10 seconds, that would mostly just invert the stale cache problem: a temporarily charmed target would stay excluded until the next set rebuild even after the charm ends. And since in that time frame also other charms can happen, we can not blindly rely on the set being correct anyway.

Unrelated, but I also noticed that CMaNGOS likely has a bug in the `GetMembersCount() + 2` check, because `GetMembersCount()` already includes the owner there as well, just like in VMaNGOS (so it should probably be `+ 1` instead there as well).

mserajnik: Anything you'd like me to change here? Would be great to get this bug fixed.

PR #3336, opened 2026-04-05
Makes the casters stop using wand forever

IamTirion:
## 🍰 Pullrequest
Currently, once a caster runs out of mana, they will use the wand forever, even if their mana is fully regenerated. This changes it so that they will return to casting spells once their mana reaches 20%

### How2Test
Get a caster to attack a Training Dummy on GM Island until it switches to wand.


0blu:
There is a missing `}`. The code currently does not compile.

PR #3343, opened 2026-04-06
Fix log format string mismatches and suppress script errors with SF_GENERAL_SKIP_MISSING_TARGETS

schell244:
## 🍰 Pullrequest
- Fix several log statements had mismatched format strings and arguments
- Script commands respawn/open/close door and despawn gameobject now suppress the missing target error when SF_GENERAL_SKIP_MISSING_TARGETS is set, as sometimes this is expected behavior

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None


NickTyrer:
I think the Gordo script needs updating so that he targets a random active gloom weed every x mins. When I did the script I copied a sniff verbatim without realising he just does a random gloom weed target.

PR #3346, opened 2026-04-06
fix typo in mangosd.conf

Wall-core:
this typo was somehow breaking BG queues.


mserajnik:
There is no typo, just whitespace changes?
Whitespace is ignored when parsing the config files (both for the config name and the value), so that shouldn't be the cause of the issue: https://github.com/vmangos/core/blob/b3a91b6ee66b1d1013a2d286a3955dd457abaf3f/src/shared/Config/Config.cpp#L125

Wall-core:
you would assume so, but BG queues were completely broken until removing the whitespace. I have no clue, but it's working afterwards.

Gamemechanicwow:
this seems weird..

- i can't reproduce it. 
- config values that can't be read are hardcoded to use their default values - the same values shown in your image on the bugtracker
- a lot of other config settings have trailing spaces

And if there is an issue here, my opinion is thats not within the conf file - its in the reader

Wall-core:
this could very well be a linux specific issue. I can't reproduce locally on win10 but it affects ashenwow (ubuntu server)

mserajnik:
> this could very well be a linux specific issue

I can't repro it either using my Docker setup (which is Ubuntu 24.04 LTS based).

Are you sure the whitespace you deleted to make it work was actually `U+0020` and not some other Unicode characters that might have rendered as space in the editor? Of course, that would also raise the question how such characters could've gotten into the file in place of the actual spaces. 😅

mserajnik:
Already in bed, but I asked Copilot (using Claude Sonnet 4.6) in the GitHub app on my phone because I was curious, and it claims to have found a legitimate bug (haven't checked it myself yet):

image

It also says non-`U+0020` whitespace would cause the code fallbacks to be used:

image

I haven't looked at the code defaults, but if those match the config file defaults, then that couldn't be the cause if the issue either, of course.

But in any case, seems like it would be a good idea to improve the parser to properly trim config key and value of all common whitespace instead of just handling literal `U+0020`.

Wall-core:
It's possible that the in-use config's encoding was modified when opened and resaved in Nano, no clue how but all I could guess.

Issue #3344, opened 2026-04-06
🐛 [Bug] battleground queue issues


Wall-core:
currently there's some issues with BG queues. I've not been able to attach a debugger and see what's breaking here directly, but queues don't fill properly at all currently. Some BGs remain open with no players, BGs will fill 5v1 and won't allow anyone else in, and several other problems.

<img width="497" height="561" alt="Image" src="https://github.com/user-attachments/assets/55289aba-6eae-42f9-ad02-766367767abe" />

Wall-core:
I just realized it may be the config, it has extra spaces after the variable name

<img width="357" height="84" alt="Image" src="https://github.com/user-attachments/assets/a8a586e6-2112-4898-9cfa-211383a02f54" />

Gamemechanicwow:
whitespaces are being skipped afaik

Wall-core:
you would assume so, but BG queues were completely broken until removing the whitespace. I have no clue, but it's working afterwards.

PR #3347, opened 2026-04-07
Normalize config whitespace.

mserajnik:
## 🍰 Pullrequest

#3344 reported a bug that apparently came from having more than one leading space in front of the `=`, which potentially accidentally got replaced by some other Unicode whitespace character when editing it, which the config parser doesn't ignore (it only skips `U+0020` spaces).

As a follow-up to #3346, this normalizes all remaining config lines in `mangosd.conf.dist.in` to have exactly one space before and after the `=`:
```
<key> = <value>
```

There were no cases that don't already match this in `realmd.conf.dist.in`.

__Note:__ I would still look into improving the parser so it can handle non-`U+0020` whitespace properly, but imo it makes sense to unify this regardless, as it appears that the current indentation wasn't used consistently anyway.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/mangosd/mangosd.conf.dist.in
+++ b/src/mangosd/mangosd.conf.dist.in
@@ -74,19 +74,19 @@ RealmID = 1
 DataDir = "."
 LogsDir = ""
 HonorDir = ""
-LoginDatabase.Info              = "127.0.0.1;3306;mangos;mangos;realmd"
-LoginDatabase.Connections       = 1
-LoginDatabase.WorkerThreads     = 1
-WorldDatabase.Info              = "127.0.0.1;3306;mangos;mangos;mangos"
-WorldDatabase.Connections       = 1
-WorldDatabase.WorkerThreads     = 1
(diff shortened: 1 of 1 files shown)


0blu:
I am rooting for an actual fix of the config reader 👍

Wall-core:
If you're interested in testing, you can try and replicate the issue and then examine the end file in 010 to check each byte.
What I had done essentially was clone the repo and build the server on ubuntu. I edited the config file in Nano several times, saving it each time. I subsequently moved the files via powershell SCP to my local Win10 system so I could diff the config in Notepad++ and update it to the newest revision. After applying the updates in Notepad++ and saving, I sent the file back over powershell SCP to the ubuntu server, and subsequently edited and saved again in Nano several times.

This likely messed with some encoding, at least that's my assumption. The same thing happened previously when moving SQL dumps in the same manner

mserajnik:
> I am rooting for an actual fix of the config reader 👍 

Maybe you should write it, considering your contributions here you seem like one of the most qualified people to do it. 😄

PR #3361, opened 2026-04-11
refactor(Packet): Add ServerPacket structure

0blu:
Converts some of the manually built `WorldPacket` (which were sent by the server) to nicely structured `ServerPackets`.

Due to the large amount of changes required this will be split into multiple PRs.

### DIFF
--- a/src/game/Battlegrounds/BattleGroundMgr.h
+++ b/src/game/Battlegrounds/BattleGroundMgr.h
@@ -70,11 +70,11 @@ enum BattleGroundQueueGroupTypes
 };
 #define BG_QUEUE_GROUP_TYPES_COUNT 4
 
-enum BattleGroundGroupJoinStatus
+enum BattleGroundGroupJoinStatus : uint32
 {
-    BG_GROUPJOIN_DESERTERS = -2,
-    BG_GROUPJOIN_FAILED = -1    // actually, any negative except 2
-                                // any other value is a MapID meaning successful join
+    BG_GROUPJOIN_DESERTERS = static_cast<uint32>(-2),
+    BG_GROUPJOIN_FAILED = static_cast<uint32>(-1) // actually, any negative except 2
+    // any other value is a MapID meaning successful join
 };
 
 class BattleGround;
--- a/src/game/Battlegrounds/BattleGroundMgr.h
+++ b/src/game/Battlegrounds/BattleGroundMgr.h
@@ -72,9 +72,9 @@ enum BattleGroundQueueGroupTypes
 
 enum BattleGroundGroupJoinStatus : uint32
 {
-    BG_GROUPJOIN_DESERTERS = static_cast<uint32>(-2),
-    BG_GROUPJOIN_FAILED = static_cast<uint32>(-1) // actually, any negative except 2
-    // any other value is a MapID meaning successful join
+    BG_GROUPJOIN_DESERTERS = 0xFFFFFFFE,
+    BG_GROUPJOIN_FAILED = 0xFFFFFFFF,
+    // any other value is a MapID (meaning successful join)
 };
 
 class BattleGround;
--- a/src/game/Handlers/PetitionsHandler.cpp
+++ b/src/game/Handlers/PetitionsHandler.cpp
@@ -174,26 +174,14 @@ void WorldSession::HandlePetitionQueryOpcode(WorldPackets::Petition::QueryPetiti
     if (!petition)
         return;
 
-    WorldPacket data(SMSG_PETITION_QUERY_RESPONSE, (4 + 8 + petition->GetName().size() + 1 + 2 + 4 * 11));
-    data << uint32(packet.petitionGuid);                    // int m_petitionID;
-    data << ObjectGuid(petition->GetOwnerGuid());           // unsigned __int64 m_petitioner;
-    data << petition->GetName();                            // char m_title[256];
-    data << uint8(0);                                       // char m_bodyText[4096];
-    data << uint32(1);                                      // int m_flags;
-    data << uint32(9);                                      // int m_minSignatures;
-    data << uint32(9);                                      // int m_maxSignatures;
-    data << uint32(0);                                      // int m_deadLine;
-    data << uint32(0);                                      // int m_issueDate;
-    data << uint32(0);                                      // int m_allowedGuildID;
-    data << uint32(0);                                      // int m_allowedClasses;
-    data << uint32(0);                                      // int m_allowedRaces;
-    data << uint16(0);                                      // __int16 m_allowedGender;
-    data << uint32(0);                                      // int m_allowedMinLevel;
-    data << uint32(0);                                      // int m_allowedMaxLevel;
-    data << uint32(0);                                      // char m_choicetext[10][64];
-    // for (int i=0; i<field13; ++i) data << chartSignersName[i];   Probably, names of the petition signers
-    data << uint32(0);                                      // int m_numChoices;
-    SendPacket(&data);
+    auto response = std::make_unique<WorldPackets::Petition::PetitionQueryResponse>();
+    response->petitionGuid = packet.petitionGuid;
+    response->ownerGuid = ObjectGuid(petition->GetOwnerGuid());
+    response->name = petition->GetName();
+    response->flags = 1;
+    response->minSignatures = 9;
+    response->maxSignatures = 9;
+    SendPacket(std::move(response));
 }
 
 void WorldSession::HandlePetitionRenameOpcode(WorldPackets::Petition::PetitionRename const& packet)
@@ -220,10 +208,10 @@ void WorldSession::HandlePetitionRenameOpcode(WorldPackets::Petition::PetitionRe
 
     if (petition->Rename(packet.newName))
     {
-        WorldPacket data(MSG_PETITION_RENAME, (8 + packet.newName.size() + 1));
-        data << ObjectGuid(packet.itemGuid);
-        data << packet.newName;
-        SendPacket(&data);
+        auto renameResult = std::make_unique<WorldPackets::Petition::PetitionRenameResult>();
+        renameResult->itemGuid = packet.itemGuid;
+        renameResult->newName = packet.newName;
+        SendPacket(std::move(renameResult));
     }
 }
 
@@ -243,11 +231,11 @@ void WorldSession::HandlePetitionSignOpcode(WorldPackets::Petition::PetitionSign
 
     if (petition->GetOwnerGuid() == _player->GetObjectGuid())
     {
-        WorldPacket data(SMSG_PETITION_SIGN_RESULTS, (8 + 8 + 4));
-        data << ObjectGuid(packet.itemGuid);
-        data << ObjectGuid(_player->GetObjectGuid());
-        data << uint32(PETITION_SIGN_CANT_SIGN_OWN);
-        SendPacket(&data);
+        auto signPacket = std::make_unique<WorldPackets::Petition::PetitionSignResults>();
+        signPacket->itemGuid = packet.itemGuid;
+        signPacket->playerGuid = _player->GetObjectGuid();
+        signPacket->result = PETITION_SIGN_CANT_SIGN_OWN;
+        SendPacket(std::move(signPacket));
         return;
     }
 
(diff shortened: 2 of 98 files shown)


src/game/Handlers/PetitionsHandler.cpp
```diff
@@ -500,23 +494,14 @@ void WorldSession::SendPetitionShowList(ObjectGuid guid)
...
-    SendPacket(&data);
+    auto showList = std::make_unique<WorldPackets::Petition::PetitionShowList>();
+    showList->npcGuid = guid;
+    showList->count = 1;
+    showList->index = 1;
+    showList->charterEntry = GUILD_CHARTER;
+    showList->charterDisplayId = CHARTER_DISPLAY_ID;
+    showList->charterCost = GUILD_CHARTER_COST;
+    showList->unknown = 1;
+    SendPacket(std::move(showList));
```
ratkosrb, line 505: Shouldn't we have the packet class reflect the structure the client reads? There should be a vector inside this class to contain the variable count substructure.

src/game/Handlers/QueryHandler.cpp
```diff
@@ -36,57 +36,34 @@ void WorldSession::SendNameQueryOpcode(Player* p)
...
 #if SUPPORTED_CLIENT_BUILD >= CLIENT_BUILD_1_12_1
-    WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + 25 + 1 + 4 + 4 + 4));   // guess size
-    data << ObjectGuid(p->GetObjectGuid());
-    data << p->GetName();                                   // CString(48): played name
-    data << "";                                             // CString(256): realm name for cross realm BG usage
-#else
-    WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + 25 + 4 + 4 + 4));   // guess size
-    data << ObjectGuid(p->GetObjectGuid());
-    data << p->GetName();                                   // CString(48): played name
+    nameResponse->realmName = "";                           // realm name for cross realm BG usage
```
ratkosrb: Shouldn't it default to "" anyway? Why do we need to set it here.

src/game/Handlers/QueryHandler.cpp
```diff
@@ -36,57 +36,34 @@ void WorldSession::SendNameQueryOpcode(Player* p)
...
 #if SUPPORTED_CLIENT_BUILD >= CLIENT_BUILD_1_12_1
-        WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + (name.size() + 1) + 1 + 4 + 4 + 4));
-        data << ObjectGuid(HIGHGUID_PLAYER, pData->uiGuid);
-        data << name;                                       // CString(48): played name
-        data << "";                                         // CString(256): realm name for cross realm BG usage
-#else
-        WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + (name.size() + 1) + 4 + 4 + 4));
-        data << ObjectGuid(HIGHGUID_PLAYER, pData->uiGuid);
-        data << name;
+        nameResponse->realmName = "";
```
ratkosrb: Same, no need to manually set std::string to "", that's what it is by default.

src/game/Handlers/QueryHandler.cpp
```diff
@@ -112,22 +89,17 @@ void WorldSession::SendNameQueryOpcodeFromDBCallBack(QueryResult* result, uint32
...
 #if SUPPORTED_CLIENT_BUILD >= CLIENT_BUILD_1_12_1
-    WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + (name.size() + 1) + 1 + 4 + 4 + 4));
-    data << ObjectGuid(HIGHGUID_PLAYER, lowguid);
-    data << name;
-    data << "";                                            // realm name for cross realm BG usage
-#else
-    WorldPacket data(SMSG_NAME_QUERY_RESPONSE, (8 + (name.size() + 1) + 4 + 4 + 4));
-    data << ObjectGuid(HIGHGUID_PLAYER, lowguid);
-    data << name;
+    nameResponse->realmName = "";
```
ratkosrb: No need to set to "" again.

src/game/Server/Packets/Spell.h
```diff
@@ -65,6 +65,21 @@ namespace WorldPackets { namespace Spell
         explicit CancelChanneling() : ClientPacket(CMSG_CANCEL_CHANNELLING) {}
         void ReadFromWorldPacket(WorldPacket& recv_data) override;
     };
+    // --- Server Packets ---
+
+    // Simplified SMSG_CAST_RESULT for a basic spell failure (status=2) with no extra data.
+    // For complex failures with additional payload (e.g. cooldown time, required area),
+    // use the full Spell::SendCastResult() path in Spell.cpp instead.
+    class CastResultSimpleFailure final : public ServerPacket
```
ratkosrb, line 75: Why don't we pass the Spell* to the packet so it can build the whole structure?

src/game/Server/Packets/Spell.cpp
```diff
@@ -34,3 +34,10 @@ void WorldPackets::Spell::CancelChanneling::ReadFromWorldPacket(WorldPacket& rec
 {
     recv_data >> spellId;
 }
+
+void WorldPackets::Spell::CastResultSimpleFailure::AppendBodyTo(ByteBuffer& buffer) const
+{
+    buffer << spellId;
+    buffer << uint8(2); // status = fail
```
ratkosrb: should be an enum tbh

src/game/Handlers/SkillHandler.cpp
```diff
@@ -36,27 +36,24 @@ void WorldSession::HandleLearnTalentOpcode(WorldPackets::Skill::LearnTalent cons
...
         GetPlayer()->RemoveSpellsCausingAura(SPELL_AURA_FEIGN_DEATH);
 
-    if (!(_player->ResetTalents()))
+    if (!_player->ResetTalents()) // TODO Move money check into this function and remove optional argument from `ResetTalents`
     {
-        WorldPacket data(MSG_TALENT_WIPE_CONFIRM, 8 + 4);   //you have not any talent
-        data << uint64(0);
-        data << uint32(0);
-        SendPacket(&data);
+        GetPlayer()->SendTalentWipeConfirm(ObjectGuid::Empty); // "You have not spent any talent points"
```
ratkosrb, line 52: Is this really more efficient than calling ObjectGuid() ?

src/game/Handlers/TaxiHandler.cpp
```diff
@@ -124,13 +124,12 @@ bool WorldSession::SendLearnNewTaxiNode(Creature* unit)
...
-        SendPacket(&msg);
+        SendPacket(std::make_unique<WorldPackets::Taxi::NewTaxiPath>());
 
-        WorldPacket update(SMSG_TAXINODE_STATUS, 9);
-        update << ObjectGuid(unit->GetObjectGuid());
-        update << uint8(1);
-        SendPacket(&update);
+        auto taxiStatus = std::make_unique<WorldPackets::Taxi::TaxiNodeStatus>();
+        taxiStatus->guid = unit->GetObjectGuid();
+        taxiStatus->known = 1;
```
ratkosrb: why not bool

src/game/Server/Packets/Channel.cpp
```diff
@@ -90,3 +90,9 @@ void WorldPackets::Channel::ChannelModerate::ReadFromWorldPacket(WorldPacket& re
 {
     recv_data >> channelName;
 }
+
+void WorldPackets::Channel::ChannelNotify::AppendBodyTo(ByteBuffer& buffer) const
+{
+    buffer << uint8(type);
```
ratkosrb: In some cases you cast the packet to it's type before writing to buffer, other times you do not. Let's keep it consistent.

src/game/Server/Packets/GmTicket.h
```diff
@@ -3,6 +3,7 @@
 
 #include "Packet.h"
 #include "SharedDefines.h"
+#include "nonstd/optional.hpp"
```
ratkosrb: Why do you include in this file when it's not used by it?

src/game/Server/Packets/Group.h
```diff
@@ -145,6 +148,85 @@ namespace WorldPackets { namespace Group
     };
 #endif
 
+    // --- Server Packets ---
+
+    static constexpr uint8 PARTY_MAX_POSITIVE_AURAS = 32;  // MAX_POSITIVE_AURAS
+    static constexpr uint8 PARTY_MAX_NEGATIVE_AURAS = 16;  // MAX_AURAS - MAX_POSITIVE_AURAS
```
ratkosrb: why do we hardcode this here instead of using the actual defines

src/game/Server/Packets/Petition.cpp
```diff
@@ -65,3 +65,60 @@ void WorldPackets::Petition::PetitionBuy::ReadFromWorldPacket(WorldPacket& recv_
...
+    buffer << uint32(0);                                    // deadLine
+    buffer << uint32(0);                                    // issueDate
+    buffer << uint32(0);                                    // allowedGuildID
+    buffer << uint32(0);                                    // allowedClasses
+    buffer << uint32(0);                                    // allowedRaces
+    buffer << uint16(0);                                    // allowedGender
+    buffer << uint32(0);                                    // allowedMinLevel
+    buffer << uint32(0);                                    // allowedMaxLevel
+    buffer << uint32(0);                                    // choicetext
+    buffer << uint32(0);                                    // numChoices
```
ratkosrb: I think we should represent the full structure the client reads in the packet class, and not hardcode zeroes.

src/game/Server/Packets/Petition.cpp
```diff
@@ -65,3 +65,60 @@ void WorldPackets::Petition::PetitionBuy::ReadFromWorldPacket(WorldPacket& recv_
...
+
+void WorldPackets::Petition::PetitionShowList::AppendBodyTo(ByteBuffer& buffer) const
+{
+    buffer << npcGuid;
+    buffer << count;
+    buffer << index;
+    buffer << charterEntry;
+    buffer << charterDisplayId;
+    buffer << charterCost;
+    buffer << unknown;
```
ratkosrb: have you tried taking a peak at how this is used to determine it's possible meaning?

src/game/Battlegrounds/BattleGroundMgr.h
```diff
@@ -70,11 +70,11 @@ enum BattleGroundQueueGroupTypes
 };
 #define BG_QUEUE_GROUP_TYPES_COUNT 4
 
-enum BattleGroundGroupJoinStatus
+enum BattleGroundGroupJoinStatus : uint32
```
ratkosrb, line 73: why is this uint32 if we use negative values

src/game/Server/Packets/Misc.h
```diff
@@ -370,6 +370,174 @@ namespace WorldPackets { namespace Misc
...
+        void AppendBodyTo(ByteBuffer& buffer) const override;
+    };
+
+    class MeetingstoneSetQueue final : public ServerPacket
+    {
+    public:
+        uint32 areaId = 0;
+
+#if SUPPORTED_CLIENT_BUILD <= CLIENT_BUILD_1_4_2
+        uint64 idempotencyToken = 0; // Guess: Incrementing counter. Must change for every response. TODO: Maybe use ms timestamp to enforce a new id on every packet
```
ratkosrb, line 532: Why not make a static uint64 we just increment on write?

ratkosrb:
For cases where the client assigns the same handler to multiple packets, reading the same structure, we should use the same packet class for them as well and just pass opcode in constructor, rather than creating duplicate structures.

src/game/Server/Packets/Spell.h
0blu: I fear `.reload`

src/game/Server/Packets/Petition.cpp
0blu: Must be `& 1` to show the UI.  🤷 
image

src/game/Handlers/SkillHandler.cpp
0blu: TC also does this too.
I think it does not make any difference performance wise and its easier to read and understand the intent of the code.

src/game/Server/Packets/Group.h
0blu: Woops these changes were not intended for this PR.
I have a larger PR where I cherry picked some changes.

I would need to import `SpellAuraDefines.h` which then gets indirectly imported in a lot of places.
(because `WorldSession.h` has all the packet definitions imported)

src/game/Server/Packets/Misc.h
0blu: Can we/you do this in a separate PR?

0blu:
> For cases where the client assigns the same handler to multiple packets, reading the same structure, we should use the same packet class for them as well and just pass opcode in constructor, rather than creating duplicate structures.

I would like to prevent mistakes in the code.
If the developer can freely choose an opcode, the struct might be invalid.

I've used `NullClientPacket` for empty client packets, which I might want to change too.
Just so every packet is uniquely typed.

ratkosrb:
> If the developer can freely choose an opcode, the struct might be invalid.

That doesn't prevent us from having invalid structs. I don't like this duplication of code. There are a million movement packets with the same structure.

src/game/Server/Packets/Spell.h
ratkosrb: It will not invalidate the pointers. We read the new data into the existing spell entry.
https://github.com/vmangos/core/blob/development/src/game/Spells/SpellMgr.cpp#L3776

src/game/Server/Packets/Misc.h
ratkosrb: ok

0blu:
> > If the developer can freely choose an opcode, the struct might be invalid.
> 
> That doesn't prevent us from having invalid structs. I don't like this duplication of code. There are a million movement packets with the same structure.

Have I even used those Packets in this PR?
I saw some in `Combat.h‎` but all of those seem reasonable. 
If you are satisfied you can squash merge the PR.

PR #3365, opened 2026-04-12
CI: Add MinGW64 build job.

mserajnik:
## 🍰 Pullrequest

__Edit:__ As discussed below, I changed the PR to add a new (modern) MinGW CI build (instead of the previously proposed Windows XP release).

I left the slight Windows packaging refactor since it's imo cleaner, and also the slight naming adjustments in the README and the releases themselves.

Also, more importantly, this also fixes a bug with the build matrix; when I fixed the Clang builds previously, I made a mistake there I only noticed today which caused all Ubuntu builds to use Clang instead of GCC since then; oops.
With this PR, that is also corrected.

### DIFF
--- a/dep/src/g3dlite/GThread.cpp
+++ b/dep/src/g3dlite/GThread.cpp
@@ -12,6 +12,12 @@
 #include "G3D/debugAssert.h"
 #include "G3D/GMutex.h"
 
+#ifdef __MINGW32__
+// MinGW needs libseh's SEH macros for this debugger thread-name exception.
+// The include needs to be outside of the G3D namespace.
+#include <seh.h>
+#endif
+
 namespace G3D {
 
 namespace _internal {
@@ -90,10 +96,6 @@ typedef struct tagTHREADNAME_INFO {
    DWORD dwFlags; // Reserved for future use, must be zero.
 } THREADNAME_INFO;
 #pragma pack(pop)
-
-#ifdef __MINGW32__
-#include <seh.h>
-#endif
 
 static void SetThreadName(DWORD dwThreadID, const char* threadName) {
     THREADNAME_INFO info;
--- a/src/shared/ByteBuffer.h
+++ b/src/shared/ByteBuffer.h
@@ -131,7 +131,7 @@ class ByteBuffer
             return *this;
         }
 
-#ifdef __MINGW32__
+#if defined(__MINGW32__) && !defined(__MINGW64__)
         ByteBuffer& operator<<(time_t value)
         {
             append<time_t>(value);
@@ -227,7 +227,7 @@ class ByteBuffer
             return *this;
         }
 
-#ifdef __MINGW32__
+#if defined(__MINGW32__) && !defined(__MINGW64__)
         ByteBuffer& operator >> (time_t& value)
         {
             value = read<time_t>();
--- a/.github/scripts/package-windows-release.sh
+++ b/.github/scripts/package-windows-release.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Copyright (C) 2026-present  VMaNGOS  https://github.com/vmangos
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <dependency-dir-name> <archive-filename>" >&2
+  exit 1
+fi
+
+dep_dir_name="$1"
+archive_filename="$2"
+release_dir="$PWD/bin/Release"
+dep_dir="$PWD/dep/windows/lib/${dep_dir_name}_release"
+archive_path="$PWD/bin/$archive_filename"
+
+if [ ! -d "$release_dir" ]; then
+  echo "Could not find release directory at $release_dir." >&2
+  exit 1
+fi
+
+if [ ! -d "$dep_dir" ]; then
+  echo "Could not find dependency directory at $dep_dir." >&2
+  exit 1
+fi
+
+cp -f "$dep_dir/libmySQL.dll" "$release_dir/libmySQL.dll"
+cp -f "$dep_dir/libeay32.dll" "$release_dir/libeay32.dll"
+
+rm -f "$archive_path"
+
+(
+  cd "$PWD/bin"
+  7z a -tzip "$archive_filename" Release >/dev/null
+)
--- a/.github/scripts/publish-moving-release.sh
+++ b/.github/scripts/publish-moving-release.sh
@@ -27,7 +27,7 @@ release_tag="$1"
 release_title_prefix="$2"
 asset_dir="$3"
 repo_url="https://github.com/$GITHUB_REPOSITORY"
-release_title="$release_title_prefix($(date -u +%F))"
+release_title="$release_title_prefix ($(date -u +%F))"
 notes_file="$(mktemp)"
 
 trap 'rm -f "$notes_file"' EXIT
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -1,4 +1,5 @@
-# Build on Ubuntu with GCC and Clang, plus a Windows Visual Studio build.
+# Build on Ubuntu with GCC and Clang, plus Windows Visual Studio and MinGW64
+# builds.
 name: CI Build
 
 on:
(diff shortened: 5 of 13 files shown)


0blu:
It seems like the version of `libstdc++6.dll` is not supported on Windows XP :/

image

_Image from https://www.dependencywalker.com/_

Edit: @mserajnik I wonder if it works if we statically link those runtime libraires.
Maybe it will recognize that we are not using this function in our code.
https://stackoverflow.com/questions/6404636/libstdc-6-dll-not-found

ratkosrb:
MingW has not supported XP in a decade. You have to use an old version of MinGW32, not 64.

mserajnik:
> MingW has not supported XP in a decade. You have to use an old version of MinGW32, not 64.

That is already the case, MinGW32 is set up through MSYS2.

I see now my comment in regards to uncertainties about future support was poorly worded; it wasn't meant to be about MinGW itself, but the required tooling around it (like MSYS2) to make this viable to do on GitHub Actions. In other words, if MSYS2 dropped MinGW32 support, we would have a problem with this approach.

I think looking for a solution to build with MSVC (ideally 2015) is the better approach, so I will explore that further beyond native runner support provided by GitHub Actions.

@0blu Thanks for testing again, I will still look into this too because I want to get the MinGW approach working, just to satisfy my own curiosity. 😅 But for the purposes of this PR, I will try to figure out a MSVC-based solution (as you suggested perhaps using a container).

ratkosrb:
I have tested the core on XP recently and I didn't need to disable cpptrace for it to run so I'm confused by that.

mserajnik:
> I have tested the core on XP recently and I didn't need to disable cpptrace for it to run so I'm confused by that.

I looked into this more already.

`fopen_s` is a Microsoft CRT extension, which MinGW32 doesn't support.

It targets `msvcrt.dll`, which doesn't have `fopen_s` to begin with (which was what I first thought was the reason why it didn't work), but even if it did, MinGW32 only supports the POSIX C subset and not the custom Microsoft extensions the DLL contains.

So if you use MSVC, `cpptrace` will work, but with MinGW32 it won't (unless we added a `fopen_s` shim/wrapper).

ratkosrb:
We've embedded cpptrace into your repository, we can simply change that line to use fopen not fopen_s.

0blu:
When we continue pursuing this path with MinGW that we cannot detect MSVC specific bugs.
As far as I remember we focus VS2017 support(?)

If we ever decide to update a dependency or introduce new ones like Protobuf (for newer clients) we might get a successful build with MinGW whereas VS2017 would fail.

The whole goal for this CI is to have peace of mind.

ratkosrb:
I use VS 2015 exclusively. So if you break build I'll sound the alarm very quickly :+1:

ratkosrb:
Can you submit the MinGW build fixes as a separate PR?

ratkosrb:
I think you should add a normal MinGW build task to the CI, not tied to Windows XP, just to verify going forward we don't break the build. You cannot really test if the core runs on XP using the CI, because it's possible to compile executables using the XP toolset that don't actually run on XP, since a newer Windows SDK was being used by the compiler, linking to non-existent functions, as was the case with CancelIoEx recently. VS 2015 was building with the XP toolset happily, but core wouldn't run because of missing function in the kernel. We should leave testing XP to users of the project, since I don't think the CI can actually run the exe on XP to verify it works.

mserajnik:
> We've embedded cpptrace into your repository, we can simply change that line to use fopen not fopen_s.

Yes, or like I said, a shim/wrapper:
```c
errno_t fopen_s(FILE **fp, const char *name, const char *mode) {
    *fp = fopen(name, mode);
    return *fp ? 0 : errno;
}
```
(very basic example, no input validation as `fopen_s` does it)
But yes, simply swapping to `fopen`, as you suggest, would probably be good enough here.

I wanted to focus on getting a working build first, and `cpptrace` is an optional tool. So simply disabling it so I can continue to look into other non-working stuff first not related to `cpptrace` (like what @0blu found yesterday) to figure out the feasibility of using MinGW seemed like the better choice than trying to patch it (especially considering this might need more patches than just `fopen_s` that haven't surfaced yet).

---

> I use VS 2015 exclusively. So if you break build I'll sound the alarm very quickly 👍

I wasn't fully aware of the exact constraints when BLU approached me, so the MinGW approach seemed good if the goal was just to build a Windows XP artififact in the CI. But if MSVC 2015 is the target, we should try to somehow get a working CI pipeline with that (and if that proves impossible on the free, GitHub-hosted runners, even with a build container or similar, we can decide then what the other options are).

> Can you submit the MinGW build fixes as a separate PR?

My plan was to swap this PR to a MSVC approach and either drop MinGW entirely (and keep it on my private fork so I can experiment with it further), or I can in addition look into getting the MinGW build working properly as well and make a separate PR (properly = a working Windows XP binary).

What's your preference on how to continue?

mserajnik:
> I think you should add a normal MinGW build task to the CI, not tied to Windows XP, just to verify going forward we don't break the build. You cannot really test if the core runs on XP using the CI, because it's possible to compile executables using the XP toolset that don't actually run on XP, since a newer Windows SDK was being used by the compiler, linking to non-existent functions, as was the case with CancelIoEx recently. VS 2015 was building with the XP toolset happily, but core wouldn't run because of missing function in the kernel. We should leave testing XP to users of the project, since I don't think the CI can actually run the exe on XP to verify it works.

Sorry, was compiling my previous comment and hadn't seen this before submitting.

If that is the preference, that's okay with me as well, and I can look into that. I am fine with any solution, I personally don't need a working XP build. My goal was to help out BLU and make some of his work easier by having a better/automated way to prove building for XP doesn't break.

ratkosrb:
I think we should fix building the project with MinGW and have a CI action for it on it's own, unrelated to XP.

ratkosrb:
It would also be cool to have a CI action for building on MacOS if that's even possible.

mserajnik:
> I think we should fix building the project with MinGW and have a CI action for it on it's own, unrelated to XP.

Do you mean building with MinGW32, still for an XP target (without publish a release), or just modern MinGW64? I assume the latter, but just to be sure.

I would repurpose this PR then for that, if that's fine.

> It would also be cool to have a CI action for building on MacOS if that's even possible.

It should be, we can use Clang on macOS. I will look into this too.

ratkosrb:
Whatever you think is best.

mserajnik:
I have looked a bit more into whether we could get a working XP build via MinGW32 after all, but have concluded it isn't worth pursuing.

The core problem is `libstdc++-6.dll` (as surfaced by BLU's testing). I did some research and it turns out that modern `libstdc++` (as shipped by current MinGW-w64) pulls in Vista+ APIs internally (`EncodePointer`, `GetTickCount64`, `InitializeConditionVariable`, etc.). Static linking sadly won't fix this, the import references will still be part of the final binary regardless. We'd have to rebuild `libstdc++` itself against an XP-only API surface to make this feasible. Note that `mingw-w64-i686` only specifies the 32-bit target, the toolchain itself is current (and using an actually old, XP-era MinGW version isn't really a feasible approach, I think).

MSYS2 also deprecated `mingw-w64-i686` in 2024, which I had missed when I initially created the PR, so this wouldn't be a future-proof solution either.

And as pointed out by ratkos, even if we did manage to build an XP artifact, the CI can't prove it actually boots on XP, as seen with [`CancelIoEx`](https://github.com/vmangos/core/commit/379e9ced36ed468629d9ae03f24998238e4f4d9a) (shows that the toolchain will happily link against APIs that XP doesn't have).
                            
So, as suggested, I'll repurpose this PR to add a MinGW64 job to the existing CI matrix, mirroring the GCC, Clang, and MSVC setups.

And I'll look into macOS builds later in a separate PR.

0blu:
Thanks for analyzing the problem in this depth.
Without a qemu vm, we are probably unable to use VS2015 or something similar.

If you like you can just add the MinGW workflow for modern Windows and MacOS in a separate PR.

mserajnik:
> Without a qemu vm, we are probably unable to use VS2015 or something similar.

Yeah, when I looked into MSVC 2015 I also noticed it's not publicly downloadable anymore; you apparently have to login to https://my.visualstudio.com to download it (if it's even available there still, I only see 2017 and newer on https://visualstudio.microsoft.com/vs/older-downloads/).

I _think_ that means we couldn't legally use it, because obviously we'd have to grab the installer somehow and bake it into our own image to use it in any automated CI context (container, QEMU, whatever). But since any official download (if still available) is gated behind a login, this means accepting ToS and puts us (at best) in a licensing grey area for redistributing MS proprietary binaries, which would be questionable even for a private repo, and probably is out of the question for a public project (that is also hosted on a Microsoft-owned platform).

mserajnik:
As discussed, repurposed this PR to add a MinGW64 build job instead (see also updated PR description).

Ready for review.

dep/src/g3dlite/GThread.cpp
```diff
@@ -12,6 +12,12 @@
 #include "G3D/debugAssert.h"
 #include "G3D/GMutex.h"
 
+#ifdef __MINGW32__
```
mserajnik, line 15: I kept the `#ifdef` as `__MINGW32__` since it's more defensive and we have existing ones, even though `__MINGW64__` would work for the CI job too. There is no `__MINGW__`, `__MINGW32__` applies to both MinGW32 and MinGW64.

src/shared/ByteBuffer.h
```diff
@@ -131,7 +131,7 @@ class ByteBuffer
             return *this;
         }
 
-#ifdef __MINGW32__
+#if defined(__MINGW32__) && !defined(__MINGW64__)
```
mserajnik, line 134: `__MINGW32__` is also defined on MinGW64 builds, where `time_t` is a typedef for `int64` and collides with the existing `int64` overloads. Adding `&& !defined(__MINGW64__)` restricts these overloads to MinGW32 where `time_t` was actually distinct. This could also be dropped entirely of course if we don't care about MinGW32, but I didn't want to break someone's potentially existing build setup.

ratkosrb:
To reduce the amount of CI actions, we could have the old version build tests be only for 1 of the compilers, instead of for 3.

mserajnik:
Good idea, I think that's reasonable.
I changed it so that <1.12.1 only builds with GCC (imo the most reasonable choice for this).

I also added 1.4.2 builds here since I saw there is basic support for it, so it's probably good to start catching compilation errors for this version too; I can also drop this if you don't want it yet, of course.

PR #3384, opened 2026-04-28
Remove unused WheatyExceptionReport

0blu:


0blu:
https://github.com/vmangos/core/pull/3386#issuecomment-4337837318 @ratkosrb 
> What does WheatyExceptionReport do?

Very old Windows specific mini-library depending on dbghelp.
Was used to print stacktraces but I believe it never worked on vMangos.

As far as I understand we could do the same with `cpptrace` by hooking an exception handler at the start of the program.

image


(could also do the same on Linux with `signal(SIGSEGV, handler)` and `signal(SIGABRT, handler)`)

0blu:
I could also try to reactive it.  
It does not print the stacktrace in the console, but rather creates a file in the `Crash` folder.

PR #3386, opened 2026-04-28
Enable strict checks of `char const*`

0blu:
## 🍰 Pullrequest

Blocked by #3384 

I also had to remove `--no-warnings` from the build, because it hid some of these errors.


ratkosrb:
What does WheatyExceptionReport do?

PR #3391, opened 2026-05-03
Enable all warnings. Mark them as error. Hide unfixed errors.

0blu:
## 🍰 Pullrequest
Using `-Wall` and `-Werror` to improve code quality and find hidden bugs.
Currently all unfixed errors are hidden which represents the behaviour before this PR.
We can now be sure to not introduce any new types of errors and progressively remove the ignored errors.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -204,7 +204,7 @@ if(MSVC)
   #  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LARGEADDRESSAWARE")
   #endif()
 
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /bigobj")
+  set(MANGOS_CXX_FLAGS "${MANGOS_CXX_FLAGS} /bigobj")
 endif()
 
 set(CMAKE_SKIP_BUILD_RPATH FALSE)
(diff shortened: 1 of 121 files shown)


ratkosrb:
I think using Werror is kind of insane. A warning does not mean the code is broken. It would lock the project to specific versions of compilers and reduce compatibility.

0blu:
If the code is well written then there are no warnings/errors. `-Werror` is best practice.

ratkosrb:
And as I predicted the build is broken on VS 2015 with a million warnings that are now errors. I cannot even post them all. There are too many errors.

ratkosrb:
https://pastebin.com/hessM7Va

ratkosrb:
I notice Wall causes even default header files that come with Visual Studio to throw warnings.

ratkosrb:
image

example

ratkosrb:
Fixing all warnings you see is good and it's nice you've tried to do this, but enabling Wall and Werror is only going to cause problems for people trying to build the core on a compiler you didn't test. This project tries to have as low compiler and OS requirements as we can. We should not make it harder for people to compile the project.

0blu:
Actually I also feared that older Visual Studio versions would emit more warnings.

I've adjusted the rules so that only newer version will treat warnings as errors.
We should be able to support modern compilers, otherwise we can remove the `/Wx` flag when someone complains.

Without this I fear that important warnings may be overlooked.
I've already fixed some of these warnings as part of this PR.

ratkosrb:
Now it compiles without errors for me on VS 2015.

ratkosrb:
This PR introduces some warnings on SocketConnector.cpp that didn't exist before.

SocketConnector.cpp(81): warning C4548: expression before comma has no effect; expected expression with side-effect
SocketConnector.cpp(85): warning C4548: expression before comma has no effect; expected expression with side-effect
SocketConnector.cpp(89): warning C4548: expression before comma has no effect; expected expression with side-effect

PR #3394, opened 2026-05-05
Add Linux development release.

mserajnik:
## 🍰 Pullrequest
This adds a Linux development release (`amd64`-only) similar to the existing Windows release, [as requested](https://github.com/vmangos/core/pull/3316#issuecomment-4276136278) by @NickTyrer.

The release is built on Ubuntu 24.04 and published as `dev-linux-amd64-<sha>.tar.gz` under a new `linux_latest` moving release tag.

I decided to make the release somewhat portable instead of relying on system dependencies, but it is not fully statically linked. Instead, the package bundles non-`glibc` runtime libraries in `lib/` and uses [PatchELF](https://github.com/NixOS/patchelf) to set relative `RPATH`s, so binaries under `bin/` can find the bundled libraries after extraction.

`glibc` and the dynamic loader are not bundled, so the archive is expected to work on `amd64` systems with `glibc` 2.39 or newer. Alpine/`musl` and older `glibc` systems are not supported.

The workflow includes a small smoke test (which I can also remove if unwanted) that extracts the archive in a fresh Ubuntu 24.04 Docker container, runs `realmd --version` and `mangosd --version`, and checks the packaged ELF files with `ldd`. ~~I have not done a full runtime test with a configured database and extracted client data, so I marked this as draft for now. (@NickTyrer: if you could try this, it would be great.)~~

__Edit:__ I did some testing and for me everything worked, Nick tested too, so this should be fine functionality-wise, removed the draft status.

### Proof
- None

### Issues
- Closes #2979

### How2Test
1. Download the test release from [my fork](https://github.com/mserajnik/vmangos-core/releases/tag/linux_latest).
2. Extract it and run from the archive root:
  ```sh
  cd vmangos-linux-amd64
  ```
3. Copy and edit the config files:
  ```sh
  cp etc/realmd.conf.dist etc/realmd.conf
  cp etc/mangosd.conf.dist etc/mangosd.conf
  ```
4. Adjust DB credentials and other settings as required for your setup.
5. Put extracted client data under the archive root, matching the default `DataDir = "."` (or put it elsewhere and adjust `DataDir` accordingly):
  ```
  vmangos-linux-amd64/
    maps/
    mmaps/
    vmaps/
    5875/
      dbc/
  ```
6. Start `realmd` and `mangosd` from the archive root:
  ```sh
  ./bin/realmd
  ./bin/mangosd
  ```

### Todo / Checklist

- [x] Test if it actually works

### DIFF
--- a/.github/scripts/package-linux-release.sh
+++ b/.github/scripts/package-linux-release.sh
@@ -1,0 +1,213 @@
+#!/usr/bin/env bash
+
+# Copyright (C) 2026-present  VMaNGOS  https://github.com/vmangos
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+  echo "Usage: $0 <install-dir> <archive-filename>" >&2
+  exit 1
+fi
+
+install_dir="$1"
+archive_filename="$2"
+package_name="vmangos-linux-amd64"
+stage_parent="$PWD/bin/linux-release"
+package_dir="$stage_parent/$package_name"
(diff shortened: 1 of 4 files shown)


NickTyrer:
Brilliant, I will test this weekend. Thanks

NickTyrer:
Tested and working fine, thanks.

0blu:
Since it is built with OpenSSL3 it is currently ""blocked"" by https://github.com/vmangos/core/pull/3402

0blu:
@mserajnik is it possible to have just one tag?
Right now we have `latest` and `linux_latest`

mserajnik:
> @mserajnik is it possible to have just one tag?
> 
> Right now we have `latest` and `linux_latest`

Possible, but requires quite a bit of restructuring.

I can tackle that, but will probably take a few weeks as I'm busy with other stuff at the moment.

ratkosrb:
And don't forget we also need a MacOS build action! :angel:

PR #3397, opened 2026-05-07
Refactor packet

0blu:

### DIFF
--- a/src/game/Commands/DebugCommands.cpp
+++ b/src/game/Commands/DebugCommands.cpp
@@ -2370,21 +2370,19 @@ bool ChatHandler::HandleDebugPvPCreditCommand(char *args)
     If uiHonorValue=0 : "Dishonorable Kill"
     */
     auto packet = std::make_unique<WorldPackets::Misc::PvpCredit>();
-
     if (pSelection->GetTypeId() == TYPEID_PLAYER)
     {
         uint32 uiHonorValue = urand(1, 100);
         packet->honor = uiHonorValue;
-        packet->victimGuid = pSelection->GetObjectGuid();
         PSendSysMessage("Honorable Kill : Rank %3u and Honor %3u.", uiRankValue, uiHonorValue);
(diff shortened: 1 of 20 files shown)


src/game/Commands/DebugCommands.cpp
```diff
@@ -2368,24 +2374,24 @@ bool ChatHandler::HandleDebugPvPCreditCommand(char *args)
...
+        packet->honor = uiHonorValue;
+        packet->victimGuid = pSelection->GetObjectGuid();
         PSendSysMessage("Honorable Kill : Rank %3u and Honor %3u.", uiRankValue, uiHonorValue);
     }
     else // Victoire deshonorante
     {
-        data << uint32(0);
-        data << pSelection->GetGUID();
+        packet->honor = 0;
+        packet->victimGuid = pSelection->GetObjectGuid();
```
ratkosrb: since we assign the same guid in same cases, shouldn't the assignment be outside of the if statement

src/game/World.cpp
```diff
@@ -2379,23 +2388,22 @@ void World::SendGMText(int32 string_id, ...)
...
     // need copy to prevent corruption by strtok call in LineFromMessage original string
     char* buf = mangos_strdup(text);
     char* pos = buf;
 
     while (char* line = ChatHandler::LineFromMessage(pos))
     {
-        ChatHandler::BuildChatPacket(data, CHAT_MSG_SYSTEM, line);
-        SendGlobalMessage(&data, self);
+        WorldPacket binaryPacket;
+        ChatHandler::BuildChatPacket(binaryPacket, CHAT_MSG_SYSTEM, line);
```
ratkosrb, line 2398: Should the logic in ChatHandler::BuildChatPacket be moved to builder of packet class?

0blu: Good idea. I will refactor this when I touch the chat packet stuff.

PR #3402, opened 2026-05-10
License: Re-add OpenSSL exception

0blu:
## 🍰 Pullrequest

This PR restores the OpenSSL exception that was already present in the initial MaNGOS commit.
It appears to have been omitted when the vMaNGOS repository was published.
_(TrinityCore had a similar issue for a while which was later corrected.)_

_MaNGOS Zero_ first commit:
https://github.com/mangoszero/server/blob/386f5374ab11bf650ed6ef29202ce7a482f1e968/README.md?plain=1#L133-L137

_CMaNGOS_ first commit:
https://github.com/cmangos/mangos-classic/blob/ea63fbe103d9242bae4c4e3933bf0e7a1bada8d6/README#L19-L27

_MaNGOS_ first commit:
https://github.com/mangos/MaNGOS/blob/ea63fbe103d9242bae4c4e3933bf0e7a1bada8d6/README#L19-L27

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -337,3 +337,14 @@ proprietary programs.  If your program is a subroutine library, you may
 consider it more useful to permit linking proprietary applications with the
 library.  If this is what you want to do, use the GNU Lesser General
 Public License instead of this License.
+
+                            OpenSSL exception
+In addition, as a special exception, the vMaNGOS project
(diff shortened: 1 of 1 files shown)


mserajnik:
Good catch!

PR #3403, opened 2026-05-10
Add case-insensitive starts-with helper

0blu:

### DIFF
--- a/src/shared/Common.h
+++ b/src/shared/Common.h
@@ -87,9 +87,6 @@ typedef std::chrono::time_point<std::chrono::system_clock, std::chrono::millisec
 
 #else
 
-#  define stricmp strcasecmp
-#  define strnicmp strncasecmp
-
 #  define I32FMT "%08X"
 #  if ACE_SIZEOF_LONG == 8
 #    define I64FMT "%016lX"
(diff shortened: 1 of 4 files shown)


src/shared/Common.h
```diff
@@ -109,6 +106,19 @@ typedef std::chrono::time_point<std::chrono::system_clock, std::chrono::millisec
 /// Will always return a finite float. If the provided float is infinite it will return 0
 inline float finiteAlways(float f) { return std::isfinite(f) ? f : 0.0f; }
 
+/// Case-insensitive check whether str starts with prefix
+inline bool StringStartsWithCaseInsensitive(std::string const& str, std::string const& prefix)
+{
+    std::size_t const n = prefix.length();
+    if (str.size() < n)
```
ratkosrb: why mixing .size and .length instead of being consistent, aren't they the same thing?

0blu: Yes those are the same. I've changed the code a bit and forgot the change it.

0blu: Maybe I should use AI to catch those errors beforehand.

src/shared/Common.h
```diff
@@ -109,6 +106,19 @@ typedef std::chrono::time_point<std::chrono::system_clock, std::chrono::millisec
...
+inline bool StringStartsWithCaseInsensitive(std::string const& str, std::string const& prefix)
+{
+    std::size_t const n = prefix.length();
+    if (str.length() < n)
+        return false;
+#if COMPILER == COMPILER_MICROSOFT
+    return _strnicmp(str.c_str(), prefix.c_str(), n) == 0;
+#else
+    return strncasecmp(str.c_str(), prefix.c_str(), n) == 0;
+#endif
```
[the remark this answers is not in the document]
0blu: @copilot apply changes based on [this feedback](https://github.com/vmangos/core/pull/3403#discussion_r3229673809)

PR #3404, opened 2026-05-10
Move random related function into own file

0blu:


ratkosrb:
There should be no *Defines.cpp, move implementation of the functions from CreatureDefines.cpp to Creature.cpp

PR #3406, opened 2026-05-11
Refactor `Unit` related ServerPackets

0blu:

### DIFF
--- a/src/game/Objects/Player.cpp
+++ b/src/game/Objects/Player.cpp
@@ -19231,10 +19231,7 @@ void Player::SendInitialPacketsBeforeAddToMap()
     // Homebind
     {
         auto packet = std::make_unique<WorldPackets::Misc::BindpointUpdate>();
-        packet->x = m_homebind.x;
-        packet->y = m_homebind.y;
-        packet->z = m_homebind.z;
-        packet->mapId = m_homebind.mapId;
+        packet->location = m_homebind;
         packet->areaId = m_homebindAreaId;
         GetSession()->SendPacket(std::move(packet));
     }
--- a/src/game/Objects/Player.cpp
+++ b/src/game/Objects/Player.cpp
@@ -767,8 +767,7 @@ uint32 Player::EnvironmentalDamage(EnvironmentalDamageType type, uint32 damage)
         sLog.Out(LOG_BASIC, LOG_LVL_DEBUG, "We are fall to death, loosing 10 percents durability");
         DurabilityLossAll(0.10f, false);
         // durability lost message
-        WorldPacket data2(SMSG_DURABILITY_DAMAGE_DEATH, 0);
-        GetSession()->SendPacket(&data2);
+        GetSession()->SendPacket(std::make_unique<WorldPackets::Misc::DurabilityDamageDeath>());
     }
 
     return damage;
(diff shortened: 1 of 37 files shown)


src/game/WorldStates.h
```diff
@@ -165,14 +165,4 @@ inline void WriteInitialWorldStatePair(ByteBuffer& data, uint16 state, int16 val
     data << value;
 }
 
-#if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_8_4
-inline void WriteUpdateWorldStatePair(ByteBuffer& data, uint32 state, int32 value)
-#else
-inline void WriteUpdateWorldStatePair(ByteBuffer& data, uint16 state, int16 value)
```
ratkosrb, line 171: Here both the state and value are 16 bit, but in the class you added above only the state is conditionally written as 16 bit.

ratkosrb: 1.8
image

1.9
image

1.10
image

1.11
image

1.12
image

0blu: Thanks for the decompilations👍

ratkosrb:
There is a merge conflict.

ratkosrb:
There is a merge conflict once again.

src/game/Objects/Player.cpp
```diff
@@ -19232,16 +19222,22 @@ void Player::SetGroup(Group* group, int8 subgroup)
...
-    data << m_homebind.x << m_homebind.y << m_homebind.z;
-    data << (uint32) m_homebind.mapId;
-    data << (uint32) m_homebindAreaId;
-    GetSession()->SendPacket(&data);
+    {
+        auto packet = std::make_unique<WorldPackets::Misc::BindpointUpdate>();
+        packet->x = m_homebind.x;
+        packet->y = m_homebind.y;
+        packet->z = m_homebind.z;
+        packet->mapId = m_homebind.mapId;
```
ratkosrb: Why not make the packet contain WorldLocation structure so we can assign m_homebind directly?

src/game/Server/Packets/Pet.h
```diff
@@ -142,6 +142,71 @@ namespace WorldPackets { namespace Pet
...
+
+#if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_5_1
+    class PetMode final : public ServerPacket
+    {
+    public:
+        ObjectGuid petGuid;
+        uint8 reactState = 0;
+        uint8 commandState = 0;
+        uint8 flag1 = 0;        // always 0
+        uint8 enabledFlags = 0; // 0x0 = enabled, 0x8 = disabled
```
ratkosrb, line 182: Would be interesting if we can reverse what the other flags are, must be others except 0x8.

src/game/Server/Packets/Pet.h
```diff
@@ -142,6 +142,71 @@ namespace WorldPackets { namespace Pet
...
+#endif
+
+#if SUPPORTED_CLIENT_BUILD > CLIENT_BUILD_1_5_1
+    class PetMode final : public ServerPacket
+    {
+    public:
+        ObjectGuid petGuid;
+        uint8 reactState = 0;
+        uint8 commandState = 0;
+        uint8 flag1 = 0;        // always 0
```
ratkosrb, line 181: Is it unused in client?

src/game/Server/Packets/Pet.h
ratkosrb: In client these 4 uint8s are read as one uint32 and there's functions that mask specific bits and call them "mode" and "orders". The variable they're saved in is called CGPetInfo::m_petMode I believe.

ratkosrb: This specific flags seems to be checked when calling Script::GetPetActionInfo for index 6. Maybe the others are unused :/

0blu: Yeah, I briefly looked into that. It looks like a lot of lua functions are using this bitset.

Unfortunately I cant dig deeper into this right now. If you are really interested, remind me in a few weeks again.

PR #3408, opened 2026-05-14
CI: Consolidate Linux and Windows releases into one tag.

mserajnik:
## 🍰 Pullrequest
This consolidates the two existing development releases for Linux and Windows into a single `latest` tag ([as requested](https://github.com/vmangos/core/pull/3394#issuecomment-4446125757) by @0blu).

Since I watch all activity here and got slightly annoyed by the two separate notifications for the releases, I decided to take a look right away after all, and it turned out to be much less work than I had anticipated since the [`publish-moving-release.sh`](https://github.com/vmangos/core/blob/c3919aa8e80e7173fceef3f3163a6e2adb4f878d/.github/scripts/publish-moving-release.sh) helper script didn't need any adjustments. 😄

Tested end-to-end in my fork: https://github.com/mserajnik/vmangos-core/releases/tag/latest

I renamed the Windows archive to `dev-windows-amd64-<short hash>.zip` so they both match in naming (in theory a breaking change for any third party using some sort of automation to grab this if done in a brittle manner, but probably rare if done at all, and not aligning the naming wouldn't have made sense to me). The Linux archive is still a tarball instead of a ZIP since that is the more common archive type there.

**Once this is merged, the existing `linux_latest` release (and the accompanying tag) can/need to be deleted manually** to make it disappear (I decided not to include a cleanup in this PR, as it would only run once and then be dead code).

Unrelated to the consolidation, I also unified the `cmake` invokes in the workflows to be multi-line since it's nicer to read and only the prior Linux release workflow was like that already.

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


mserajnik:
@ratkosrb:

image

Might want to delete this now outdated release (as mentioned in the PR description) so it doesn't confuse people.

PR #3409, opened 2026-05-14
CI: Add macOS CI build.

mserajnik:
## 🍰 Pullrequest
This adds a macOS CI build ([as requested](https://github.com/vmangos/core/pull/3394#issuecomment-4448009942) by @ratkosrb), running on `macos-26` (Apple Silicon only; Intel Macs will be abandoned with macOS 27 this year anyway) for client version `1.12.1.5875`, matching the existing Ubuntu Clang and Windows builds.

Split into two commits:
- __[`Fix macOS build.`](https://github.com/vmangos/core/commit/0433255243e7ab1c4675ce31b3e3f256b57b01c2)__: code and CMake changes needed to compile on macOS with `-Werror`:
  - [`cmake/find/FindMySQL.cmake`](https://github.com/vmangos/core/blob/0433255243e7ab1c4675ce31b3e3f256b57b01c2/cmake/find/FindMySQL.cmake): use the `PATHS` keyword properly in `find_program` (the old syntax was silently treating paths as additional names; it only worked on Linux because `/usr/bin` is in the default search) and add Homebrew search paths (for both Apple Silicon and Intel, so Intel Mac users should still be able to compile on older macOS systems), mirroring what [`FindOpenSSL.cmake`](https://github.com/vmangos/core/blob/0433255243e7ab1c4675ce31b3e3f256b57b01c2/cmake/find/FindOpenSSL.cmake) already does.
  - [`cmake/platform/unix/settings.cmake`](https://github.com/vmangos/core/blob/0433255243e7ab1c4675ce31b3e3f256b57b01c2/cmake/platform/unix/settings.cmake): `STREQUAL "Clang"` was changed to `MATCHES "Clang"` so Apple's `AppleClang` compiler ID doesn't end up using the GCC branch and pick up GCC-only flags like `-Wno-nonnull-compare`.
  - [`dep/include/nlohmann/json.hpp`](https://github.com/vmangos/core/blob/0433255243e7ab1c4675ce31b3e3f256b57b01c2/dep/include/nlohmann/json.hpp): Made two whitespace-only changes because the PR runner image funnily got updated from Xcode 26.2 / AppleClang 17 to Xcode 26.4.1 / AppleClang 21 within the ~30 minutes between my fork run and the PR run, and AppleClang 21 enables `-Wdeprecated-literal-operator` by default.
  - All `sprintf` / `strcpy` / `strcat` call sites across `src/` and `contrib/` have been rewritten to `snprintf` / `memcpy`. Apple's SDK marks these with `__deprecated_msg`, so `-Werror,-Wdeprecated-declarations` fails the build otherwise. The replacements should be safe, but worth reviewing closely (I threw Claude Opus 4.7 Max at these replacements).
- __[`CI: Add macOS CI build.`](https://github.com/vmangos/core/commit/a6d01e71513425ef82a7a381f21d12bc5ae4a6dc)__: adds the actual CI build.

The alternative to the `sprintf`/`strcpy`/`strcat` rewrites would be suppressing the deprecation on Apple Clang only by adding `-Wno-deprecated-declarations` in `cmake/platform/unix/settings.cmake` gated on `CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang"` and drop the relevant changes from the first commit. Happy to do that instead if preferred.
__Edit:__ and the same thing for `-Wdeprecated-literal-operator` since the PR runner image got updated.

Tested end-to-end in my fork: https://github.com/mserajnik/vmangos-core/actions/runs/25850039327

### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/shared/ServiceWin32.cpp
+++ b/src/shared/ServiceWin32.cpp
@@ -60,7 +60,8 @@ bool WinServiceInstall()
         return false;
     }
 
-    std::strcat(path, " -s run");
+    size_t pathLen = strlen(path);
+    snprintf(path + pathLen, sizeof(path) - pathLen, " -s run");
 
     SC_HANDLE service = CreateService(serviceControlManager,
         serviceName,                                // name of service
--- a/.github/workflows/ci-build.yaml
+++ b/.github/workflows/ci-build.yaml
@@ -65,6 +65,10 @@ jobs:
               job_name: Ubuntu Clang
           - client_build: 5875
             config:
+              os: macos-26
+              job_name: macOS Clang
+          - client_build: 5875
+            config:
               os: windows-2022
               job_name: Windows Visual Studio
           - client_build: 5875
@@ -92,6 +96,10 @@ jobs:
       - name: Install Clang compiler
         if: matrix.config.install_clang
         run: sudo apt install -y --no-install-recommends clang
+
+      - name: Install macOS build dependencies
+        if: matrix.config.os == 'macos-26'
+        run: brew install mysql-client
 
       - name: Set up MSYS2 (MINGW64)
         if: matrix.config.use_mingw
@@ -122,6 +130,21 @@ jobs:
             -DBUILD_WARNINGS_AS_ERROR=1 \
             -DSUPPORTED_CLIENT_BUILD="${{ matrix.client_build }}"
           make -j"$(nproc)"
+          make install
+
+      - name: Build and install on macOS
+        if: matrix.config.os == 'macos-26'
+        run: |
+          set -euo pipefail
+          mkdir -p build _install
+          cd build
+          cmake ../ \
+            -DCMAKE_INSTALL_PREFIX="../_install" \
+            -DBUILD_EXTRACTORS=1 \
+            -DENABLE_MAILSENDER=1 \
+            -DBUILD_WARNINGS_AS_ERROR=1 \
+            -DSUPPORTED_CLIENT_BUILD="${{ matrix.client_build }}"
+          make -j"$(sysctl -n hw.ncpu)"
           make install
 
       - name: Build and install on Windows (Visual Studio)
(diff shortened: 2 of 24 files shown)


mserajnik:
Looking into why the PR run failed when it worked in my fork. 👀

__Edit:__ Found it, the runner image got updated between my fork run and the PR run:

  - __Fork run:__ Xcode 26.2 / AppleClang 17
  - __PR run:__ Xcode 26.4.1 / AppleClang 21

AppleClang 21 added `-Wdeprecated-literal-operator`, so need some more changes.

src/shared/ServiceWin32.cpp
```diff
@@ -60,7 +60,8 @@ bool WinServiceInstall()
         return false;
     }
 
-    std::strcat(path, " -s run");
+    size_t pathLen = strlen(path);
+    snprintf(path + pathLen, sizeof(path) - pathLen, " -s run");
```
ratkosrb, line 64: By the name of this file I would expect that it shouldn't get compiled at all on MacOS.

mserajnik: Yep, want me to drop it or keep it for consistency?

mserajnik: > Yep, want me to drop it or keep it for consistency?

` std::strcat()` would still warn (and thus error with `-Werror`) if it were outside of a Windows-only context, so imo it would make sense to keep everything consistent and not have different ways of doing basic things that will work on some platforms/compilers but not on others (even if in this particular case it doesn't matter since it's gated in `#ifdef`).

ratkosrb:
Should we also have a release published for MacOS or would those binaries not be portable? Also since we are building on clang with MacOS, perhaps the Linux clang CI action is redundant?

mserajnik:
> Should we also have a release published for MacOS or would those binaries not be portable?

Should be possible in a similar manner to the Linux release I believe. I will look into it next week or so.

> Also since we are building on clang with MacOS, perhaps the Linux clang CI action is redundant?

I would leave that as it is still useful to have a CI build with regular Clang on Linux. Apple Clang is built with different options and defaults and as such is not a great general indicator of _"can VMaNGOS be compiled with Clang?"_.

PR #3410, opened 2026-05-14
Update G3DLite

0blu:
## 🍰 Pullrequest
This PR will upgrade G3DLite to have improved compatibility with newer compilers and reducing the amount of warnings/errors.

We are now in sync with TrinityCore's version of G3DLite and can more easily apply future updates or fixes when new issues arise.


dep/g3dlite/source/Ray.cpp
```diff
@@ -21,6 +21,105 @@ void Ray::set(const Vector3& origin, const Vector3& direction) {
     debugAssert(direction.isUnit());
     
     m_invDirection = Vector3::one() / direction;
+    
+    // ray slope
```
Gamemechanicwow: unused code. it was intentionally removed.

dep/g3dlite/include/G3D/UprightFrame.h
```diff
@@ -74,14 +74,15 @@ class UprightFrame {
...
+/** 
     \brief Shortest-path linear velocity spline for camera positions.  Always keeps the camera from rolling.
 
     \sa G3D::UprightSplineManipulator, G3D::UprightFrame
 */
 class UprightSpline : public Spline<UprightFrame> {
 protected:
-    void ensureShortestPath(UprightFrame* A, int N) const override {
+
+    virtual void ensureShortestPath(UprightFrame* A, int N) const {
```
Gamemechanicwow: why not override? 

/home/runner/work/core/core/dep/g3dlite/include/G3D/UprightFrame.h:85:18: warning: 'ensureShortestPath' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
   85 |     virtual void ensureShortestPath(UprightFrame* A, int N) const {

0blu: Hmm, interesting that TrinityCore did not fix this. I am not sure what to do about this.

Gamemechanicwow: i had the same thought reading this PR. Should changes to our g3dlite code be redirected to TC?

do we really need a 1:1 copy?

0blu: No, we don't need a 1:1 copy but its easier to maintain.  
I also created two vmangos related patches.

dep/g3dlite/source/Ray.cpp
0blu: I've removed the code again and documented it in a diff file.

dep/g3dlite/include/G3D/UprightFrame.h
0blu: I've disabled all the warnings for g3d internals to avoid build log spams as its not part of the code we actively maintain.  
It's been field-tested for many years, and as long as we don't modify it, it should remain stable.

Gamemechanicwow: > This PR will upgrade G3DLite to have improved compatibility with newer compilers and reducing the amount of warnings/errors.

also

> I've disabled all the warnings for g3d internals

if we keept the code as-is - we dont have to disable warning, which we shouldnt have to in the first place.

> It's been field-tested for many years, and as long as we don't modify it, it should remain stable.

just like the current implementation.

0blu: The current code is not compatible with C++20.

0blu: And the warnings I had were from vmangos related files which has `#include` statements to G3D causing build errors at `-Werror`.

TrinityCore is also built using `-Werror` and they fixed those publicly visible parts of G3D (that's why the PR).

I've just disabled the internal warnings that G3D had.

Gamemechanicwow: i dont see that as a reason to introduce warnings. also im pretty sure that "override" exists in c++20

0blu: You are right 😄 I've added another patch.

ratkosrb: This project is not moving to c++20...

dep/g3dlite/source/GThread.cpp
```diff
@@ -12,12 +12,6 @@
 #include "G3D/debugAssert.h"
 #include "G3D/GMutex.h"
 
-#ifdef __MINGW32__
-// MinGW needs libseh's SEH macros for this debugger thread-name exception.
-// The include needs to be outside of the G3D namespace.
-#include <seh.h>
-#endif
```
ratkosrb, line 19: Are you sure you didn't just break mingw32 support?

0blu: Now it works.

PR #3414, opened 2026-05-18
Refactor `Unit` related packets with `SpellEntry`

0blu:

### DIFF
--- a/src/game/Objects/MirrorTimer.h
+++ b/src/game/Objects/MirrorTimer.h
@@ -20,6 +20,8 @@
 #include "Common.h"
 #include "Timer.h"
 
+class SpellEntry;
+
 class MirrorTimer
 {
 public:
@@ -43,7 +45,7 @@ public:
         STATUS_UPDATE = 2,
     };
 
-    MirrorTimer(Type type) : m_type(type), m_scale(-1), m_spellId(0), m_status(UNCHANGED), m_active(false), m_frozen(false) {}
+    MirrorTimer(Type type) : m_type(type), m_scale(-1), m_maybeSpellEntry(nullptr), m_status(UNCHANGED), m_active(false), m_frozen(false) {}
 
     bool     IsActive() const { return m_active; }
     bool     IsRegenerating() const { return (m_scale > 0); }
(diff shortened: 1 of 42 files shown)


0blu:
I have to rebase it <.<

src/game/Objects/MirrorTimer.h
```diff
@@ -74,7 +78,7 @@ class MirrorTimer
 private:
     Type m_type;
     int32 m_scale;
-    uint32 m_spellId;
+    SpellEntry const* m_maybeSpellEntry; // might be nullptr
```
ratkosrb: The naming is strange, also why does the timer need to store a spell at all?

0blu: I needed a way to communicate the nullablilty of this field without looking at the docs.
In most other cases `SpellEntry` is not nullable.

The field is needed for `StartMirrorTimer`.

src/game/Server/Packets/Misc.h
```diff
@@ -715,6 +717,20 @@ namespace WorldPackets { namespace Misc
...
 
+    class StartMirrorTimer final : public ServerPacket
+    {
+    public:
+        uint32 timerType = 0;   // mirror timer type (breath, fatigue, environmental)
+        uint32 remaining = 0;   // remaining time in ms
+        uint32 duration = 0;    // full timer duration in ms
+        int32 scale = 0;        // rate of change (-1 = decreasing, 10 = increasing)
+        bool paused = false;
+        ::SpellEntry const* maybeSpellEntry = nullptr; // spell causing the timer (nullptr if no spell)
```
ratkosrb: Why store SpellEntry pointer in the packet struct when the only thing sent is spell id?

src/game/Server/Packets/Pet.h
```diff
@@ -207,6 +209,17 @@ namespace WorldPackets { namespace Pet
         void AppendBodyTo(ByteBuffer& buffer) const override;
     };
 
+    class PetCastFailed final : public ServerPacket
+    {
+    public:
+        ::SpellEntry const* spellEntry = nullptr;
```
ratkosrb: Again, I don't like this. You should only store spell entry if you need to pull some other data from it. Otherwise store an uint32 with the id.

0blu: I changed all the spell related packets. I thought that is what you wanted.

I've also noticed that some packets are using uint16 and some uint32.

ratkosrb: I never said to store a SpellEntry pointer in every packet. It only makes sense if the packet pulls multiple things from it.

PR #3416, opened 2026-05-19
Prevents mobs from getting stuck or moving through doors

Wall-core:
This fix prevents units from getting stuck when performing CutPathWithDynamicLoS(). This is a notable bug that typically affects hunter pets getting stuck on mailboxes. It also ensures units do not move through closed doors.
Note: I don't know what units (npcs, pets, or players) shouldn't move through doors in blizzlike. Adjust as necessary.

### DIFF
--- a/src/game/Maps/PathFinder.cpp
+++ b/src/game/Maps/PathFinder.cpp
@@ -1109,12 +1109,15 @@ void PathInfo::CutPathWithDynamicLoS()
     Vector3 out;
     // We have always keep at least 2 points (else, there is no mvt !)
     for (uint32 i = 1; i <= maxIndex; ++i)
+    {
         if (m_sourceUnit->GetMap()->GetDynamicObjectHitPos(m_pathPoints[i - 1], m_pathPoints[i], out, -0.1f))
         {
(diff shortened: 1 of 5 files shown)


Wall-core:
example of new behavior, preventing player AI from running through door:

https://github.com/user-attachments/assets/ec265704-bb0e-4235-925c-c4170b523f77

Gamemechanicwow:
the real problem with LOS on small or narrow objects like doors, is that the dyn tree returns the first found collision instead of the nearest. (it can also happen on larger objects)

For LOS this is fine - for movement its not.

Here is some old code to demonstrate: https://github.com/Gamemechanicwow/core/tree/stopatfirst

Wall-core:
Yeah I had considered sorting through the dyntree in pathbuilding but it uses gameobjectmodel instead of object pointers. This method simply finds the first collision and moves the final point to a safe angle away from the object, which resolves pathing issues due to gobjs while preventing units from slipping through doors

Wall-core:
I'm gonna work on implementing a comprehensive solution to this during path generation time

PR #3430, opened 2026-05-26
Respawn Rogue Vale Screecher (credit cmangos)

NickTyrer:
## 🍰 Pullrequest


ratkosrb:
Some of the waypoints for guid 51211 do not match the terrain. The creature goes underground.

NickTyrer:
That is blizzlike.

ratkosrb:
Have you checked out the creature yourself to verify?

NickTyrer:
yes

mserajnik:
> Some of the waypoints for guid 51211 do not match the terrain. The creature goes underground.

Would it make sense to document such confirmed blizz-like bugs somewhere (meaning somewhere else than fragmented GitHub PR comments) so they will not be misidentified as VMaNGOS bugs/inaccuracies in the future (when official Classic servers might no longer be available to verify against)?

PR #3440, opened 2026-05-29
Guard the honor system behind patch 1.4.0.

### LINKED ISSUE #2744: 🐛 [Bug][WowPatch = 0] The server shuts down to count honor every time it is turned on.
## 🐛 Bug report
After starting the server with the configured world patch 0, the server shuts down after 15 minutes to diagnose the honor system.
This happens on the second and third runs.

### Expected behavior
The server does not restart 15 minutes after startup.

### Steps to reproduce
1. Set WowPatch = 0;
2. Run server
3. Profit

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  No

OS Client: Win
OS Server: Ubuntu

### Crashlog

- None

mserajnik:
## 🍰 Pullrequest

This effectively disables the honor system pre patch `1.4.0` (when `PvP.AccurateTimeline` is also set `true`), including the weekly honor maintenance and all the honor-related GM commands.

A few edge cases not covered on purpose:
- If rank gets set through the database directly, VMaNGOS will still read the rank as-is; this could be tightened to force an in-memory rank of `0` when a higher one is detected with the honor system disabled, but I decided against overriding DB data (happy to add this if wanted).
- If a bot gets copied from a character with existing rank it will still inherit the rank; I think it would be incorrect to prevent the copy of the rank here if the bot is supposed to be a verbatim copy of the character (regardless of how the character acquired the rank).
- If a premade gear template includes an item with a required honor rank, that item is no longer auto-granted the rank and so ends up unequipped in the recipient's bags instead of being skipped outright; this could be tightened to skip such items entirely when the honor system is disabled, but that should only affect custom templates, so it didn't seem worth explicitly covering.

__Note:__ I also removed the now-redundant inline `1.4.0` guards in `Player::RewardHonor()` and `Player::RewardHonorOnDeath()`, since `HonorMgr::Add()` is guarded centrally now.

### Proof

- None

### Issues

- Fixes #2744

### How2Test

- None; not reachable on currently supported patches (>= `1.5.1`); only affects an unsupported pre-`1.4.0` config, so this is effectively just some groundwork for later.

### Todo / Checklist

- [X] None

### DIFF
--- a/sql/migrations/20260529212939_world.sql
+++ b/sql/migrations/20260529212939_world.sql
@@ -1,0 +1,20 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20260529212939');
+IF v = 0 THEN
+INSERT INTO `migrations` VALUES ('20260529212939');
+-- Add your query below.
+
+-- Error shown by honor commands when the honor system is unavailable (before patch 1.4.0).
+DELETE FROM `mangos_string` WHERE `entry` = 269;
+INSERT INTO `mangos_string` (`entry`, `content_default`, `content_loc1`, `content_loc2`, `content_loc3`, `content_loc4`, `content_loc5`, `content_loc6`, `content_loc7`, `content_loc8`) VALUES (269, 'The honor system is not available before patch 1.4.0.', NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL);
+
+-- End of migration.
+END IF;
+END??
+DELIMITER ;
+CALL add_migration();
+DROP PROCEDURE IF EXISTS add_migration;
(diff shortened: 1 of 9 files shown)


ratkosrb:
But honor gain is already disabled before 1.4, so what difference do all these checks make.

mserajnik:
> But honor gain is already disabled before 1.4, so what difference do all these checks make.

Honor gain through regular gameplay is already guarded, yes.

This adds a few things:

1. The weekly honor maintenance doesn't run at all with this (there is no need for it pre-`1.4.0`, so imo it should be skipped outright).
2. GM commands can't be used to modify honor and rank (which, granted, is mostly a UX change to explicitly state "the honor system is disabled").
3. Any potential future code introduced that calls `HonorMgr::Add()` doesn't need to be guarded separately.

I think 1.) is worth having and 3.) is a good pattern even if at the moment the only "beneficiaries" are the two call sites that currently (pre-PR) have the inline guard (since the content that would trigger other call sites only exists from `1.4.0` onwards).

If you think 2.) is unnecessary, I'm happy to drop it. Or, of course, feel free to close the PR if you think it adds nothing of value altogether. 😄

PR #3447, opened 2026-05-31
Implement Volume Caching

Wall-core:
This commit adds volume collision caching. This caches object bounding boxes on AddToWorld and used in dtNavMeshQuery filter construction, allowing dynamic collision to be calculated. This system is done as efficiently as possible, using small vectors with Slab Method calculation to determine movepoints and whether or not they fall within bounds for an object within that specific map instance.

This PR enables DT Virtual Functions in order to override passFilter checks where the volume collision is culled. Currently, we use dynamic ray casting at each movepoint to determine collision with CutPathWithDynamicLoS(); this new system requires no raycasting, which should yield a performance increase in all movement functions.

Currently, only door objects are flagged to be cached; this can be updated to include more object types if necessary.

### DIFF
--- a/src/game/Maps/PathFinder.cpp
+++ b/src/game/Maps/PathFinder.cpp
@@ -26,6 +26,7 @@
 #include "Geometry.h"
 
 #include "Detour/Include/DetourCommon.h"
+#include "Detour/Include/DetourNavMeshQuery.h"
 
 // Distance to target
 #define SMOOTH_PATH_SLOP 0.4f
@@ -71,6 +72,15 @@ bool PathInfo::calculate(Vector3 const& start, Vector3 dest, bool forceDest, boo
     setEndPosition(dest);
     setStartPosition(start);
 
+    // Volume Box Filter handles door collision
+    VolumeBoxFilter volumeFilter(m_sourceUnit->GetMap());
+    volumeFilter.setIncludeFlags(m_filter.getIncludeFlags());
+    volumeFilter.setExcludeFlags(m_filter.getExcludeFlags());
+    for (int i = 0; i < DT_MAX_AREAS; ++i)
+    {
+        volumeFilter.setAreaCost(i, m_filter.getAreaCost(i));
+    }
+
     // A m_navMeshQuery object is not thread safe, but a same PathInfo can be shared between threads.
     // So need to get a new one.
     MMAP::MMapManager* mmap = MMAP::MMapFactory::createOrGetMMapManager();
@@ -85,7 +95,7 @@ bool PathInfo::calculate(Vector3 const& start, Vector3 dest, bool forceDest, boo
     }
     else if (!(m_navMeshQuery = mmap->GetNavMeshQuery(m_sourceUnit->GetMapId())))
     {
-        BuildPathWithoutMMaps(start, dest);
+        BuildPathWithoutMMaps(start, dest, &volumeFilter);
         return true;
     }
 
@@ -99,7 +109,7 @@ bool PathInfo::calculate(Vector3 const& start, Vector3 dest, bool forceDest, boo
     if (!m_navMesh || !m_navMeshQuery || m_sourceUnit->HasUnitState(UNIT_STATE_IGNORE_PATHFINDING) ||
         !HaveTiles(start) || !HaveTiles(dest))
     {
-        BuildShortcut();
+        BuildShortcut(&volumeFilter);
         m_type = PathType(PATHFIND_NORMAL | PATHFIND_NOT_USING_PATH);
         return true;
     }
@@ -119,8 +129,7 @@ bool PathInfo::calculate(Vector3 const& start, Vector3 dest, bool forceDest, boo
     }
     else
     {
-        // target moved, so we need to update the poly path
-        BuildPolyPath(start, dest);
+        BuildPolyPath(start, dest, &volumeFilter);
         return true;
     }
 }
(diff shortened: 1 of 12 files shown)


src/game/Movement/TargetedMovementGenerator.cpp
```diff
@@ -379,10 +368,13 @@ bool ChaseMovementGenerator<T>::Update(T &owner, uint32 const&  time_diff)
             {
                 if (!creature->HasExtraFlag(CREATURE_FLAG_EXTRA_CHASE_GEN_NO_BACKING) && !creature->IsPet() && !i_target.getTarget()->IsMoving())
                 {
-                    if (m_bRecalculateTravel && TargetDeepInBounds(owner, i_target.getTarget()))
-                        DoBackMovement(owner, i_target.getTarget());
-                    else if (m_bCanSpread)
-                        DoSpreadIfNeeded(owner, i_target.getTarget());
+                    if (owner.IsWithinLOSInMap(i_target.getTarget()))
```
ratkosrb: This is not optimal, you are checking LoS too many times. Move and duplicate the check inside each case like this:

if (m_bRecalculateTravel && TargetDeepInBounds(owner, i_target.getTarget()))
{
    if (owner.IsWithinLOSInMap(i_target.getTarget()))
        DoBackMovement(owner, i_target.getTarget());
}
else if (m_bCanSpread)
{
    if (owner.IsWithinLOSInMap(i_target.getTarget()))
        DoSpreadIfNeeded(owner, i_target.getTarget());
}

Or it can be placed inside DoSpreadIfNeeded too, if this is the only place it's called.

src/game/Maps/PathFinder.cpp
```diff
@@ -474,23 +484,31 @@ void PathInfo::BuildPointPath(float const* startPoint, float const* endPoint, fl
...
 
-    // set start and a default next position
-    m_pathPoints[0] = getStartPosition();
-    m_pathPoints[1] = getActualEndPosition();
-
-    m_type = PATHFIND_SHORTCUT;
-    if (m_sourceUnit->CanFly())
-        m_type |= PATHFIND_FLYPATH | PATHFIND_NORMAL;
+    // Check if we're experiencing door collision
+    if (const VolumeBoxFilter* isVolumeFilter = dynamic_cast<const VolumeBoxFilter*>(filter))
```
Gamemechanicwow, line 495: wouldnt this always be true?

src/game/Maps/PathFinder.h
```diff
@@ -146,7 +148,50 @@ class PathInfo
...
+                    polyMax[axis] = vertex[axis];
+            }
+        }
+
+        for (const auto& obj : m_map->GetVolumeCache())
+        {
+            if (!obj.enabled)
+                continue;
+
+            if (PathInfo::IntersectVolumeBox(polyMin, polyMax, obj.worldBounds))
```
Gamemechanicwow: what is the point of using polymin/max? doesnt look like IntersectVolumeBox expects polymin/polymax to be a box boundary - but instead a linesegment?

src/game/Maps/PathFinder.cpp
```diff
@@ -142,12 +151,11 @@ dtPolyRef PathInfo::FindWalkPoly(dtNavMeshQuery const* query, float const* point
...
 }
 
-dtPolyRef PathInfo::getPolyByLocation(float const* point, float *distance, uint32 allowedFlags)
+dtPolyRef PathInfo::getPolyByLocation(float const* point, float *distance, dtQueryFilter* filter, uint32 allowedFlags)
 {
     float closestPoint[VERTEX_SIZE] = {0.0f, 0.0f, 0.0f};
-    dtQueryFilter filter;
-    filter.setIncludeFlags(m_filter.getIncludeFlags() | allowedFlags);
-    dtPolyRef polyRef = FindWalkPoly(m_navMeshQuery, point, filter, closestPoint);
+    filter->setIncludeFlags(filter->getIncludeFlags() | allowedFlags);
```
Gamemechanicwow, line 157: filter used to be local to this function, now it can leak. idk if this affects anything else, but looks like mistake

Gamemechanicwow:
generally not a fan of introducing a new system, since i am still of the opinion that using the stopatfirst option in the dyntree is the best way togo for GO collision. the code is already (almost) in place, but has never been utilized.

src/game/Maps/PathFinder.cpp
Wall-core: No, if the default filter is used (which doesn't account for collision) then it will return false. In all current cases though, the new filter is implemented. But this is mostly to allow multiple filters to exist and be used.

src/game/Maps/PathFinder.h
Wall-core: It's just the name chosen for the diagonal vertices used to determine box boundaries.

src/game/Movement/TargetedMovementGenerator.cpp
Wall-core: yeah this is only called when performing back/spread movements, but I can move the check deeper. This shouldnt be called often though at all.

src/game/Maps/PathFinder.cpp
Wall-core: it can't be called locally with the volume filter as that is created during path generation. How will it leak?

src/game/Maps/PathFinder.h
Gamemechanicwow: box of the polygon? that box boundary can include a GO, without the GO being on the polygon.

src/game/Maps/PathFinder.cpp
Gamemechanicwow: filter is now a pointer.

src/game/Maps/PathFinder.h
Wall-core: it has nothing to do with polygons; the name is a leftover from the previous iteration of the system wherein I had dynamic poly filtering enabled as well, but I didn't include it in this commit.

Wall-core:
I've made updates to the things requested

PR #3450, opened 2026-06-04
Convert even more loose SMSG_ into well defined packet classes

0blu:


lineagedr:
I wonder why each packet structure is heap allocated before being sent.

0blu:
@lineagedr 

This is just preparation.

In the future the whole send system-call and serialization will be performed on a different thread.

PR #3452, opened 2026-06-08
Add missing header files

Stoabrogga:
## 🍰 Pullrequest
Fix build with USE_PCH=OFF

### Proof

- None

### Issues

- None

### How2Test

- Build with USE_PCH=OFF

### Todo / Checklist

- [X] None


ratkosrb:
Can we move the implementation of `passFilter` into the cpp file to avoid including Map.h in the header file?

Stoabrogga:
Done. Hope that's the correct way to do this.

PR #3473, opened 2026-07-03
Improve map extraction tools and fix navmesh generation bugs

schell244:
### Extractors
  - Add `--silent` flag for scripted runs and interactive prompts for resolution options
  - Fix `misc.MPQ` never being added to the archive list in the vmap extractor

### MoveMapGenerator
  - Validate `config.json` at startup, document all parameters, use zero-padded tile keys (`XXYY`)
  - Fix `--skipLiquid` being ignored and per-tile `quick` override leaking into later tiles
  - Fix single-tile builds writing an `.mmap` header incompatible with full builds
  - Fix stale/wrapped terrain height data in `TerrainBuilder::getHeight` and uniform liquid being ignored
  - Fix under-mesh removal for tilted/scaled and overlapping model instances
  - Treat fully submerged steep slopes as swimmable
  - Validate off-mesh connections at build time, clean up `offmesh.txt`
  - Exit non-zero when tiles fail to build instead of pretending success

### Server / Detour
  - `.mmap connect`: per-player state, `cancel` subcommand, radius argument, tile derived from start point
  - Fix swapped x/y in `.mmtile` log filenames and `unpackTileID` masking tileY with `0xFF`
  - PathFinder: keep stored height instead of failing the path when `getPolyHeight` fails at an off-mesh link landing
  - ~~Detour: return early from `raycast` when the path buffer is full~~

### Checked plenty of known problematic spots:
- deadmines bridge no longer needs an offmesh and works both ways dock <-> ship
- stormwind castle pathing seems flawless now
- npc no longer "bump in the air" when walking around fences/bridges
- tested many tents (tanaris, barrens) - ok
- Razorfen (Glutton) - ok
- many more spots...

Co-Authored by Claude Fable 5

### Proof
- None

### Issues
- None

### How2Test
- None

### Todo / Checklist
- [X] None

### DIFF
--- a/contrib/mmap/src/TileWorker.cpp
+++ b/contrib/mmap/src/TileWorker.cpp
@@ -482,7 +482,9 @@ namespace MMAP
         if (config.walkableClimb == 0)
             config.walkableClimb = (int)floorf(agentMaxClimbModelTerrainTransition / config.ch); // For models
         uint32 walkableClimbTerrain = (int)floorf(agentMaxClimbTerrain / config.ch);
-        uint32 walkableClimbModelTransition = config.walkableClimb; // default derived above; config.json can override
+        uint32 walkableClimbModelTransition = config.walkableClimb; // follows walkableClimb unless overridden below
+        if (int const climbOverride = jsonTileConfig["walkableClimbModelTransition"].get<int>())
+            walkableClimbModelTransition = climbOverride;
 
         config.width = config.tileSize + config.borderSize * 2;
         config.height = config.tileSize + config.borderSize * 2;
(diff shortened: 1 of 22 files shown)


ratkosrb:
@Gamemechanicwow Any comments as our expert on maps?

dep/recastnavigation/Detour/Source/DetourNavMeshQuery.cpp
```diff
@@ -2517,8 +2517,15 @@ dtStatus dtNavMeshQuery::raycast(dtPolyRef startRef, const float* startPos, cons
 		if (n < hit->maxPath)
 			hit->path[n++] = curRef;
 		else
+		{
 			status |= DT_BUFFER_TOO_SMALL;
 
+			// Fix: Return early when path buffer is full to prevent continuing with invalid state.
```
Gamemechanicwow: I dont think calling it an invalid state is correct.

Function fills hit->path until its full - and its not used in this function but should (opt) be handled by the caller.

/// If the path array is too small to hold the result, it will be filled as 
/// far as possible from the start postion toward the end position.

if we return early we have t<1 that indicates a hit - when there potentially is a clear path

contrib/mmap/src/TileWorker.cpp
```diff
@@ -471,18 +482,12 @@ namespace MMAP
...
             config.walkableClimb = (int)floorf(agentMaxClimbModelTerrainTransition / config.ch); // For models
         uint32 walkableClimbTerrain = (int)floorf(agentMaxClimbTerrain / config.ch);
-        uint32 walkableClimbModelTransition = (int)floorf(agentMaxClimbModelTerrainTransition / config.ch);
-        if (config.walkableRadius == 0)
-            config.walkableRadius = (int)ceilf(agentRadius / config.cs);
-        if (config.maxEdgeLen == 0)
-            config.maxEdgeLen = (int)(12 / config.cs);
-        if (config.borderSize == 0)
-            config.borderSize = config.walkableRadius + 3;
+        uint32 walkableClimbModelTransition = config.walkableClimb; // default derived above; config.json can override
```
Gamemechanicwow: Can it override? I dont see it anywhere?

PR #3475, opened 2026-07-04
fix creatures not facing their target (client visual bug)

Otari98:
On the video you can see how emote state 193 (EMOTE_STATE_SPELLPRECAST) messes with their facing on the client.

https://github.com/user-attachments/assets/b9cbbc1c-9c91-41a6-897e-d9cb47f9f3a3

https://github.com/user-attachments/assets/ad56916f-31bf-4b2d-968c-6deadfe2f829

### DIFF
--- a/sql/migrations/20260704124546_world.sql
+++ b/sql/migrations/20260704124546_world.sql
@@ -1,0 +1,21 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
+DECLARE v INT DEFAULT 1;
+SET v = (SELECT COUNT(*) FROM `migrations` WHERE `id`='20260704124546');
+IF v = 0 THEN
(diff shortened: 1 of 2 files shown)


Gamemechanicwow:
shouldnt the creatures have this emote state?

my hackfix senses are tingling

Otari98:
Emote state 193 doesnt seem to do anything except make creature unsheath their weapon when not in combat and break their facing in combat, at least this is the case for 1.12 client, i cant test other builds.
Im not sure if there is logic tied to them having this emote state. The only mention of EMOTE_STATE_SPELLPRECAST in core i found is in DM east last boss ai but its commented out.

Gamemechanicwow:
There is nothing wrong with them having the emote_state. At least 27675, 27681, 27683, 27684, 27685 are confirmed by vanilla sniffs - couldn't bother to check all of them.

The problem imo is that the emote_state isn't removed when entering combat.

PR #3483, opened 2026-07-11
Add missing include

Stoabrogga:
## 🍰 Pullrequest
Add a missing include, otherwise the build with `USE_PCH=OFF` throws errors.

### Proof

Error message:
```
/home/u1/vmangos/core/src/game/Server/Packets/Combat.h:165:9: error: ‘nonstd’ does not name a type; did you mean ‘const’?
  165 |         nonstd::optional<DebugMeleeAttackingStateInfo> debugInfo;
      |         ^~~~~~
      |         const
```

### How2Test
Use `USE_PCH=OFF` to build the core.

### DIFF
--- a/src/game/Server/Packets/Combat.h
+++ b/src/game/Server/Packets/Combat.h
@@ -4,6 +4,7 @@
 #include "Packet.h"
 #include "ObjectGuid.h"
 #include "DamageStructs.h"
+#include "nonstd/optional.hpp"
 
 namespace WorldPackets { namespace Combat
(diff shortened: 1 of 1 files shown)


mserajnik:
Might be worth it to have an additional `USE_PCH=OFF` CI build to catch these?

Stoabrogga:
> Might be worth it to have an additional `USE_PCH=OFF` CI build to catch these?

I think that's a good idea.

PR #3487, opened 2026-07-11
GetHeightStatic - add a delta to Z before choosing vmap height over mapheight

### LINKED ISSUE #3485: 🐛 [Bug] Pet falls through the ground
## 🐛 Bug report

The pet falls through the ground:
https://github.com/user-attachments/assets/832d2f4b-c962-4392-8ca1-29ee66d29f5c

### Expected behavior

Pet should stay on the ground.

### Steps to reproduce
`.go -10271.253906 -3949.695801 19.237492 0`

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash: c46aa93f970cbaed2bff63004882437491fc9660

OS Client: Debian 13
OS Server: Debian 13

### Crashlog

None

Gamemechanicwow:
## 🍰 Pullrequest

Prevent switching to vmap height, when z is slightly below mapheight,

Issue from https://github.com/vmangos/core/issues/3485 explained:

Pet is moving from ground to water. Underneath the terrain we have the sunken temple.
image

issue was traced to:
UpdateAllowedPositionZ -> GetWaterOrGroundLevel -> getHeightStatic

returning Z based on vmap height instead of mapheight.

Notes: 

- Probably needs more thorough testing (does it cause issues elsewhere in the world)
- Maybe some tinkering needed to find the best delta value

### Proof

- None

### Issues

- fixes https://github.com/vmangos/core/issues/3485

### How2Test

- None

### Todo / Checklist

- [X] None

### DIFF
--- a/src/game/Maps/GridMap.cpp
+++ b/src/game/Maps/GridMap.cpp
@@ -860,7 +860,7 @@ float TerrainInfo::GetHeightStatic(float x, float y, float z, bool useVmaps/*=tr
             // we have mapheight and vmapheight and must select more appropriate
 
             // we are already under the surface or vmap height above map heigt
-            if (z < mapHeight || vmapHeight > mapHeight)
+            if (z + 1.0f  < mapHeight || vmapHeight > mapHeight)
                 return vmapHeight;
             return mapHeight;                               // better use .map surface height
         }


Stoabrogga:
Tested this and it seems to fix the issue. But I don't know if there are some edge cases where this change could cause other issues. Played a bit with my warlock, no issues yet.

schell244:
How about selecting by proximity instead of using a fixed delta, picking the candidate that's closer to z?

```cpp
  // vmap height above map height
  if (vmapHeight > mapHeight)
      return vmapHeight;

  // under the terrain surface: pick the height nearest to z, so a unit barely below the terrain snaps back to the terrain,
  //  while a unit deep inside a model (cave, building) keeps the model floor
  if (z < mapHeight && mapHeight - z > std::abs(z - vmapHeight))
      return vmapHeight;
  return mapHeight;
```

relh:
Data point on the delta, since "maybe some tinkering needed" was left open: **0.05f** also fixes the #3485 case — the wobble there is sub-millimeter, and 0.05f is the cushion the codebase already trusts for exactly this (`WorldObject::UpdateGroundPositionZ` uses `z = new_z + 0.05f`, "few pixel under the surface"). The cost of 1.0f is the other direction: a unit genuinely up to a yard below the .map surface (shallow cave mouths, undercut terrain seams where the vmap floor is correct) now snaps up to the outdoor surface. A smaller delta may be the more conservative choice.

mserajnik:
> The cost of 1.0f is the other direction: a unit genuinely up to a yard below the .map surface (shallow cave mouths, undercut terrain seams where the vmap floor is correct) now snaps up to the outdoor surface.

Do you have a concrete example where this happens?

Gamemechanicwow:
> The cost of 1.0f is the other direction: a unit genuinely up to a yard below the .map surface (shallow cave mouths, undercut terrain seams where the vmap floor is correct) now snaps up to the outdoor surface. A smaller delta may be the more conservative choice.

1.0f is halfheight of the human male playermodel. Can you see the problem with your logic?

PR #3492, opened 2026-07-18
Add CI build for USE_PCH=OFF

Stoabrogga:
## 🍰 Pullrequest
Add an additional CI build to check building with "USE_PCH=OFF". Also fix a "range-loop-construct" warning.

### How2Test
Check CI build "Ubuntu GCC (PCH OFF)"

### DIFF
--- a/src/game/SniffFile.h
+++ b/src/game/SniffFile.h
@@ -51,7 +51,7 @@ public:
     void WriteToFile(Container<LoggedPacket, Args...> const& container)
     {
         WriteHeader();
-        for (auto const &itr : container)
+        for (auto const& itr : container)
             WritePacket(itr);
     }
 private:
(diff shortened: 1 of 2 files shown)


src/game/SniffFile.h
```diff
@@ -51,7 +51,7 @@ class SniffFile
     void WriteToFile(Container<LoggedPacket, Args...> const& container)
     {
         WriteHeader();
-        for (auto const itr : container)
+        for (auto const &itr : container)
```
0blu: We technically have the `&` right next to the `const` in this project.

```suggestion
        for (auto const& itr : container)
```

ratkosrb:
Why not change the existing job to not use PCH, do we really need to test both with and without? I figure if it compiles without then it will compile with it too.

Stoabrogga:
> Why not change the existing job to not use PCH, do we really need to test both with and without? I figure if it compiles without then it will compile with it too.

So, you mean to switch off PCH for the already existing job "Ubuntu GCC (client build 5875)"?

ratkosrb:
Or the Ubuntu clang one, so we don't torture the CI with too many jobs.

Stoabrogga:
Ok, I added it to the Ubuntu Clang job.

PR #3499, opened 2026-07-20
Remove stalked-target auras from the target when the caster dies

### LINKED ISSUE #1615: 🐛 [Bug] Hunter's Mark remains on death
## 🐛 Bug report

If a hunter uses Hunter's Mark and dies the debuff remains on the applied enemy.

### Expected behavior

Hunter's Mark debuff should fade from the applied enemy if the hunter dies.

Video from classic:
https://youtu.be/udf7wTHsPoQ?t=240

Video from vanilla:
https://www.youtube.com/watch?v=2lpw74pRbFA&t=306s

I've also tested this on classic myself by applying Hunter's Mark on enemy and then dying.
The debuff will fade.

### Steps to reproduce

1. Create a Night Elf hunter
2. .learn 14325
3. Use Hunter Mark on closest enemy target.
5. Target yourself
6. .die
7. The debuff will remain on the enemy target despite your death.

### Version & Environment

Client Version: 1.12.1.5875

Commit Hash:  https://github.com/vmangos/core/commit/373fde44b3c8c6f5236656e31dabffbb5f728037

OS Client: Microsoft Windows 11 Pro N 10.0.22000 Build 22000
OS Server: Microsoft Windows 11 Pro N 10.0.22000 Build 22000

### Crashlog

- None

relh:
## 🍰 Pullrequest

When a hunter dies, Hunter's Mark currently stays on the target for its full duration. In vanilla the mark drops when its caster dies.

Per review feedback this now keys on `SPELL_AURA_MOD_STALKED` instead of checking for Hunter's Mark specifically: on death, the dying unit's single-cast-target registry is walked and stalked-target auras are removed from their targets, before local death-aura cleanup runs.

In the 1.12 DBC the only single-target spells carrying the stalked aura are the four Hunter's Mark ranks, Mind Vision (a channel that already ends on caster death), and the internal Syndicate Tracker (MURP) DND spell, so the generic check stays tightly scoped in practice.

### Proof
Vanilla footage of the mark dropping on hunter death (see the videos in #1615). Tested with a stock 1.12.1.5875 client: the mark drops from the target frame on caster death; other single-target auras are unaffected.

### Issues
- fixes #1615

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -4209,6 +4209,27 @@ void Unit::RemoveAuraTypeOnDeath(AuraType auraType)
 
 void Unit::RemoveAllAurasOnDeath()
 {
+    // Auras that track their target for the caster (SPELL_AURA_MOD_STALKED,
+    // i.e. Hunter's Mark) end when the tracking caster dies. Other
+    // caster-owned single-target auras retain their existing death behavior.
+    SingleCastSpellTargetMap& singleTargets = GetSingleCastSpellTargets();
+    for (SingleCastSpellTargetMap::iterator itr = singleTargets.begin(); itr != singleTargets.end();)
+    {
+        SpellEntry const* spellInfo = itr->first;
+        if (!spellInfo->HasAura(SPELL_AURA_MOD_STALKED))
+        {
+            ++itr;
+            continue;
+        }
+
+        ObjectGuid const targetGuid = itr->second;
+        singleTargets.erase(itr);
+        if (IsInWorld())
+            if (Unit* target = GetMap()->GetUnit(targetGuid))
+                target->RemoveAurasByCasterSpell(spellInfo->Id, GetObjectGuid(), AURA_REMOVE_BY_DEATH);
+        itr = singleTargets.begin();
+    }
+
     // used just after dieing to remove all visible auras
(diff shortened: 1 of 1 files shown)


ratkosrb:
I'm not sure what's the best way to handle this. I think it would make more sense if it's a generic check for auras with SPELL_AURA_MOD_STALKED. Or perhaps a spell script on Hunter's Mark. But checking for Hunter's Mark specifically in generic Unit code is not good.

ratkosrb:
Another case that should be tested on official servers is what should happen if you cast Hunter's Mark on someone and then log out. Does it stay on target with offline caster? This should tell us if it should be a spell script or on death event for SPELL_AURA_MOD_STALKED.

relh:
Good call — reworked it to key on SPELL_AURA_MOD_STALKED, so there's no Hunter's Mark special-casing in Unit code anymore. For what it's worth, the only single-target spells with that aura in the 1.12 DBC are the four Hunter's Mark ranks, Mind Vision (a channel, so it already ends on death anyway), and an internal DND test spell, so it stays pretty contained.

relh:
I'm going to test all these weird exciting remaining cases and report back

ratkosrb:
Seems like it's not removed if hunter logs out, so we can merge it as it is.

PR #3502, opened 2026-07-20
Share vendor item visibility between list and purchase paths

### LINKED ISSUE #2591: 🐛 [Bug] Vendor issue when conditions are used
Example:
Evie Whirlbrew (Entry 11188) has the Mooncloth Pattern on vendor list, using condition 4000 (won't show in shop unless you're a tailor). The problem seems to be that this throws off limited stock items when purchasing, since there's an item above it in the list but the client doesn't know this. If you purchase the Greater Mana Potion from her, the next item on the list has its count decremented instead. This is simply a visual bug and doesn't prevent shops from working properly.

relh:
## 🍰 Pullrequest

`SendListInventory` filters vendor rows (class, race, reputation, conditions) and numbers only the visible survivors, but `BuyItemFromVendor` still works in raw backing-list indices:

- `SMSG_BUY_ITEM` reports the raw index, so a hidden earlier row makes the client apply the limited-stock decrement to the wrong displayed item — exactly the Evie Whirlbrew case in #2591.
- Rows hidden from the list are still directly purchasable, because the buy path re-implements only a subset of the list's filters.

This introduces a single `IsVendorItemVisible` contract used by both paths: the list keeps its exact current output, the buy path rejects filtered rows like any invalid slot, and the purchase response reports the compact visible slot the client actually displays.

### Proof
Reproduces on current `development` per #2591's recipe; with this change the limited-stock decrement lands on the purchased row. Tested with a stock 1.12.1.5875 client across vendors with condition-hidden rows; normal vendors unchanged.

### Issues
- fixes #2591

### DIFF
--- a/src/game/Handlers/ItemHandler.cpp
+++ b/src/game/Handlers/ItemHandler.cpp
@@ -754,35 +754,8 @@ void WorldSession::SendListInventory(ObjectGuid vendorguid, uint8 menu_type)
         {
             if (ItemPrototype const* pProto = sObjectMgr.GetItemPrototype(crItem->item))
             {
-                if (!_player->IsGameMaster())
-                {
-                    // class wrong item skip only for bindable case
-                    if ((pProto->AllowableClass & _player->GetClassMask()) == 0 && pProto->Bonding == BIND_WHEN_PICKED_UP)
-                        continue;
-
-                    // race wrong item skip always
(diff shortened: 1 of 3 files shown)


ratkosrb:
Please restore all the original comments to the code in Player::IsVendorItemVisible.

relh:
Done — restored the original comments in IsVendorItemVisible, and moved the BuyItemFromVendor comment back above the right function while I was at it.

PR #3505, opened 2026-07-20
Send melee attack state before lethal damage

### LINKED ISSUE #3326: 🐛 [Bug] No Floating Combat text on auto attack killing blows
## 🐛 Bug report
If the hit that kills a creature is a white auto attack, it is not displayed as floating combat text on the client (but is in the combat log). Apparently this started happening ~half a year ago on vmangos.

 Expected behavior
Should see floating combat text even on killing blows with auto attacks.
https://youtu.be/SwLjJH8LXfc?list=PL49Ga_-r7iKkxrbckm9CkUhWzxqrOnw-n&t=132 as seen here and many other videos from vanilla

### Steps to reproduce
Just use auto attack to kill any creature.

  Client Version:
  - 1.12.1.5875

OS Client: Windows 11
OS Server: Windows 11

### Crashlog
- None

nishu-builder:
## Pull request

Send `SMSG_ATTACKERSTATEUPDATE` after chance-on-hit procs but before applying melee damage. On a lethal white hit, applying damage first can advance the victim through its death lifecycle before the client receives the swing packet, leaving no victim object to anchor floating combat text.

The change preserves the sniff-backed ordering relative to chance-on-hit spell casts and keeps damage after procs, so victim auras can still affect the attacker on a killing blow.

### Proof

- The Vanilla footage and reproduction are documented in #3326.
- The nonlethal packet contents and combat-log behavior are unchanged; only dispatch relative to the server-side death transition moves.

### Issues

- Fixes #3326

### How2Test

1. Create a character and attack a low-level creature using only white auto attacks.
2. Let an auto attack deliver the killing blow.
3. Confirm the killing-blow damage appears as floating combat text as well as in the combat log.
4. Repeat with a chance-on-hit weapon effect and confirm its spell animation still precedes the melee swing feedback.

### Todo / Checklist

- [x] Keep chance-on-hit proc ordering unchanged.
- [x] Keep melee damage after proc processing.
- [x] Limit the change to `Unit::AttackerStateUpdate`.

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -2196,12 +2196,12 @@ void Unit::AttackerStateUpdate(Unit* pVictim, WeaponAttackType attType, bool ext
 
     ProcDamageAndSpell(ProcSystemArguments(damageInfo.target, damageInfo.procAttacker, damageInfo.procVictim, damageInfo.procEx, damageInfo.totalDamage, damageInfo.totalDamage + damageInfo.totalAbsorb + damageInfo.totalResist, damageInfo.attackType));
 
-    // In sniffs SMSG_ATTACKERSTATEUPDATE follows chance-on-hit spell casts.
+    // In sniffs SMSG_ATTACKERSTATEUPDATE is sent after chance on hit spell casts from CastItemCombatSpell. This fixes animation for Frostbrand Attack.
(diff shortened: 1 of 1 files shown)


ratkosrb:
Please keep the original comments.

PR #3533, opened 2026-08-18
Prevent visual bug on older clients where mobs appear to teleport to player

### LINKED ISSUE #1262: 🐜 [Bug][1.6.1] Mob teleport to the new highest hate player's side instantly when mob's hate transfer to other player.
### Expected behavior

mod should move to player normally , not teleport.
### Steps to reproduce

1. Build a server of CLIENT_BUILD_1_6_1, use 1.6.1 client to log on server.
2. create a paladin, learn spell=853 .
3. attack a mob, try to stun this mod , then run away, you will see mob teleport to you.

This issue only happen on `CLIENT_BUILD_1_6_1`, I record two video, you can tell the different between two versions
in **1.12** , there is no this issue. proof: https://streamable.com/q1jqbo
in **1.6**, mod teleport to player's side. proof https://streamable.com/grm0cp
anothe 1.6 video , two players battle a mob , plz watch from 0:30 https://streamable.com/y4dy9x
### Version & Environment

Client Version: 1.6.1.4544
CLIENT_BUILD_1_6_1
Server Patch: 1.2

Commit Hash:  

OS Client: Windows 10
OS Server: Windows server 2008 r2
### Crashlog

- None

moyashiparty:
## 🍰 Pullrequest

If a mob breaks out of root or stun, and the player has moved away, the mobs will appear to teleport beside the player (presumably to the end location of their existing movement spline?) on clients 1.6.1 and older.

Now I really have no idea if this fix is proper or not... Honestly, I've had sporadic attempts fixing this problem for years now without any luck, but I was getting annoyed at how irritating it was to AoE farm mobs with this bug present so I sat down for a few hours tonight trying random shit until surprisingly this worked. Maybe it'll get the ball rolling on a real fix, or maybe the conditional `StopMoving` is good enough of an improvement for the older clients.

Feel free to just close this if there's a better solution.

### Proof
Before (stun):
https://github.com/user-attachments/assets/5f22ad07-1b46-42a2-b2d2-f281bd1b6801

Before (root):
https://github.com/user-attachments/assets/8d2e6b26-e687-4f2a-90b5-d07c22782a0b

After (stun):
https://github.com/user-attachments/assets/a1f8b9c6-9ee4-46d5-91b4-479150e4266d

After (root):
https://github.com/user-attachments/assets/6b56f68d-6240-4f82-a001-1ab0d00494bc

### Issues

- I believe this fixes https://github.com/vmangos/core/issues/1262

### How2Test

- Use an older client (I've seen this happen on 1.6.1 and 1.5.1) 
- Stun or root a mob
- Move away from the mob while it is stunned/rooted
- Most of the time (not always) it will appear to teleport beside you when the stun/root breaks

### DIFF
--- a/src/game/Objects/Unit.cpp
+++ b/src/game/Objects/Unit.cpp
@@ -7425,6 +7425,14 @@ void Unit::SetRooted(bool apply)
     {
         SetRootedReal(apply);
         MovementPacketSender::SendMovementFlagChangeToAll(this, MOVEFLAG_ROOT, apply);
+
+        // Hackfix: Old clients teleport unit to last point if new spline begins right after unroot.
+        // In 1.8 sniffs it appears that the creature only begins chasing on the next world update,
+        // but motion is updated right away in our core, so send a sacrificial STOP spline after unroot.
+#if SUPPORTED_CLIENT_BUILD <= CLIENT_BUILD_1_6_1
+        if (!apply)
+            StopMoving(true);
+#endif
     }
 }
 


ratkosrb:
It seems that the next movement spline starting too soon after the unroot trips up the client. In the 1.8 sniffs it's the same packet used to unroot the creature (MSG_MOVE_UNROOT) but it only begins chasing again on the next world update. In our core it happens immediately after, Unit::Update updates MotionMaster after it's updated the auras (where unroot happens).

ratkosrb:
Can you move the hackfix inside Unit:SetRooted so we don't duplicate the code? And we shouldn't check for combat, cause the teleport bug can affect any kind of movement that begins right after root ends.

image

moyashiparty:
> Can you move the hackfix inside Unit:SetRooted so we don't duplicate the code? And we shouldn't check for combat, cause the teleport bug can affect any kind of movement that begins right after root ends.

Sure I made that change and included your comment, also confirmed it still fixes the issue for me.

src/game/Objects/Unit.cpp
```diff
@@ -7425,6 +7425,14 @@ void Unit::SetRooted(bool apply)
     {
         SetRootedReal(apply);
         MovementPacketSender::SendMovementFlagChangeToAll(this, MOVEFLAG_ROOT, apply);
+
+        // Hackfix: Old clients teleport unit to last point if new spline begins right after uproot.
```
ratkosrb: typo uproot instead of unroot

ratkosrb: I am not able to edit it myself, my browser doesn't support the web editor anymore, I need you to fix the typo before I can merge it.

moyashiparty: Yep fixed, sorry about that

PR #3541, opened 2026-08-23
Fix/innodb engine conversion

tomaszkubiak-courses:
### Proof

- None

### Issues

- None

### How2Test

- None

### Todo / Checklist

- [X] None


Wall-core:
you should leave the tables using MyIsam as they are; they use it for performance over InnoDB

tomaszkubiak-courses:
it was my mistake to do this PR. sorry for disturbance. i'm using InnoDB for my local version and it was not meant to be published here.

PR #3549, opened 2026-09-01
Add Events list for Khan Hratha (credit cmangos)

NickTyrer:
## 🍰 Pullrequest


NickTyrer:
Is there any way to find which creatures have text associated with them in the sniffs and have no EventAI in vmangos?

ratkosrb:
```
SELECT *
FROM sniffs_combined3.creature_unique_text
WHERE entry IN (SELECT entry FROM mangos.creature_template WHERE ai_name='' && script_name='')
```

ratkosrb:
Most results are likely already used but are triggered from a script that doesn't run on the creature itself. So you would have to sift through them manually and figure out which are aggro texts.

PR #3550, opened 2026-09-01
Fix log errors

Stoabrogga:
## 🍰 Pullrequest
Fix these errors:
```
ERROR: CreatureEventAI: EventMap for Creature 2941 is empty but creature is using CreatureEventAI.
ERROR: CreatureEventAI: EventMap for Creature 10042 is empty but creature is using CreatureEventAI.
ERROR: CreatureEventAI: EventMap for Creature 10378 is empty but creature is using CreatureEventAI.
ERROR: Table creature_movement contains path for creature guid 40706, but this creature guid does not exist. Skipping.
```

### DIFF
--- a/sql/migrations/20260901174710_world.sql
+++ b/sql/migrations/20260901174710_world.sql
@@ -1,0 +1,19 @@
+DROP PROCEDURE IF EXISTS add_migration;
+DELIMITER ??
+CREATE PROCEDURE `add_migration`()
+BEGIN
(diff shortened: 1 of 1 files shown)


ratkosrb:
Corrupted Saber (10042) has to use EventAI despite having no events, for the quest script to work. It's intentionally this way. The other cases will be corrected soon.

Stoabrogga:
Ok, thanks, will close the PR.

PR #3563, opened 2026-09-20
Fix AQ20 Kurinnaxx, Obsidian Destroyer, Moam and AQ40 Obsidian Eradicator.

### LINKED ISSUE #2809: 🦂 [Bug]NPC Moam 15340 ,after the battle, the health will not be restored full,Applied to the same model
### Steps to reproduce

1.NPC Moam 15340 ,after the battle, the health will not be restored full,Applied to the same model 
2.That's also the case      creature 15338 and 15262
3.
4. ...
5. Profit

### Version & Environment

Client Version: 

Commit Hash:  

OS Client: 
OS Server: 

### Crashlog

- None

Daribon:
## 🍰 Pullrequest

- Reworked Moam using sniffs, with phase two filled in from videos and Warcraftlogs.
(I had no sniffs for phase 2, but it isn't complex enough to require sniffing. The only thing that might still be slightly off is the counterspell timing. Before this, Mana Fiends were spam counterspelling with just a 1s delay, which was clearly wrong.)
- Reworked Kurinnaxx based on sniffs, sand trap now uses spell as intended.
- Moved Obsidian Destroyer to database.
- Cleaned up Obsidian Eradicator in AQ40 to use correct spells.
- Added Moam summon spell script for Mana Fiends, Stone Form now either ends when all fiends are dead or timer expired.
- Fixed CreatureEventAI warning for PetEventAI. (With just PetAI the mana fiend spells would never cast, so I had to use EventAI, which then results in PetEventAI.)
- Moam, Obsidian Destroyer and Obsidian Eradicator now reset HP and mana on reset correctly.
- Removed custom boss script npc_sandstalker.cpp (not used anywhere) 

### Proof

- My own sniffs from Season of Mastery (patch 1.14.3 between June and July 2022)

### Issues

- fixes #2809

### How2Test

- .tele aq20
- .go creature destroyer
- .go creature moam
- .tele aq40
- .go creature eradicator

### Todo / Checklist

- [X] None


Daribon:
Not ready for merge yet, still investigation final things.

